PR #13, opened 2020-06-10
Rename module

FFY00:


merwok:
`build` is used so often for a build directory, ignored and removable, that it looks really weird as a Python module.

Why not use a unique, searchable name?
Also, if this is small, no need to make a package, a one-file module is fine :slightly_smiling_face:

FFY00:
> `build` is used so often for a build directory, ignored and removable, that it looks really weird as a Python module.

Since `pybuild` is already taken by Debian I think I would keep the name as `python_build`. The question now is if we should allow it to be invoked as `python -m build` or not? In this PR I renamed it to `python_build` but kept a `python -m build` entrypoint.

> Why not use a unique, searchable name? (FYI pybuild is already a tool used by Debian)

Hence the `python-build` naming, I believe that term is searchable enough.

>  Also, if this is small, no need to make a package, a one-file module is fine :slightly_smiling_face:

The module right now is pretty small, but it is at a state where if I join `__init__.py` and `__main__.py` it doesn't give us much space to grow IMO. On top of that, I like the logical separation between the two files. What do you think?

uranusjr:
The problem is not what to use for `python -m`, but that `build` is too common a name for a directory.

Personally I would go with a more unique name. `python-build` is also not a very good name since it is [the pyenv command to build a Python installation](https://github.com/pyenv/pyenv/tree/master/plugins/python-build), and I expect most Google results to point there due to pyenv’s popularity.

May I suggest naming the module `casei`? The term *wheel* is a reference to *cheeseshop*, code name of the old PyPI implementation, because the idea is like a cheese wheel, pre-built and ready to be sliced and consumed. Cheesemaking is sometimes called caseiculture (the *casei* part came from Latin caseus, i.e. cheese).

We can have a `build.py` in the repository (but not in the distributed package) that imports and calls the module, so `python -m build` continues to work, but avoids it being a directory.

FFY00:
Okay, that is fair.

> May I suggest naming the module `casei`? The term _wheel_ is a reference to _cheeseshop_, code name of the old PyPI implementation, because the idea is like a cheese wheel, pre-built and ready to be sliced and consumed. Cheesemaking is sometimes called caseiculture (the _casei_ part came from Latin caseus, i.e. cheese).

My goal here was to make the name dead simple, so that it was obvious what it was meant to do, and to be something that was easy to remember. I wanted this project to be seen as _the_ way to build sdists/wheels in a standalone fashion, and not just a different way. I modeled it after `compileall`, that is what I am trying to achieve.

I am hesitant to use `casei` because of this, but I am not ruling it out.

> We can have a `build.py` in the repository (but not in the distributed package) that imports and calls the module, so `python -m build` continues to work, but avoids it being a directory.

Why not distribute it with the package?

uranusjr:
> Why not distribute it with the package?

I have no hard objections, but it’s generally considered suboptimal practice for a package to install modules with drastically different names. So I’d think this would need further discussions.

FFY00:
I have been trying to brainstorm some names but without much success. I think `casei` would work, though it is not optimal IMO.

> I have no hard objections, but it’s generally considered suboptimal practice for a package to install modules with drastically different names. So I’d think this would need further discussions.

I agree. I acknowledge that generally not great to install such modules, but given the context I think it would be reasonable. Other people might not feel the same way though, and I am more than happy to hear their POV.

pganssle:
I'm not crazy about `casei` because it's strange to pronounce and possibly hard to remember, but I imagine people will get over it. Another option is `pyproject-build`, I suppose.

Coming up with alternate names is harder than I thought.

pyproject.toml
```diff
@@ -1,32 +1,51 @@
...
-author-email = 'lains@archlinux.org'
-home-page = 'https://github.com/FFY00/python-build'
-description-file = 'README.md'
-requires = [
-    'toml',
-    'pep517',
-    'packaging',
-    'importlib_metadata;python_version<"3.8"',
-    'typing;python_version<"3"',
+requires = ['poetry>=0.12']
```
pganssle, line 2: Why the switch to `poetry`? Seems like overkill for a super minimal tool.

Also, it seems weird that this package has both a `setup.py` and is built with something other than `setuptools`.

FFY00: > Why the switch to `poetry`? Seems like overkill for a super minimal tool.

I was using flit but it does not support specifying multiple packages.

> Also, it seems weird that this package has both a `setup.py` and is built with something other than `setuptools`.

Well, I liked the nicer CLI but I guess I can just use `setuptools`, it's not a big deal. I kept `setuptools` to be easier for packagers.

FFY00:
A possible name would be `builder`, but that is also a bit common.

pyproject.toml
pganssle: Do you actually want to distribute multiple packages? I generally see that as an anti-pattern except in very rare situations. It's usually better to namespace them under a top-level package.

I don't think you have to change whatever you don't want to change, but I would expect that you wouldn't use `setuptools`' CLI at all, correct? Just `nox` or something of that nature (or possibly `casei` itself, for the build).

In any case, the main problem I see is that you are trying to maintain compatibility with bootstrapping via `setuptools` (or presumably `distutils`, I guess, though `distutils` will soon be part of `setuptools` and not in the standard library), so you end up needing to duplicate all the project metadata, which can easily get out of sync.

merwok:
I find `casei` a little far-fetched (bacteria → yogurt/milk → cheese → Monty Python reference in previous version of the Package Index), but at leasts it’s searchable, compared to `packaging` or `cryptography` (these are Python projects too).

pyproject.toml
FFY00: > Do you actually want to distribute multiple packages? I generally see that as an anti-pattern except in very rare situations. It's usually better to namespace them under a top-level package.

Yes. I would really like to be able to ship a `build` module, so that plain builds (no virtual envs) can be done by invoking `python -m build`.

> I don't think you have to change whatever you don't want to change, but I would expect that you wouldn't use `setuptools`' CLI at all, correct? Just `nox` or something of that nature (or possibly `casei` itself, for the build).

Yeah, generally I wouldn't really be using the `setuptools` CLI as the package can build itself via PEP517. I just remembered my main motivation against `setuptools`, it does not create `py2.py3` wheels, only `py2` or `py3`. Either way, this is a nitpick.

> In any case, the main problem I see is that you are trying to maintain compatibility with bootstrapping via `setuptools` (or presumably `distutils`, I guess, though `distutils` will soon be part of `setuptools` and not in the standard library), so you end up needing to duplicate all the project metadata, which can easily get out of sync.

Yes, but since right now I am the only maintainer, I have it handled. That could change in the future.

merwok: > I would really like to be able to ship a `build` module

But why is it so important to use a very generic name here?
To give the impression that this is somehow built-in to Python?

FFY00: > But why is it so important to use a very generic name here?
> To give the impression that this is somehow built-in to Python?

To give the impression it just *builds*.

FFY00:
I have broken out the poetry patch from this PR. I will probably re-evaluate the build system status quo in the future, but I don't want to worry about it for now. It is provided as a "nice to have".

FFY00:
Closing this as we decided against the rename.

PR #18, opened 2020-06-12
main: build in a virtual env by default

FFY00:
As discussed in [1] python-build should now build inside a virtual
environment, as recommended in PEP517. This patch implements that
functionality and sets it as the default behavior. It also introduces a
--no-isolation flag to disable this.

[1] https://discuss.python.org/t/moving-python-build-to-pypa/4390

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,8 +33,8 @@ jobs:
     - name: Install build backend
       run: pip install setuptools wheel
 
-    - name: Build python-build
-      run: python -m build -x
+    - name: Build python-build (without isolation)
+      run: python -m build -x -n
 
 
   pip:
@@ -67,8 +67,8 @@ jobs:
     - name: Install build backend
       run: pip install setuptools wheel
 
-    - name: Build pip
-      run: python -m build -x -w ../pip
+    - name: Build pip (without isolation)
+      run: python -m build -x -n -w ../pip
 
 
 #  python-install:
@@ -132,7 +132,7 @@ jobs:
       if: ${{ matrix.python == 2.7 }}
       run: pip install typing
 
-    - name: Build dateutil
+    - name: Build dateutil (with isolation)
       run: python -m build -x -w ../dateutil
 
 
@@ -159,5 +159,5 @@ jobs:
     - name: Install dependencies
       run: pip install toml pep517 packaging importlib_metadata
 
-    - name: Build Solaar
+    - name: Build Solaar (with isolation)
       run: python -m build -x -w ../solaar
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -134,6 +134,8 @@ jobs:
 
     - name: Build dateutil (with isolation)
       run: python -m build -x -w ../dateutil
+      env:
+        SETUPTOOLS_SCM_PRETEND_VERSION: dummy
 
 
   Solaar:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -132,9 +132,6 @@ jobs:
       if: ${{ matrix.python == 2.7 }}
       run: pip install typing
 
-    - name: Install build backend
-      run: pip install setuptools wheel
-
     - name: Build dateutil
       run: python -m build -x -w ../dateutil
 
@@ -161,9 +158,6 @@ jobs:
 
     - name: Install dependencies
       run: pip install toml pep517 packaging importlib_metadata
-
-    - name: Install build backend
-      run: pip install setuptools wheel
 
     - name: Build Solaar
       run: python -m build -x -w ../solaar
--- a/build/__init__.py
+++ b/build/__init__.py
@@ -24,6 +24,15 @@ if sys.version_info < (3,):
 ConfigSettings = Dict[str, Union[str, List[str]]]
 
 
+_DEFAULT_BACKEND = {
+    'build-backend': 'setuptools.build_meta:__legacy__',
+    'requires': [
+        'setuptools >= 40.8.0',
+        'wheel'
+    ]
+}
+
+
 class BuildException(Exception):
     '''
     Exception raised by ProjectBuilder
@@ -92,16 +101,14 @@ class ProjectBuilder(object):
         except toml.decoder.TomlDecodeError as e:
             raise BuildException("Failed to parse pyproject.toml: {} ".format(e))
 
-        try:
-            self._build_system = self._spec['build-system']
-        except KeyError:
-            self._build_system = {
-                'build-backend': 'setuptools.build_meta:__legacy__',
-                'requires': [
-                    'setuptools >= 40.8.0',
-                    'wheel'
-                ]
-            }
+        self._build_system = self._spec.get('build-system', _DEFAULT_BACKEND)
+
+        if 'build-backend' not in self._build_system:
+            self._build_system['build-backend'] = _DEFAULT_BACKEND['build-backend']
+            self._build_system['requires'] = self._build_system.get('requires', []) + _DEFAULT_BACKEND['requires']
+
+        if 'requires' not in self._build_system:
+            raise BuildException("Missing 'build-system.requires' in pyproject.yml")
 
         self._backend = self._build_system['build-backend']
 
@@ -112,6 +119,10 @@ class ProjectBuilder(object):
 
         self.hook = pep517.wrappers.Pep517HookCaller(self.srcdir, self._backend,
                                                      backend_path=self._build_system.get('backend-path'))
+
+    @property
+    def build_dependencies(self):  # type: () -> Set[str]
+        return set(self._build_system['requires'])
 
     def get_dependencies(self, distribution):  # type: (str) -> Set[str]
         '''
--- a/build/__main__.py
+++ b/build/__main__.py
@@ -7,6 +7,8 @@ import traceback
 
 from typing import List, Optional
 
+import pep517.envbuild
+
 from . import BuildBackendException, BuildException, ConfigSettings, ProjectBuilder
 
 
@@ -27,8 +29,26 @@ def _error(msg, code=1):  # type: (str, int) -> None  # pragma: no cover
     exit(code)
 
 
-def build(srcdir, outdir, distributions, config_settings=None, skip_dependencies=False):
-    # type: (str, str, List[str], Optional[ConfigSettings], bool) -> None
+def _build_in_isolated_env(builder, outdir, distributions):  # type: (ProjectBuilder, str, List[str]) -> None
+    with pep517.envbuild.BuildEnvironment() as env:
+        env.pip_install(builder.build_dependencies)
+        for distribution in distributions:
+            builder.build(distribution, outdir)
+
+
+def _build_in_current_env(builder, outdir, distributions, skip_dependencies=False):
+    # type: (ProjectBuilder, str, List[str], bool) -> None
+    for dist in distributions:
+        if not skip_dependencies:
+            missing = builder.check_dependencies(dist)
+            if missing:
+                _error('Missing dependencies:' + ''.join(['\n\t' + dep for dep in missing]))
+
+        builder.build(dist, outdir)
+
+
+def build(srcdir, outdir, distributions, config_settings=None, isolation=True, skip_dependencies=False):
+    # type: (str, str, List[str], Optional[ConfigSettings], bool, bool) -> None
     '''
     Runs the build process
 
@@ -36,6 +56,7 @@ def build(srcdir, outdir, distributions, config_settings=None, skip_dependencies
     :param outdir: Output directory
     :param distributions: Distributions to build (sdist and/or wheel)
     :param config_settings: Configuration settings to be passed to the backend
+    :param isolation: Isolate the build in a separate environment
     :param skip_dependencies: Do not perform the dependency check
     '''
     if not config_settings:
@@ -44,13 +65,10 @@ def build(srcdir, outdir, distributions, config_settings=None, skip_dependencies
     try:
         builder = ProjectBuilder(srcdir, config_settings)
 
-        for dist in distributions:
-            if not skip_dependencies:
-                missing = builder.check_dependencies(dist)
-                if missing:
-                    _error('Missing dependencies:' + ''.join(['\n\t' + dep for dep in missing]))
-
-            builder.build(dist, outdir)
+        if isolation:
+            _build_in_isolated_env(builder, outdir, distributions)
+        else:
+            _build_in_current_env(builder, outdir, distributions, skip_dependencies)
     except BuildException as e:
         _error(str(e))
     except BuildBackendException as e:
@@ -136,7 +154,7 @@ def main(cli_args, prog=None):  # type: (List[str], Optional[str]) -> None
     if not distributions:
         distributions = ['sdist', 'wheel']
 
-    build(args.srcdir, args.outdir, distributions, config_settings, args.skip_dependencies)
+    build(args.srcdir, args.outdir, distributions, config_settings, not args.no_isolation, args.skip_dependencies)
 
 
 if __name__ == '__main__':  # pragma: no cover
--- a/build/__main__.py
+++ b/build/__main__.py
@@ -7,9 +7,8 @@ import traceback
 
 from typing import List, Optional
 
-import pep517.envbuild
-
 from . import BuildBackendException, BuildException, ConfigSettings, ProjectBuilder
+from .env import IsolatedEnvironment
 
 
 __all__ = ['build', 'main', 'main_parser']
@@ -30,8 +29,8 @@ def _error(msg, code=1):  # type: (str, int) -> None  # pragma: no cover
 
 
 def _build_in_isolated_env(builder, outdir, distributions):  # type: (ProjectBuilder, str, List[str]) -> None
-    with pep517.envbuild.BuildEnvironment() as env:
-        env.pip_install(builder.build_dependencies)
+    with IsolatedEnvironment() as env:
+        env.install(builder.build_dependencies)
         for distribution in distributions:
             builder.build(distribution, outdir)
 
--- a/build/env.py
+++ b/build/env.py
@@ -1,0 +1,93 @@
+import os
+import shutil
+import subprocess
+import sys
+import sysconfig
+import tempfile
+import types
+
+
+if False:  # TYPE_CHECKING  # pragma: no cover
+    from typing import Dict, Optional, Iterable, Type
+
+
+class IsolatedEnvironment(object):
+    '''
+    Isolated build environment context manager
+
+    Non-standard paths injected directly to sys.path still be passed to the environment.
+    '''
+
+    def __init__(self):  # type: () -> None
+        self._env = {}  # type: Dict[str, Optional[str]]
+
+    def _replace_env(self, key, new):  # type: (str, Optional[str]) -> None
+        if not new:  # pragma: no cover
+            return
+
+        self._env[key] = os.environ.get(key, None)
+        os.environ[key] = new
+
+    def _restore_env(self):  # type: () -> None
+        for key, val in self._env.items():
+            if val is None:
+                os.environ.pop(key, None)
+            else:
+                os.environ[key] = val
+
+    def _get_env_path(self, path):  # type: (str) -> Optional[str]
+        return sysconfig.get_path(path, vars=self._env_vars)
+
+    def __enter__(self):  # type: () -> IsolatedEnvironment
+        self._path = tempfile.mkdtemp(prefix='build-env-')
+        self._env_vars = {
+            'base': self._path,
+            'platbase': self._path,
+        }
+
+        sys_path = sys.path[1:]
+
+        remove_paths = os.environ.get('PYTHONPATH', '').split(os.pathsep)
+
+        for path in ('purelib', 'platlib'):
+            our_path = sysconfig.get_path(path)
+            if our_path:
+                remove_paths.append(our_path)
+
+            for scheme in sysconfig.get_scheme_names():
+                our_path = sysconfig.get_path(path, scheme)
+                if our_path:
+                    remove_paths.append(our_path)
+
+            env_path = self._get_env_path(path)
+            if env_path:
+                sys_path.append(env_path)
+
+        for path in remove_paths:
+            if path in sys_path:
+                sys_path.remove(path)
+
+        self._replace_env('PATH', self._get_env_path('scripts'))
+        self._replace_env('PYTHONPATH', os.pathsep.join(sys_path))
+        self._replace_env('PYTHONHOME', self._path)
+
+        return self
+
+    def __exit__(self, typ, value, traceback):
+        # type: (Optional[Type[BaseException]], Optional[BaseException], Optional[types.TracebackType]) -> None
+        if self._path and os.path.isdir(self._path):
+            shutil.rmtree(self._path)
+
+        self._restore_env()
+
+    def install(self, requirements):  # type: (Iterable[str]) -> None
+        '''
+        Installs the specified requirements on the environment
+        '''
+        if not requirements:
+            return
+
+        subprocess.check_call([sys.executable, '-m', 'ensurepip'], cwd=self._path)
+
+        cmd = [sys.executable, '-m', 'pip', 'install', '--ignore-installed', '--prefix', self._path] + list(requirements)
+        subprocess.check_call(cmd)
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -23,6 +23,10 @@ can use it:
    usage: python-build [-h] ...
 
 
+By default python-build will build the package in a isolated environment, but
+this behavior can be disabled with ``--no-isolation``.
+
+
 Mission Statement
 =================
 
@@ -67,9 +71,9 @@ with :pep:`517` support.
 ``python -m pep517.build``
 --------------------------
 
-``python -m pep517.build`` will invoke pip_ to install missing dependencies,
-``python-build`` is *stricly* a build tool, it does not do any sort of
-dependency management. If the dependencies are not met, the build will fail.
+``python-build`` implements a CLI tailored to end users. ``python -m
+pep517.build`` *"implements essentially the simplest possible frontend tool,
+to exercise and illustrate how the core functionality can be used"*.
 
 
 Custom Behaviors
--- a/tests/test_env.py
+++ b/tests/test_env.py
@@ -1,0 +1,30 @@
+# SPDX-License-Identifier: MIT
+
+import os
+import subprocess
+import sys
+import sysconfig
+
+import build.env
+
+
+def test_isolated_environment(mocker):
+    with build.env.IsolatedEnvironment() as env:
+        if os.name != 'nt':
+            assert os.environ['PATH'] == os.path.join(env._path, 'bin')
+        assert os.environ['PYTHONHOME'] == env._path
+
+        for path in ('purelib', 'platlib'):
+            assert sysconfig.get_path(path) not in os.environ['PYTHONPATH'].split(os.pathsep)
+
+        mocker.patch('subprocess.check_call')
+
+        env.install([])
+        subprocess.check_call.assert_not_called()
+
+        env.install(['some', 'requirements'])
+        if sys.version_info[:2] != (3, 5):
+            subprocess.check_call.assert_called()
+        assert subprocess.check_call.call_args[0][0] == [
+            sys.executable, '-m', 'pip', 'install', '--ignore-installed', '--prefix', env._path, 'some', 'requirements'
+        ]
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -5,7 +5,6 @@ import io
 import os
 import sys
 
-import pep517
 import pytest
 
 import build
@@ -82,9 +81,10 @@ def test_build(mocker):
     mocker.patch('build.ProjectBuilder.check_dependencies')
     mocker.patch('build.ProjectBuilder.build')
     mocker.patch('build.__main__._error')
-    mocker.patch('pep517.envbuild.BuildEnvironment.pip_install')
+    mocker.patch('build.env.IsolatedEnvironment.install')
 
     build.ProjectBuilder.check_dependencies.side_effect = [[], ['something'], [], []]
+    build.env.IsolatedEnvironment._path = mocker.Mock()
 
     # isolation=True
     build.__main__.build('.', '.', ['sdist'])
@@ -93,7 +93,7 @@ def test_build(mocker):
     # check_dependencies = []
     build.__main__.build('.', '.', ['sdist'], isolation=False)
     build.ProjectBuilder.build.assert_called_with('sdist', '.')
-    pep517.envbuild.BuildEnvironment.pip_install.assert_called_with(set(build._DEFAULT_BACKEND['requires']))
+    build.env.IsolatedEnvironment.install.assert_called_with(set(build._DEFAULT_BACKEND['requires']))
 
     # check_dependencies = ['something']
     build.__main__.build('.', '.', ['sdist'], isolation=False)
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -5,8 +5,10 @@ import io
 import os
 import sys
 
+import pep517
 import pytest
 
+import build
 import build.__main__
 
 
@@ -23,31 +25,32 @@ out = os.path.join(cwd, 'dist')
 @pytest.mark.parametrize(
     ('cli_args', 'build_args'),
     [
-        ([], [cwd, out, ['sdist', 'wheel'], {}, False]),
-        (['-s'], [cwd, out, ['sdist'], {}, False]),
-        (['-w'], [cwd, out, ['wheel'], {}, False]),
-        (['source'], ['source', out, ['sdist', 'wheel'], {}, False]),
-        (['-o', 'out'], [cwd, 'out', ['sdist', 'wheel'], {}, False]),
-        (['-x'], [cwd, out, ['sdist', 'wheel'], {}, True]),
+        ([], [cwd, out, ['sdist', 'wheel'], {}, True, False]),
+        (['-n'], [cwd, out, ['sdist', 'wheel'], {}, False, False]),
+        (['-s'], [cwd, out, ['sdist'], {}, True, False]),
+        (['-w'], [cwd, out, ['wheel'], {}, True, False]),
+        (['source'], ['source', out, ['sdist', 'wheel'], {}, True, False]),
+        (['-o', 'out'], [cwd, 'out', ['sdist', 'wheel'], {}, True, False]),
+        (['-x'], [cwd, out, ['sdist', 'wheel'], {}, True, True]),
         (
             ['-C--flag1', '-C--flag2'],
             [cwd, out, ['sdist', 'wheel'], {
                 '--flag1': '',
                 '--flag2': '',
-            }, False]
+            }, True, False]
         ),
         (
             ['-C--flag=value'],
             [cwd, out, ['sdist', 'wheel'], {
                 '--flag': 'value',
-            }, False]
+            }, True, False]
         ),
         (
             ['-C--flag1=value', '-C--flag2=other_value', '-C--flag2=extra_value'],
             [cwd, out, ['sdist', 'wheel'], {
                 '--flag1': 'value',
                 '--flag2': ['other_value', 'extra_value'],
-            }, False]
+            }, True, False]
         ),
     ]
 )
@@ -72,7 +75,6 @@ def test_prog():
             build.__main__.main(['--help'], prog='something')
 
 
-@pytest.mark.skipif(sys.version_info[:2] == (3, 5), reason='bug in mock')
 def test_build(mocker):
     open_mock = mocker.mock_open(read_data='')
     mocker.patch('{}.open'.format(build_open_owner), open_mock)
@@ -80,26 +82,33 @@ def test_build(mocker):
     mocker.patch('build.ProjectBuilder.check_dependencies')
     mocker.patch('build.ProjectBuilder.build')
     mocker.patch('build.__main__._error')
+    mocker.patch('pep517.envbuild.BuildEnvironment.pip_install')
 
     build.ProjectBuilder.check_dependencies.side_effect = [[], ['something'], [], []]
 
-    # check_dependencies = []
+    # isolation=True
     build.__main__.build('.', '.', ['sdist'])
-    build.ProjectBuilder.build.assert_called()
+    build.ProjectBuilder.build.assert_called_with('sdist', '.')
 
-    # check_dependencies = ['something]
-    build.__main__.build('.', '.', ['sdist'])
-    build.__main__._error.assert_called()
+    # check_dependencies = []
+    build.__main__.build('.', '.', ['sdist'], isolation=False)
+    build.ProjectBuilder.build.assert_called_with('sdist', '.')
+    pep517.envbuild.BuildEnvironment.pip_install.assert_called_with(set(build._DEFAULT_BACKEND['requires']))
+
+    # check_dependencies = ['something']
+    build.__main__.build('.', '.', ['sdist'], isolation=False)
+    build.ProjectBuilder.build.assert_called_with('sdist', '.')
+    build.__main__._error.assert_called_with('Missing dependencies:\n\tsomething')
 
     build.ProjectBuilder.build.side_effect = [build.BuildException, build.BuildBackendException]
     build.__main__._error.reset_mock()
 
     # BuildException
     build.__main__.build('.', '.', ['sdist'])
-    build.__main__._error.assert_called()
+    build.__main__._error.assert_called_with('')
 
     build.__main__._error.reset_mock()
 
     # BuildBackendException
     build.__main__.build('.', '.', ['sdist'])
-    build.__main__._error.assert_called()
+    build.__main__._error.assert_called_with('')
--- a/tests/test_projectbuilder.py
+++ b/tests/test_projectbuilder.py
@@ -20,9 +20,9 @@ else:  # pragma: no cover
     email_message_from_string = importlib_metadata._compat.email_message_from_string
 
 if sys.version_info >= (3,):  # pragma: no cover
-    build_open_owener = 'builtins'
+    build_open_owner = 'builtins'
 else:  # pragma: no cover
-    build_open_owener = 'build'
+    build_open_owner = 'build'
     FileNotFoundError = IOError
     PermissionError = OSError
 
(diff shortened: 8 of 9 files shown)


FFY00:
@pganssle @gaborbernat could you review?

FFY00:
I choose `--no-isolation` over something like `--isolation pip` as I don't think we will be providing different isolation types. I don't really think it makes sense here. But if that happens in the future we can still keep `--no-isolation` as an equivalent of `--isolation none` for backward compatibility.

README.md
```diff
@@ -17,6 +17,7 @@ optional arguments:
                         output directory
   --skip-dependencies, -x
                         does not check for the dependencies
+  --no-isolation, -n    do not isolate the build in a virtual environment
```
gaborbernat, line 20: We should probably not show it here for maintenance reasons, and move it to the readme docs entirely via https://sphinx-argparse.readthedocs.io/en/stable/

build/__main__.py
```diff
@@ -80,6 +80,9 @@ def main(cli_args):  # type: (List[str]) -> None
     parser.add_argument('--skip-dependencies', '-x',
                         action='store_true',
                         help='does not check for the dependencies')
+    parser.add_argument('--no-isolation', '-n',
+                        action='store_true',
+                        help='do not isolate the build in a virtual environment')
```
gaborbernat, line 85: would be worthwhile to mention that will raise error if build dependencies are not specified; ``do not create an isolated build environment (raise error when host is missing build requirements)`` 🤔

build/__main__.py
```diff
@@ -27,33 +29,31 @@ def _error(msg, code=1):  # type: (str, int) -> None  # pragma: no cover
...
+        :param outdir: Output directory
+        :param distributions: Distributions to build (sdist and/or wheel)
+        :param skip_dependencies: Do not perform the dependency check
+        '''
+        try:
+            for dist in distributions:
+                if not skip_dependencies:
+                    missing = self.check_depencencies(dist)
+                    if missing:
+                        _error('Missing dependencies:' + ''.join(['\n\t' + dep for dep in missing]))
```
gaborbernat, line 46: These checks seem redundant for the isolated builds 🤔 They can only happen with no isolation, not?

docs/source/index.rst
```diff
@@ -38,6 +39,10 @@ can use it:
    usage: python-build [-h] ...
 
 
+By default python-build will build the package in a virtual environment, but
```
gaborbernat, line 42: considering we're not creating virtual environments it's better to say, by default will build the package in an isolated environment.

tests/test_main.py
```diff
@@ -19,21 +20,36 @@
...
+    mocker.patch('build.__main__.MainBuilder.get_dependencies')
+    mocker.patch('build.__main__.MainBuilder.build_distributions')
 
     build.__main__.main(cli_args)
-    build.__main__.build.assert_called_with(*build_args)
+    build.__main__.MainBuilder.build_distributions.assert_called_with(*build_args)
+    build.__main__.MainBuilder.__init__.assert_called_with(*init_args)
+
+
+@pytest.mark.skipif(sys.version_info[:2] == (3, 5), reason='bug in mock')
```
gaborbernat, line 43: generally, I tend to put here a link to the bug report 🤔 because other people will not know what you're talking about, without 🤔

build/__main__.py
FFY00: I would go with "requires build dependencies to be present". What do you think?

gaborbernat: Works for me 👍

tests/test_main.py
FFY00: I was too lazy to do that :P. Anyway, it is actually not a bug, I was too quick to judge the backtrace. It is fixed now.

README.md
FFY00: I am having some trouble getting it to find the parser, so I am gonna do it in a separate PR.

FFY00:
I refactored the patch based on the comments, should be better now.

.github/workflows/build.yml
```diff
@@ -34,7 +34,7 @@ jobs:
       run: pip install poetry
 
     - name: Build python-build
-      run: python -m build -x
+      run: python -m build -x -n
```
pganssle, line 37: Not really related to this change, but what is the purpose of this test? Who would want to build this package by manually using `pip` to install its build-time dependencies, then building with isolation disabled?

The kind of people who would want build isolation would presumably want to use the no-dependencies / minimal dependencies bootstrap mode, rather than installing `poetry` and `typing` and such, no?

Also, this file in general seems a bit fragile. It's hard-coding the build-time dependencies of a bunch of projects rather than pulling those dependencies from the respective projects' `build-system.requires` file.

Anyway, just hopefully some constructive criticism, not anything that needs to block this particular set of changes.

build/__init__.py
```diff
@@ -106,6 +106,10 @@ def __init__(self, srcdir='.'):  # type: (str) -> None
         self.hook = pep517.wrappers.Pep517HookCaller(self.srcdir, self._backend,
                                                      backend_path=self._build_system.get('backend-path'))
 
+    @property
+    def build_dependencies(self):  # type: () -> Set[str]
+        return set(self._build_system.get('requires', []))
```
pganssle, line 111: Hm... I am a bit surprised by this. I suppose it's fair, but I was originally thinking that *any* missing fields in the `[build-system]` table would come from the defaults, in which case you should be able to use `self._build_system["requires"]`).

The way it is now, you supply a default value if no `[build-system]` table is provided, but if `[build-system]` is only partially specified, you leave it fully unspecified.

The thing is, there was a decently long era (6 months-1 year) where `pip` supported PEP 518 but not PEP 517, and there are some people who only care about PEP 518 and not PEP 517, so I wouldn't be surprised if there are a bunch of projects looking like this:

```toml
[build-system]
requires = ["Cython >= 28.0", "setuptools >= 48.0.3", "wheel"]
```

Or something of that nature, which would leave the build-backend unspecified! I recognize that it's a judgement call, but I was thinking that instead of [this code for setting the defaults](https://github.com/FFY00/python-build/blob/8db9c1110fb1d965d95326daf06a96a136e76d6b/build/__init__.py#L88), you'd want something more like this:

```python
self._build_system = {
    'build-backend': 'setuptools.build_meta:__legacy__',
    'requires': [
        'setuptools >= 40.8.0',
        'wheel'
    ]
}
self._build_system.update(self._spec.get('build-system', {}))
```

Though probably it's safe to figure out exactly what `pip` does and do that.

.github/workflows/build.yml
FFY00: With `-n`, it does not perform any isolation and does not use `pip` to install the dependencies.

This test just makes sure that the `python -m build` actually works. We mock too much in `pytest` and I feel much more comfortable having this extra tests. I Selected 4 packages (`python-build`, `dateutil`, `pip` and `Solaar`), to verify the command actually works, this particular patch changes two of them to the no isolation mode. Building `python-build` via poetry and without isolation does not have any particular meaning.

And yes, it would be safer to fetch `build-system.requires`. It should come in a separate PR. Just a thought but would it make sense to introduce a command to read the dependencies?

tests/test_main.py
```diff
@@ -18,15 +19,25 @@
 out = os.path.join(cwd, 'dist')
 
 
+_DEFAULT_BACKEND = {
```
pganssle, line 22: Not really important, but it might be a good idea to make this a fixture or the result of a function call, so that you can make a fresh copy each time, since dictionaries are mutable (not that you mutate it, but it would be annoying to track down a bug if you accidentally started mutating it).

FFY00: Yes. I did not do it because we only use in 1 place, so I thought it was better to not over optimize. If we need it in another place we can make it a fixture :)

.github/workflows/build.yml
pganssle: > With `-n`, it does not perform any isolation and does not use `pip` to install the dependencies.

Yeah, I meant that you manually use `pip` to install the dependencies. For `python-dateutil`, `Solaar` and `pip` I could see that being a reasonable thing to do, but the only people who would ever want to do this would be distro packagers, and for `python-build` specifically, they're unlikely to want to because it's intended to be one of the first things you bootstrap.

pganssle:
I think overall this PR is good, but I think it's a bit under-tested considering the heavy use of mocks.

I've never found it terribly easy to test CLI applications, but I think I would get rid of `.github/workflows/build.yml` and move that whole test into some parameterized `pytest` tests that call the function via `subprocess` or another system call.

I also think that it would be better to use fake packages designed to test a single thing, rather than real packages that you pull from PyPI (though presumably — at least to start with — you can still hit PyPI during the tests anyway). You can see we do [something like that in setuptools](https://github.com/pypa/setuptools/blob/74de4e985eda49e38ece5805e05197dd4d2d9c8a/setuptools/tests/test_build_meta.py#L66-L129); in that case, we define the files in code and then construct the package in a temporary location as part of the test, but you could just as easily have a "tests/packages" directory or something with a bunch of stub packages in it.

build/__init__.py
FFY00: Where does `Cython >= 28.0` come from? Is it tied to the `setuptools` requirement? If so, I don't believe we should default to that.

When `build-backend` is missing we default to:
```toml
[build-system]
requires = ['setuptools >= 40.8.0', 'wheel']
build-backend = 'setuptools.build_meta:__legacy__'
```

But I believe it's wrong to default to any `requires` **if** `build-backend` is specified. I could have an in-tree backend and not specify any `requires`, in this case I wouldn't expect a user to add `requires = []`.

We could select the default value for `requires` based on `build-backend`. For `build-backend = 'setuptools.build_meta:__legacy__'` we know that `requires = ['setuptools >= 40.8.0', 'wheel']`. But I do not believe this would be the right approach, it's a slippery slope, who decides which backends we will and will not provide default values for?

For the other way round, I am not sure. I would be okay with defaulting `build-backend` to `setuptools.build_meta:__legacy__` and appending `['setuptools >= 40.8.0', 'wheel']` to ``requires.

FFY00:
`.github/workflows/build.yml` are just sanity checks to make sure we can actually build real projects, they are not meant as a replacement for other tests. I am not sure if they should be in `pytest`, especially since they do take a bit of time to run. If anything, we could add them in nox? But since they are *just* sanity tests, I feel like they are better fit for the CI.

I think adding proper tests with fake packages is great, it's something that I have in my TODO.

build/__init__.py
pganssle: > Where does `Cython >= 28.0` come from? Is it tied to the `setuptools` requirement? If so, I don't believe we should default to that.

Sorry, I should have been more clear. There are a bunch of packages out there that specify `build-system.requires` but not `build-system.build-backend`, because if you don't specify `build-system.build-backend`, `pip` will default to `setuptools`. Some people deliberately don't set it because they want `__legacy__`, and other people just never set it because they set up their `pyproject.toml` before `pip` had any support for the `build-backend` key.

> But I believe it's wrong to default to any `requires` **if** `build-backend` is specified. I could have an in-tree backend and not specify any `requires`, in this case I wouldn't expect a user to add `requires = []`.

This is a matter of taste, I suppose.  I think that you *are* supposed to explicitly specify `requires = []` and the behavior is undefined-ish if you don't. PEP 518 says that if the table exists but the `requires` field is missing the tool [should consider it an error](https://www.python.org/dev/peps/pep-0518/#build-system-table).

I think `requires = []` is kind of a bad default, because almost no one has in-tree backends, and there's actually [a separate mechanism for declaring that you have one](https://www.python.org/dev/peps/pep-0517/#in-tree-build-backends), so anyone who hasn't specified the `backend-path` field *must* have dependencies.

That said, it is definitely more important to have a default value for an unspecified `build-backend` if `requires` is specified than to have a default value for `requires` if `build-backend` is specified. I think if `build-backend` is specified without `requires`, the reasonable things to do are to default to the normal default value of `requires` or to throw an exception. The arguments in favor of choosing the "populate a default and use `.upgrade`"-style strategy are:

1. A lot of tools just assume that `setuptools` and `wheel` are present, even now. People who need them might fail to specify it, but people who need anything *else* will definitely specify it.
2. If the person actually wanted `requires=[]`, it probably doesn't actively break their build to also install `setuptools` and `wheel` (and if it does, they can add `requires=[]`), so it's a much softer kind of failure.
3. It's easier for end users to understand the defaulting behavior if it follows a simple rule: "Anything missing from the `build-system` table is taken from these defaults." is much easier to understand than, "If you specify `requires` then `build-backend` comes from this default, but if you specify `build-backend` then `requires` is something different, and if the table is missing entirely it's these default values".

tests/test_main.py
pganssle: I would argue that you should probably put it in-line in the single test that you're using rather than at module level if you are relying on the fact that it's only used in one place. You can always refactor it out into a fixture or function later if you need the same thing, and "future you" (or me, or some random contributor) is not going to know the history of this default value, and during review the implications of "we are re-using a mutable data structure" may be missed.

build/__init__.py
FFY00: > This is a matter of taste, I suppose. I think that you _are_ supposed to explicitly specify `requires = []` and the behavior is undefined-ish if you don't. PEP 518 says that if the table exists but the `requires` field is missing the tool [should consider it an error](https://www.python.org/dev/peps/pep-0518/#build-system-table).

Then I think we should consider it an error :)
I would like to be as strict as possible to avoid leaking implementation details. The problem is people complaining "it works in `pip` but doesn't work with `build`". The project description says we are a "*correct* PEP517 package builder" so I think it makes sense to go down that route, what do you think?

> I think if `build-backend` is specified without `requires`, the reasonable things to do are to default to the normal default value of `requires` or to throw an exception.

I agree. I vote to throw an exception.

tests/test_main.py
FFY00: That is fair :)

FFY00: Well, actually I could put it in the package.

gaborbernat:
> `.github/workflows/build.yml` are just sanity checks to make sure we can actually build real projects, they are not meant as a replacement for other tests. I am not sure if they should be in `pytest`, especially since they do take a bit of time to run. If anything, we could add them in nox? But since they are *just* sanity tests, I feel like they are better fit for the CI.
> 
> I think adding proper tests with fake packages is great, it's something that I have in my TODO.

IMHO they should be part of the test suite. They are integration tests, and as such should be marked as such in the tests, giving you the option to not run them if you want a fast test.

FFY00:
Okay, I can add a `--run-integration` flag to pytest to opt in.

gaborbernat:
What I tend to do is to make it auto on by default for nox, but disabled if you just invoke pytest. If you check tox/virtualenv you'll find exact how to of this.

FFY00:
I agree with this :). On by default in nox but off in pytest.

build/__main__.py
```diff
@@ -80,6 +90,9 @@ def main(cli_args):  # type: (List[str]) -> None
     parser.add_argument('--skip-dependencies', '-x',
                         action='store_true',
                         help='does not check for the dependencies')
+    parser.add_argument('--no-isolation', '-n',
```
gaborbernat, line 93: The ``-n`` feels hard to understand, considering would be used rarely manually should we just provide ``--no-isolation`` and no short form, to remove confusion on the caller side?

tests/test_main.py
```diff
@@ -21,12 +23,13 @@
...
-        (['-s'], [cwd, out, ['sdist'], False]),
-        (['-w'], [cwd, out, ['wheel'], False]),
-        (['source'], ['source', out, ['sdist', 'wheel'], False]),
-        (['-o', 'out'], [cwd, 'out', ['sdist', 'wheel'], False]),
-        (['-x'], [cwd, out, ['sdist', 'wheel'], True]),
+        ([],            [cwd, out, ['sdist', 'wheel'], True, False]),
+        (['-n'],        [cwd, out, ['sdist', 'wheel'], False, False]),
+        (['-s'],        [cwd, out, ['sdist'], True, False]),
+        (['-w'],        [cwd, out, ['wheel'], True, False]),
+        (['source'],    ['source', out, ['sdist', 'wheel'], True, False]),
```
gaborbernat, line 30: Not sure I like this indentation technique. Looks ok now, but as soon as we add another line that's longer than the current longest we have to reindent all other lines which feel to me carries more risk than worth it (more chance of merge conflict, more lines changed than needed and potentially doing unintended changes).

build/__main__.py
```diff
@@ -27,25 +29,33 @@ def _error(msg, code=1):  # type: (str, int) -> None  # pragma: no cover
...
 
-        for dist in distributions:
-            if not skip_dependencies:
-                missing = builder.check_depencencies(dist)
-                if missing:
-                    _error('Missing dependencies:' + ''.join(['\n\t' + dep for dep in missing]))
-
-            builder.build(dist, outdir)
+        if isolation:
+            with pep517.envbuild.BuildEnvironment() as env:
```
pganssle, line 47: I just had a conversation that made me realize that this may actually not be sufficient to achieve proper isolation. I was under the impression that `BuildEnvironment()` would create an *isolated* build environment, but it actually adds a layer on top of your existing environment. For example, if I create a project with the following `setup.py`:

```python
from setuptools import setup

try:
    import attr
    raise ValueError("Attrs should not be available!")
except ImportError:
    pass

setup()
```

If I run `python -m build` on the project without installing `attrs` it works fine, but if I `pip install attrs` into the same environment that `build` is installed in, it raises the exception.

That's especially troubling because `build` itself has runtime dependencies, so any undeclared dependencies on `packaging`, `pep517`, `pyparsing`, `toml`, etc, would go unnoticed.

I don't know if we should try and fix this here or in `pep517`. I assume @gaborbernat will have a better understanding of the options than I do.

gaborbernat: AFAIK (@takluyver would know for sure) it's the caller responsibility to ensure that on the host python only the standard library is available. So I'd say the responsibility would be on this lib to fix this. In practice, this would imply that you would need to remove from ``sys.path`` all paths that are not stdlib (you can get this via https://docs.python.org/3/library/sysconfig.html#sysconfig.get_path and passing in stdlib).

pganssle: "The caller" as in the user of `python-build`, or "The caller" as in `python-build`?

I think the way it is now we're basically doing the equivalent of building a virtualenv with `--system-site-packages`, which is not a good default, so I feel strongly that we should *not* be doing that. How does `virtualenv` achieve this?

I suppose this explains why `pip` has their own logic for build isolation and doesn't directly use `pep517`'s. We may want to copy them (though their approach does bring in [at least one problem](https://github.com/pypa/pip/issues/6264)).

gaborbernat: > How does `virtualenv` achieve this?

It knows the interpreter deeply to be able to tell how to achieve. There's no single global answer, various interpreters do it differently. Here's a list https://github.com/pypa/virtualenv/blob/master/src/virtualenv/discovery/py_info.py#L35-L103 of options that can influence things. Either pick pips way or pull in virtualenv package to create a virtual environment the way I see it. I agree we should not pull in global site package.

tests/test_main.py
FFY00: That is fair. We can put them on different lines, what do you think?

```python
(
    ['-n'],
    [cwd, out, ['sdist', 'wheel'], False, False],
)
```

build/__main__.py
FFY00: It would be rarely used by users but a lot of packagers will still need to use it manually. A quick query on the arch repos shows that I maintain ~90 python packages, that means I will have to write `--no-isolation` a lot when migrating from `setup.py` calls. Anyway, I think exposing `-n` to users is a fair price, anyone can still use `--help` to find out what it is. Would you be okay keeping it, or is it something you feel strongly about?

build/__main__.py
takluyver: We're thinking of removing `BuildEnvironment` from `pep517` (along with all the higher-level code that uses it) - it's not very good, as you've noticed, and it distracts people from the useful part of the pep517 package. See https://github.com/pypa/pep517/pull/83

I wrote `BuildEnvironment` and the code using it as a kind of example of the simplest possible PEP 517 frontend that puts all the pieces together (installing dependencies, calling the hooks with those dependencies available). The spec only says that the build environment SHOULD be isolated, so that 'overlay' model is technically compliant, but it's not best practice.

FFY00: I think having this high level code would be in-scope for this project, we just need to be careful with what we choose to export as public API.

So I propose for now to move `BuildEnvironment` here as private API, and then in the future think about exporting it.

> I wrote `BuildEnvironment` and the code using it as a kind of example of the simplest possible PEP 517 frontend that puts all the pieces together (installing dependencies, calling the hooks with those dependencies available).

I have expressed this opinion previously, but I'll reiterate as I think it's relevant. Nothing you consider an "example" should be ever exported publicly, being this API, CLI, etc. Once you make something public, people will use it. Providing code as an example forces people to copy it, tailor it to their needs, and most importantly, lets you off the hook :smile:.

takluyver: Yeah, I'm right with you there. I must have been feeling rather gung-ho when I decided to make it part of the library. ;-)

build/__main__.py
gaborbernat: I would have thought machines/scripts  to invoke this build not you manually. No strong feelings.

tests/test_main.py
gaborbernat: I'm happy to keep it one line which is what black also would do as long as it fits in one line.

build/__main__.py
gaborbernat: I don't think the issue here was if it's private or not. The issue is that it's not working, so we should use something else, which can be private. I'd say perhaps virtualenv, but I'm happy with venv too.

FFY00: I don't think the current `pep517` code should be hard to fix up, we just need to adjust `sys.path`, no? I think `virtualenv` would be the direction forward if we needed to deal with custom interpreters, but we don't. I would like to avoid adding a new dependency.

build/__main__.py
FFY00: I usually start from a template with the commands already there, but not always. I think a few people would be against adding macros that abstract the commands because, well, they abstract the commands. And we can't be fully automatic either, we prefer to get sources from github than PyPI, which is not as easy to automate and a lot of the time people would still need to do manually steps.

build/__init__.py
FFY00: This should be resolved, @pganssle can you check if the implementation is okay?

build/__main__.py
gaborbernat: > I don't think the current `pep517` code should be hard to fix up, we just need to adjust `sys.path`, no?

Well sure, in theory, you *just** need to fix the sys.path; however, in practice, it's non-trivial how to fix it up. Things differ per python implementation what needs to stay and what needs to go. virtualenv contains this complex logic. Hence why suggesting we use the slower venv on python 3, as that does not pull in any dependency 🤷

FFY00:
I implemented a fully isolated environment. The import paths are derived from `sys.path`, we filter out everything we don't want (system `{pure,plat}lib` and `PYTHONPATH`). If someone manually injects a non-standard path to `sys.path`, we have no way to filter it, so it is kept. This behavior is documented.

I also tried on pypy and it works as expected.

tests/test_env.py
```diff
@@ -0,0 +1,29 @@
...
+import subprocess
+import sys
+import sysconfig
+
+import build.env
+
+
+def test_isolated_environment(mocker):
+    with build.env.IsolatedEnvironment() as env:
+        assert os.environ['PATH'] == os.path.join(env._path, 'bin')
```
gaborbernat, line 13: bin, what about windows where it's scripts?

FFY00: We get the scripts path from `sysconfig`, so it should be right. We are only testing on Linux right now, so I missed this.

gaborbernat: We really should test on all platforms: Linux, Windows, macOs. There's no working around this if we want this to move under PyPa, so I consider this major blocking.

FFY00: I agree, I will do it in a future PR.

FFY00:
Codecov is having server troubles again. The Python 2 test coverage failed to send so the coverage doesn't show up as 100%.

If this keeps happening we might have to move to a different service, like https://coveralls.io.

build/env.py
```diff
@@ -0,0 +1,98 @@
...
+import tempfile
+import types
+
+
+if False:  # TYPE_CHECKING  # pragma: no cover
+    from typing import Dict, Optional, Iterable, Type
+
+
+class IsolatedEnvironment(object):
+    '''
```
pganssle, line 15: Is this `'''` style already consistent across the whole package? It's very unusual to use single-quotes for docstrings, even among code-bases that exclusively use single-quotes for strings (I even have them highlighted differently from `'''` strings in my syntax highlighter).

gaborbernat: +1 on this

FFY00: It is. I am open to change it, but it should come in a separate PR.

FFY00:
I re-triggered the tests, codecov is now fixed.

build/env.py
```diff
@@ -0,0 +1,98 @@
...
+
+            env_path = self._get_env_path(path)
+            if env_path:
+                sys_path.append(env_path)
+
+        for path in remove_paths:
+            if path in sys_path:
+                sys_path.remove(path)
+
+        self._replace_env('PATH', self._get_env_path('scripts'))
```
gaborbernat, line 75: Any reason why you try to alter the current environment, instead of just cloning the current environ, pass that in to subprocess calls?

build/env.py
```diff
@@ -0,0 +1,98 @@
...
+        self._env_vars = {
+            'base': self._path,
+            'platbase': self._path,
+        }
+
+        sys_path = sys.path[1:]
+
+        remove_paths = os.environ.get('PYTHONPATH', '').split(os.pathsep)
+
+        if os.name == 'nt':  # pragma: no cover
```
gaborbernat, line 52: do we need to define these? can't we acquire it from sysconfig?

build/__init__.py
```diff
@@ -92,16 +101,14 @@ def __init__(self, srcdir='.', config_settings=None):  # type: (str, Optional[Co
...
-            self._build_system = self._spec['build-system']
-        except KeyError:
-            self._build_system = {
-                'build-backend': 'setuptools.build_meta:__legacy__',
-                'requires': [
-                    'setuptools >= 40.8.0',
-                    'wheel'
-                ]
-            }
+        self._build_system = self._spec.get('build-system', _DEFAULT_BACKEND)
```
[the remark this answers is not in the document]
FFY00: No, it's just to make `_DEFAULT_BACKEND` available for the tests to use. Could be in a different commit sure, let me know if you want that.

build/env.py
```diff
@@ -0,0 +1,98 @@
...
+import shutil
+import subprocess
+import sys
+import sysconfig
+import tempfile
+import types
+
+
+if False:  # TYPE_CHECKING  # pragma: no cover
+    from typing import Dict, Optional, Iterable, Type
```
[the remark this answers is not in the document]
FFY00: What do you mean? `# TYPE_CHECKING` will proceed with the `if` if type checking is active.

It is guarded here because `Type` isn't in 3.5.0 and 3.5.1.

build/env.py
FFY00: Because we want everything inside the context manager to get this environment, not just the pip call. If we don't alter the environment the `pep517` hook calls won't be isolated for eg.

build/env.py
FFY00: Yes, but I don't want to add a test that is just same the logic.

build/__main__.py
```diff
@@ -136,7 +153,7 @@ def main(cli_args, prog=None):  # type: (List[str], Optional[str]) -> None
     if not distributions:
         distributions = ['sdist', 'wheel']
 
-    build(args.srcdir, args.outdir, distributions, config_settings, args.skip_dependencies)
+    build(args.srcdir, args.outdir, distributions, config_settings, not args.no_isolation, args.skip_dependencies)
```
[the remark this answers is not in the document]
FFY00: I disagree here. I avoid using keyword arguments when the variable names are already pretty self descriptive. I find that doing that generated a lot of visual noise.

```
..., config_settings=config_settings, isolation=not args.no_isolation, skip_dependencies=args.skip_dependencies)
```

build/__init__.py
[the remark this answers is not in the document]
gaborbernat: Maybe tests should replicate the code, not import it? Or have it as a before PR 🤔 IMHO this change makes sense, but should not be in this PR.

build/__main__.py
```diff
@@ -27,15 +28,34 @@ def _error(msg, code=1):  # type: (str, int) -> None  # pragma: no cover
...
+            builder.build(distribution, outdir)
+
+
+def _build_in_current_env(builder, outdir, distributions, skip_dependencies=False):
+    # type: (ProjectBuilder, str, List[str], bool) -> None
+    for dist in distributions:
+        if not skip_dependencies:
+            missing = builder.check_dependencies(dist)
+            if missing:
+                _error('Missing dependencies:' + ''.join(['\n\t' + dep for dep in missing]))
```
gaborbernat, line 44: instead of addition consider using ``'{}{}'.format``? ( I know I don't like + for string addition, makes it hard to read and type inspect)

build/__main__.py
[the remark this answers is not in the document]
gaborbernat: no strong feelings, just suggestion; note if you do full keyword args you should separate in separate lines which then makes it easy to read that they match up correctly visually)

build/env.py
gaborbernat: At the very least should be global. What I was saying though is that will OS always contain this? pypy for example defines its own schemes and any python distribution has a right to define so. IMHO we should always use sysconfig one.

build/env.py
gaborbernat: Maybe put these three in a dictionary and write a for loop then to make it easy to read and extend.

FFY00: IMO that's just more code I'll have to read when looking at the code. I will have to see that we are creating a dictionary, remember what is there and then see what we do to the dictionary.

Right now we can see directly what we are doing.

If you have strong feelings about this I can change though, I don't.

build/env.py
FFY00: > At the very least should be global. What I was saying though is that will OS always contain this? pypy for example defines its own schemes and any python distribution has a right to define so.

I do not follow.

build/__init__.py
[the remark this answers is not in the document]
FFY00: > Maybe tests should replicate the code, not import it?

That means that when we change the backend the tests will fail, and then we need to go there and manually change it. I do not have strong feeling about this, I can change if you really want to.

build/env.py
gaborbernat: I gather you took this list from https://docs.python.org/3/library/sysconfig.html#installation-paths but that's only true for CPython as distributed by the CPython core developer team. The scheme list may be extended and often is by other distributions (such as Debian repackaging), or other python implementations, see here for pypy https://foss.heptapod.net/pypy/pypy/-/blob/branch/default/lib-python/2.7/sysconfig.py#L29-38 

So you should use https://docs.python.org/3/library/sysconfig.html#sysconfig.get_scheme_names

FFY00: Ah, yes. Github doesn't show the full code and I thought this was other line :joy:.

FFY00:
The integration tests already run in the CI, and they are passing. This PR is already big, I don't think adding more stuff is a good idea. I see no reason to block it due to the integration tests. IMO we should merge it and then work on the integration tests. I would definitely block a release though.

gaborbernat:
> The integration tests already run in the CI, and they are passing.

I have a distinct feeling that they are passing by chance, not by design; because how they are set up (e.g. the reason why a sdist is unusable on the master, but the CI still passes). I'm not saying you need to do more in this PR. I'm saying we should not merge until we have the infrastructure in place to fully validate that what you made is 100% right.

tests/test_env.py
```diff
@@ -0,0 +1,30 @@
...
+import os
+import subprocess
+import sys
+import sysconfig
+
+import build.env
+
+
+def test_isolated_environment(mocker):
+    with build.env.IsolatedEnvironment() as env:
```
gaborbernat, line 12: I think this test should run in an isolated python environment (one that has no pip/setuptools/etc installed) so there's really no chance things are being pulled from some other path on the python path. E.g. is this past failing if you change from ``sys. get_scheme_names`` to that manual thing you defined before? It should if it's an accurate test.

FFY00: That "manual thing" I had before was hardcoding the scheme names, it should work the same. Why would this be failing?

> really no chance things are being pulled from some other path on the python path

What things being pulled? There is nothing being pulled. In this test I check if the environment is correct, after that I check if pip is called correctly, but we never actually call pip. I look at the path and make sure it is correct.

For eg. I import `sysconfig` before entering the isolated environment and make sure `purelib` or `platlib` are not there.

```python
for path in ('purelib', 'platlib'):
    assert sysconfig.get_path(path) not in os.environ['PYTHONPATH'].split(os.pathsep)
```

gaborbernat: > I had before was hardcoding the scheme names, it should work the same.

It should have been failing on pypy, as pypy uses different scheme names then what you hardcoded. The fact that it did not, shows something is not totally correct in this sense. 

> For eg. I import `sysconfig` before entering the isolated environment and make sure `purelib` or `platlib` are not there.

To be fair I'm not sure this is enough. I'd expect we'd isolate from any potential PYTHONPATH the user might have set, or additional paths the user injects via user/sitecustomize.

FFY00: It shouldn't.

```
>>>> import sysconfig
>>>> sysconfig.get_path('purelib')
'/opt/pypy/site-packages'
>>>> sysconfig.get_path('platlib')
'/opt/pypy/site-packages'
```

If you read the code you'll notice I remove both the default paths and the different schemes.

https://github.com/FFY00/python-build/blob/e5efbcca2df30b7de5e14477cdbd4fb83f1d932c/build/env.py#L55

The test is correct and working as expected.

FFY00:
The integration tests are definitely not passing by chance. I agree that we should have more of them but I keep my position this should not block this PR. We have unit tests which should fail when someone changes the behavior, and we have integration tests that build both this project and pip. To me, this should be enough to merge this PR.

build/__main__.py
FFY00: This should come in a separate PR, I just move the code to a different place.

FFY00:
@pganssle can you have a look?

tests/test_projectbuilder.py
```diff
@@ -148,26 +146,53 @@ def test_check_dependencies(mocker):
...
+
+
+def test_missing_backend(mocker, pyproject_no_backend_mock):
+    mocker.patch('importlib.import_module')
+    mocker.patch('{}.open'.format(build_open_owner), pyproject_no_backend_mock)
+    mocker.patch('pep517.wrappers.Pep517HookCaller')
+
+    builder = build.ProjectBuilder()
+
+    assert builder._build_system == build._DEFAULT_BACKEND
```
pganssle, line 189: I am not really an expert in testing CLIs, so maybe this is the most expedient way to do this, but my general rule for tests is that I only test the public API, so anything that's making assertions about the internal private structure is usually off-limits (or I mark it with a comment or something to indicate why this is the only / best way to test it).

Put another way, if you did an otherwise safe refactoring of the `ProjectBuilder` class, you would need to change this test, but that defeats one of the purposes of tests — allowing you to refactor with confidence, because you've made assertions about the behavior that should remain invariant.

Of course, I rarely have to test CLIs, so I think the rules may be a little different there (when there's no public API available from Python *anyway*).

gaborbernat:
> Is there anything particularly controversial or any major decisions that I'm glossing over here? I can probably form opinions on individual questions a lot easier than I can grok the entire change.

IMHO it's a matter of taste at the end of the day. Looking through the code I'm not confident that we have a good enough test suite to validate the approach here. So I was advocating for more tests; ideally on top of moving our current integration tests from CI to pytest. @FFY00 wants to move ahead as is now. The current state at a glance looks ok, so I'll not block it, though the tests we have does not give me enough confidence to approve it either.

tests/test_projectbuilder.py
FFY00: > I am not really an expert in testing CLIs, so maybe this is the most expedient way to do this, but my general rule for tests is that I only test the public API, so anything that's making assertions about the internal private structure is usually off-limits (or I mark it with a comment or something to indicate why this is the only / best way to test it).

We are not testing CLI here, we are testing `ProjectBuilder`, which is public API. python-build provides a CLI but also a minimal API to build projects, which is `ProjectBuilder`.

> Put another way, if you did an otherwise safe refactoring of the `ProjectBuilder` class, you would need to change this test, but that defeats one of the purposes of tests — allowing you to refactor with confidence, because you've made assertions about the behavior that should remain invariant.

If I do a safe refactoring of `ProjectBuilder` these tests should not fail. They might break if I change `ProjectBuilder`'s public API.

pganssle: > If I do a safe refactoring of `ProjectBuilder` these tests should not fail. They might break if I change `ProjectBuilder`'s public API.

Well, if you removed or renamed `build._DEFAULT_BACKEND` or `ProjectBuilder._build_backend`, these tests would fail, but presumably those are not part of the public interface.

It seems like our public interface exposes a `hook` attribute, which would be the `PEP517HookCaller` object, and that *does* expose a `.build_backend` attribute, publicly.

I think you'd need to ease off the mocking a bit, but you can define a top-level `DEFAULT_BACKEND` in this test file, and then do:

```python
assert builder.hook.build_backend == DEFAULT_BACKEND
```

Unless it is unintentional that `builder.hook` is part of the public API, or that it is guaranteed to be a `Pep517HookCaller` object, in which case probably mocking out the default backend, calling `.build` and then using `Mock.assert_called_once` or `Mock.assert_called_once_with` would be more approriate.

tests/test_projectbuilder.py
```diff
@@ -148,26 +146,53 @@ def test_check_dependencies(mocker):
...
+
+
+def test_missing_backend(mocker, pyproject_no_backend_mock):
+    mocker.patch('importlib.import_module')
+    mocker.patch('{}.open'.format(build_open_owner), pyproject_no_backend_mock)
+    mocker.patch('pep517.wrappers.Pep517HookCaller')
```
pganssle, line 185: Why is this mocked out? Because the constructor will try and import the build backend? If so, can we try a more targeted mock that would simply *provide* the build backend?

pganssle:
> Looking through the code I'm not confident that we have a good enough test suite to validate the approach here. So I was advocating for more tests; ideally on top of moving our current integration tests from CI to pytest. @FFY00 wants to move ahead as is now. The current state at a glance looks ok, so I'll not block it, though the tests we have does not give me enough confidence to approve it either.

I tend to agree that the testing I see here is not really adequate given that this is the _core functionality of this module_. I am very dubious about the heavy use of mocks.

If it were me, I'd take the time to do the testing beforehand, because I know that I very rarely go back to add more tests, and if I don't add the tests stuff will be fundamentally broken, but that's my workflow.

I don't think we should do a production release of this software without adding the tests, but if there's some compelling reason to merge first and refactor the tests later (like avoiding annoying merge conflicts if/when we move to the `src/` layout), I'm fine with merging and then opening a release-blocker bug for "add significantly more thorough tests" here.

tests/test_projectbuilder.py
```diff
@@ -148,26 +146,53 @@ def test_check_dependencies(mocker):
...
         builder.build('sdist', '.')
 
     with pytest.raises(build.BuildBackendException):
         builder.build('wheel', '.')
+
+
+def test_default_backend(mocker, empty_file_mock):
+    mocker.patch('importlib.import_module')
+    mocker.patch('{}.open'.format(build_open_owner), empty_file_mock)
+    mocker.patch('pep517.wrappers.Pep517HookCaller')
```
pganssle, line 175: I'm not super familiar with this `mocker` fixture. Is there any way to use something equivalent to [`autospec`](https://docs.python.org/3/library/unittest.mock.html#auto-speccing) with it? If we really need to mock this stuff out, we should be as surgical with the mocking as possible.

tests/test_projectbuilder.py
FFY00: Yes, this can be removed when we refactor the test suite to use proper packages.

tests/test_projectbuilder.py
FFY00: > This plugin provides a mocker fixture which is a thin-wrapper around the patching API provided by the mock package

https://github.com/pytest-dev/pytest-mock/

Everything supported by `unittest.mock.path` is supported by `mocker.patch`.

FFY00:
> I don't think we should do a production release of this software without adding the tests, but if there's some compelling reason to merge first and refactor the tests later (like avoiding annoying merge conflicts if/when we move to the `src/` layout), I'm fine with merging and then opening a release-blocker bug for "add significantly more thorough tests" here.

Yes, I want to avoid having to keep this in sync with new changes, especially since it is such a big PR. The current test suite is not optimal, but it does its job. I want to refactor it and remove a lot of the mocks, but that will have a lot of conflicts with this.
As I said before, we should not release without reworking the test suite. But stalling this until the test suite gets refactor will only slow things down as it will force me to spend time maintaining this in parallel.

gaborbernat:
> But stalling this until the test suite gets refactor will only slow things down as it will force me to spend time maintaining this in parallel.

You would not need to if you switch to improving the test suite first and then coming back to this 🤔

FFY00:
I was gonna open a PR to this branch but would be a bit of a weird review process, I prefer it this way. We can always revert the commits if needed.

gaborbernat:
> We can always revert the commits if needed.

Not really once we do the refactor needed, e.g. src layout, the revert would cause merge conflicts.

FFY00:
Which can be solved. Only if you build on top of the current code it might become an issue. There are no plans for this right now, what we have in hand is nox > tox (in.c src layout change) and refactoring the tests.

gaborbernat:
You forgot about https://github.com/FFY00/python-build/issues/19 which is bound to make changes everywhere

FFY00:
I would stall it until all this is sorted out.

gaborbernat:
> I would stall it until all this is sorted out.

In the end no need for that. Reverting it has no value at the end of the day because it's not like we'd go ahead without a similar feature. If it's not working we can just fix-it in a follow-up PR. In the end, the decision is with you as primary maintainer, but it would likely prevent me from doing PRs. Changing code that's not auto-formatted raises too many potential style arguments I'm not willing to engage as part of my free time.

FFY00:
> In the end no need for that. Reverting it has no value at the end of the day because it's not like we'd go ahead without a similar feature. If it's not working we can just fix-it in a follow-up PR.

Hence the "if needed".

> but it would likely prevent me from doing PRs

You can make your PRs and after approving I can push a commit fixing the style and squash merge. You don't need to worry about this, feel free to send code PRs without worrying about the style. This would be temporary until we adopt black.

Anyway, this conversation is not going anywhere, we should focus on actually writing the code :stuck_out_tongue:.

gaborbernat:
No thanks I'll wait until black is in place 👍will take a look at tox/nox though.

PR #21, opened 2020-06-13
doc: generate program usage automatically

FFY00:


docs/source/index.rst
```diff
@@ -11,30 +11,15 @@ It is a simple build tool, it does no dependency management.
...
-
-   optional arguments:
-   -h, --help            show this help message and exit
-   --sdist, -s           build a source package
-   --wheel, -w           build a wheel
-   --outdir ./dist, -o ./dist
-                           output directory
-   --skip-dependencies, -x
-                           does not check for the dependencies
+.. program-output:: python -m build --help
```
gaborbernat, line 14: Can we instead use https://pypi.org/project/sphinxcontrib-autoprogram/#description that generates CLI flags that can be URL linked? (see https://tox.readthedocs.io/en/latest/config.html#cli)

FFY00: Done.

On top for #22 for the isolated `main_parser` function.

docs/source/conf.py
```diff
@@ -15,9 +15,9 @@
 
 from pallets_sphinx_themes import ProjectLink
 
-import build
+sys.path.insert(0, os.path.abspath(os.path.join(__file__, '..', '..', '..')))  # isort:skip
```
gaborbernat, line 18: Generally, you should not be injecting paths manually on sys.path. Consider instead making the docs environment install the package and just import. See https://github.com/pypa/virtualenv/blob/master/.readthedocs.yml#L8-L14 for the readthedocs incantation

FFY00:
@gaborbernat btw, at least for my PRs, please use rebase instead of squashing the commits.

8fdbd931aed8cfcec01226f59df5a465e09f2219 has a bunch of random changes that should be in their own commits.

Sorry for being picky! I really care about git history.

gaborbernat:
IMHO if you think they should have been separate commits they should have been separate pul requests.🤷‍♂️

FFY00:
It doesn't make sense to me to create 4 pull requests for commits that are part of a logic block. When I group commits into one pull request, it's my way of saying that they should be merged at the same time. The commits are logically linked and should be reviewed together, but this doesn't mean they should be the same commit. Does that make sense?

gaborbernat:
Makes sense, I just think in one pr one feature/bugfix. So at the end one pr is one commit. Makes revert also easy and straight forward to use.

FFY00:
I usually don't use this model, it might not be an issue when you are dealing with simple stuff but I have found that it does not scale well in terms of complexity. I've found myself needing to retrace the steps leading to certain changes in codebases, this is significantly easier if we have small commit logically separated. For me git history is also documentation, I sometimes add some context on the change, why it's needed and/or how it works. All that metadata is lost when you the commits are squashed.

I understand that not everybody does this, and that is fine. In the future just try to remember to change the merge button to rebase in my commits if you can :smiley:.

Also, reverting multiple commits is almost the same.
```
git revert HEAD~6..HEAD~8
```

gaborbernat:
I guess we use git differently. I personally view the in between commits as noise only useful for review time, but not at all once merged.

pganssle:
FWIW I _also_ try and keep my git history pristine. Each commit is a logically independent change, and frequently one that can at least mostly be rolled back _independent of the other changes_.

The biggest downsides to doing rebase-merge or merge commits vs. squash merges are:

1. A lot of tools can't be configured as "this must pass on every commit", so it's common to find random commits in a PR that won't pass the test suite or the linters or something (which can make automated `git bisect` runs a bit annoying).
2. It's a lot of overhead to ask your random contributors to do interactive rebases to clean up their history, and it makes it a lot harder to just click the "merge" button on a good PR.

One of the biggest downsides of "always squash merge" we've already run into together — when `dateutil`'s test suite stopped working because of a change in `tox`, we did a `git bisect` and found that it was some mega-PR with tons of semi-related changes in it. I had to throw it to you to uncover what was going on, where smaller, granular changes even within a single PR would have highlighted the exact issue from the start.

I think it's reasonable to default to using squash merges for contributors who are not _obviously_ curating their commit history, but to do rebase merges for people like @FFY00 and myself.

gaborbernat:
> I had to throw it to you to uncover what was going on, where smaller, granular changes even within a single PR would have highlighted the exact issue from the start.

I consider this a worthwhile trade-off for myself. In that case, I don't think the fix would have been any easier if it would have been separate PR. And I'm 99% sure I would have not made that one change separate PR. That changeset was a major rework of a part of the tool that touched multiple parts, so was made more with one big sweep, not many small changesets; that could have been kept separate.

FFY00:
> so it's common to find random commits in a PR that won't pass the test suite or the linters or something

Personally I always try to avoid this. I try to make sure everything works on every commit.

> It's a lot of overhead to ask your random contributors to do interactive rebases to clean up their history, and it makes it a lot harder to just click the "merge" button on a good PR.

This is true, and it is my main use of the squash and merge feature. Btw, Github enables the "allow maintainers to push to my branch" checkmark by default, so if you really want to, you can fix it for them :smile:. I don't generally do this, but it has happened in the past.

PR #26, opened 2020-06-16
pyproject: switch to setuptools legacy backend

FFY00:
I need stop being picky and just drop poerty as it doesn't make that much sense for this project. Maybe in the future...


pyproject.toml
```diff
@@ -1,50 +1,3 @@
...
-
-[tool.poetry.dev-dependencies]
-pytest = '^5.0'
-pytest-mock = '^3.0'
-pytest-cov = '^2.0'
-
-[tool.poetry.scripts]
-python-build = 'build.__main__:main'
+requires = ['setuptools >= 40.8.0', 'wheel']
+build-backend = 'setuptools.build_meta:__legacy__'
```
gaborbernat, line 3: Why the legacy backend?

pganssle: Yeah definitely don't use this. It's intended for backwards compatibility only.

FFY00: Ah, sorry. I just defaulted to `__legacy__` because it's our fallback on missing `build-backend`.

FFY00:
Nothing made it unusable. It was another place to maintain metadata.

gaborbernat:
I meant why not drop setuptools over poetry 😅

FFY00:
Ah, because I want it to be easily bootstrapable.

.github/workflows/build.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       run: pip install typing
 
     - name: Install build backend
-      run: pip install poetry
+      run: pip install setuptools wheel
```
gaborbernat, line 34: You should not need this, pip installs them on demand.

FFY00: Only after #18, right now we don't build in an isolated environment.

PR #27, opened 2020-06-19
build: allow passing config_settings to the backend

FFY00:

### DIFF
--- a/build/__init__.py
+++ b/build/__init__.py
@@ -16,7 +16,7 @@ import toml
 import toml.decoder
 
 
-if sys.version_info < (3,):  # pragma: no cover
+if sys.version_info < (3,):
     FileNotFoundError = IOError
     PermissionError = OSError
 
@@ -40,9 +40,9 @@ def check_version(requirement_string, extra=''):  # type: (str, str) -> bool
     '''
     import packaging.requirements
 
-    if sys.version_info >= (3, 8):  # pragma: no cover
+    if sys.version_info >= (3, 8):
         from importlib import metadata as importlib_metadata
-    else:  # pragma: no cover
+    else:
         import importlib_metadata
 
     req = packaging.requirements.Requirement(requirement_string)
--- a/build/__init__.py
+++ b/build/__init__.py
@@ -9,7 +9,7 @@ import importlib
 import os
 import sys
 
-from typing import Set
+from typing import Dict, List, Optional, Set, Union
 
 import pep517.wrappers
 import toml
@@ -19,6 +19,9 @@ import toml.decoder
 if sys.version_info < (3,):
     FileNotFoundError = IOError
     PermissionError = OSError
+
+
+ConfigSettings = Dict[str, Union[str, List[str]]]
 
 
 class BuildException(Exception):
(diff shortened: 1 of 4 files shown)


FFY00:
I should probably add some documentation on how the data is treated.

build/__init__.py
```diff
@@ -70,11 +70,12 @@ def check_version(requirement_string, extra=''):  # type: (str, str) -> bool
 
 
 class ProjectBuilder(object):
-    def __init__(self, srcdir='.'):  # type: (str) -> None
+    def __init__(self, srcdir='.', config_settings={}):  # type: (str, Dict[str, Union[str, List[str]]]) -> None
         '''
         :param srcdir: Source directory
         '''
         self.srcdir = srcdir
+        self.config_settings = config_settings
```
merwok, line 78: Is it ok to share the dict here?  (compared to setting a new empty dict if received param is None)

gaborbernat: Yeah this definitely, you should never use mutable objects as default.

build/__main__.py
```diff
@@ -27,17 +27,19 @@ def _error(msg, code=1):  # type: (str, int) -> None  # pragma: no cover
     exit(code)
 
 
-def build(srcdir, outdir, distributions, skip_dependencies=False):  # type: (str, str, List[str], bool) -> None
+def build(srcdir, outdir, distributions, config_settings={}, skip_dependencies=False):
```
gaborbernat, line 30: no mutable object defaults

build/__main__.py
```diff
@@ -96,6 +101,28 @@ def main(cli_args, prog=None):  # type: (List[str], Optional[str]) -> None
     args = parser.parse_args(cli_args)
 
     distributions = []
+    config_settings = {}
+
+    if args.config_setting:
+        for arg in args.config_setting:
+            if sys.version_info >= (3,):
+                data = arg.split('=', maxsplit=1)
+            else:  # pragma: no cover
```
gaborbernat, line 110: eh, why do not cover this?

FFY00: Because then we have to merge the reports, and currently Python 3 and Python 2 jobs run in different jobs. I might fix it later, by creating a new job that gathers the report from the tests and pushes them to codecov.

gaborbernat: I'd prefer to leave it uncovered to indicate job to be done rather than override it like this 👍

FFY00: Hum, looks like code codecov should be able to merge multiple reports, we just might have to pass the `PYTHON` env var. I will open a PR for this.

FFY00:
Rebased on top of #30.

FFY00:
Should I expand the other tests?

```diff
diff --git a/tests/test_main.py b/tests/test_main.py
index 4e411a5..e0e8548 100644
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -23,12 +23,30 @@ out = os.path.join(cwd, 'dist')
 @pytest.mark.parametrize(
     ('cli_args', 'build_args'),
     [
-        ([], [cwd, out, ['sdist', 'wheel'], {}, False]),
-        (['-s'], [cwd, out, ['sdist'], {}, False]),
-        (['-w'], [cwd, out, ['wheel'], {}, False]),
-        (['source'], ['source', out, ['sdist', 'wheel'], {}, False]),
-        (['-o', 'out'], [cwd, 'out', ['sdist', 'wheel'], {}, False]),
-        (['-x'], [cwd, out, ['sdist', 'wheel'], {}, True]),
+        (
+            [],
+            [cwd, out, ['sdist', 'wheel'], {}, False]
+        ),
+        (
+            ['-s'],
+            [cwd, out, ['sdist'], {}, False]
+        ),
+        (
+            ['-w'],
+            [cwd, out, ['wheel'], {}, False]
+        ),
+        (
+            ['source'],
+            ['source', out, ['sdist', 'wheel'], {}, False]
+        ),
+        (
+            ['-o', 'out'],
+            [cwd, 'out', ['sdist', 'wheel'], {}, False]
+        ),
+        (
+            ['-x'],
+            [cwd, out, ['sdist', 'wheel'], {}, True]
+        ),
         (
             ['-C--flag1', '-C--flag2'],
             [cwd, out, ['sdist', 'wheel'], {
```

PR #28, opened 2020-06-20
Use setup.cfg over setup.py

gaborbernat:
Resolves https://github.com/FFY00/python-build/issues/40


gaborbernat:
> This breaks `distutils` support which might be wanted in some scenarios when bootstrapping, so I am -1.

Can you detail such use cases? The way I know distutils is not planned to be supported going ahead (@pganssle can confirm this). It certainly did not cause issues for virtualenv. Even in cases where you don't have access to setuptools, you can just use the wheel we'll publish.

FFY00:
I agree `distutils` should not be used by anyone, but there are some scenarios where we might not have `setuptools`. Distributions like arch prefer to build from source than using distributed wheels. Using distributed wheels is not even an option when the package is not pure.

I would be slightly inclined the other way if this provided any kind of improvement other than just storing the metadata in a nicer way.

gaborbernat:
Let's agree to disagree on this too, I'll not push on it if you feel so strongly about it.

FFY00:
Bootstrapping python envrionments from scratch is already a pain, I don't want to make it even harder :smile:.

gaborbernat:
> Bootstrapping python envrionments from scratch is already a pain, I don't want to make it even harder 😄.

If you're worried about this IMHO we should be removing setuptools dependency entirely, and use distutils only. Otherwise, we're telling a blatant lie to ourself: that we fallback to distutils and it works. In the current CI we never test this, so we have exactly 0 confidence in this. We'll never know if we use a flag that doesn't work with distutils, until we release and someone gets back to us complaining. 

> Bootstrapping python envrionments from scratch is already a pain, I don't want to make it even harder 😄.

Didn't we introduce the no isolation build for this? I'd expect one would resolve to bootstrap this project the same way. Install our build dependencies from pyproject.toml and then start building this package.

pganssle:
@gaborbernat I also was thinking to use `setup.cfg` instead of `setup.py`, but I assumed that the context here is that @FFY00 is an Arch Linux maintainer involved in packaging Python, and all these Linux distros have weird rules about what tools need to be built from source.

They'll have to solve the problem of "Python doesn't have distutils in it anymore" at some point after [this PR is merged](https://github.com/pypa/setuptools/pull/2143) and we start the process of removing `distutils` from Python in favor of `setuptools`, but I think the point here is that, at least at the moment, it would presumably be convenient if at least some version of `build` could be built _without_ any dependencies (including `setuptools`).

I think this is a pretty simple package, though. Depending on the needs of the Linux distro packaging community, possibly the best thing to do in the long term would be to keep a hard `setuptools` dependency for building the `sdist`, but customize include additional information in the `sdist` that would allow for a zero-dependency wheel-building or installation step.

That said, I don't know how important it is to be able to do a zero-dependency _build_ of this package. Actually _using_ the package requires the `pep517` library, which is built with `flit`, and `flit` has run-time dependencies on a few packages that have hard build-time dependencies on `setuptools`, so I don't think that "bootstrap build so that you can build `setuptools`" will be a feasible strategy right now anyway.

Presumably @FFY00 has more information here. If a hard requirement on `setuptools` is actually going to be a problem for distro packagers, then I think it's reasonable to keep this compatible with `distutils` while distro packagers figure out what their setup will look like in a post-`distutils` world. If there's no immediate _practical_ value to this, I'm inclined to ditch `setup.py` entirely now as we explore options that will ease the burden on distro packagers.

gaborbernat:
> then I think it's reasonable to keep this compatible with `distutils`

If we do this we might as well use for now distutils as the primary... and add setuptools as fallback once https://github.com/pypa/setuptools/pull/2143 lands. I'm ok with not using setuptools for this if bootstrapping is an issue, but then we should drop setuptools entirely, and test only with distutils. In the current form, the fallback to distutils is a wish at best, and not guaranteed to work.

pganssle:
> If we do this we might as well use for now distutils as the primary... and add setuptools as fallback once [pypa/setuptools#2143](https://github.com/pypa/setuptools/pull/2143) lands. I'm ok with not using setuptools for this if bootstrapping is an issue, but then we should drop setuptools entirely, and test only with distutils. In the current form, the fallback to distutils is a wish at best, and not guaranteed to work.

I disagree strongly with this. If we keep `distutils` support at all, we should just test that whatever feature of `build` it's important to use it for works if you build it with `distutils`, but `distutils` doesn't support stuff like `python_requires`, and generally speaking has a ton of bugs in it that wouldn't be acceptable either for people installing from `sdist` or for _us_ when building the wheel. The kind of people who care about building this with `distutils` will probably _not_ care about at least some of those problems (like `python_requires`), and will have specific workarounds for others. The vast majority of people should be either building with `setuptools` or (even more preferable) consuming a wheel built with `setuptools`.

gaborbernat:
Fair enough, but long story short if we claim distutils support we should validate that subset, or drop supporting it.

gaborbernat:
I personally the way things stand to vote for removing support for it unless @FFY00 can describe an actual use case where without is not possible.

FFY00:
> and all these Linux distros have weird rules about what tools need to be built from source

For arch is simple, *everything* should build from sources. There are some exceptions for cases where this is not possible, like proprietary tools.

I am not going to dive into how to bootstrap the dependency tree for Linux distributions since there are several different workflows that can be chosen. It is possible to build this project with setuptools instead of distutils when bootstrapping, but it might impact the amount of effort needed, depending on the particular workflow.

I am not opposed to drop support for distutils, I just don't think having a slightly nicer way to declare metadata is compelling enough.

@pganssle btw, `pep517` needs `flit_core`, not `flit`. `flit_core` has no runtime dependencies and can build itself.

Anyway, I am -1 on this. I don't really think we should remove it, but I have no issue if it is removed. In arch, we can bootstrap using setuptools.

gaborbernat:
I think we should remove it until someone has an actual use case where using setuptools is not possible, or alternatively keep it but then add tests to validate it actually works.

FFY00:
I re-triggered the tests, codecov is now fixed.

pganssle:
I really don't have enough insight into how much more difficult this would make the lives of distro maintainers, so it's hard for me to have a strong opinion on it.

One way to "split the baby" is to use `setuptools` unconditionally, but also not migrate everything to `setup.cfg`, so that if distro maintainers want to patch the `sdist` to make it easier to bootstrap with `distutils`, it's a simpler change for them. Of course, that would be "best effort". That might be worse than doing nothing, though.

I also think that if we *do* want to switch to declarative metadata, we'll want to switch to [PEP 621](https://www.python.org/dev/peps/pep-0621/) in ~a few months (or whenever support for PEP 621 lands in `setuptools`), so I'm not as gung-ho about `setup.cfg` as I would have been a few months ago.

Sorry I can't be more definitive, I just don't feel like I have enough information.

FFY00:
If @gaborbernat is super keen on making this change, I am willing to accept it.

gaborbernat:
> If @gaborbernat is super keen on making this change, I am willing to accept it.

The alternative of testing distutils is painful and troublesome enough that I would not worry about supporting distutils until someone comes and presented a case where it's impossible to provision setuptools instead. I'd expect all distributions by now figured out how to inject setuptools given it's popularity, so I'd not expect this to be a problem. And @FFY00 just confirmed this for Arch. So I yes let's merge this if someone is willing to approve it.

setup.cfg
```diff
@@ -1,3 +1,48 @@
...
+license_file = LICENSE
+classifiers =
+    License :: OSI Approved :: MIT License
+    Programming Language :: Python :: 2
+    Programming Language :: Python :: 2.7
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.5
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
```
pganssle, line 18: 3.9?

setup.cfg
```diff
@@ -1,3 +1,48 @@
...
+project_urls =
+    homepage=https://github.com/FFY00/python-build
```
pganssle, line 20:
```suggestion
    homepage = https://github.com/FFY00/python-build
```

setup.cfg
```diff
@@ -1,3 +1,48 @@
...
+    packaging
+    pep517
+    toml
+    importlib-metadata;python_version < "3.8"
+    typing;python_version<"3"
+python_requires = >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*
+
+[options.entry_points]
+console_scripts =
+    python-build=build.__main__:main
```
pganssle, line 34:
```suggestion
    python-build = build.__main__:main
```

setup.cfg
```diff
@@ -1,3 +1,48 @@
...
+
+[options.extras_require]
+doc =
+    sphinx
+    sphinx-autodoc-typehintssphinxcontrib-autoprogram
```
FFY00, line 39: Missing a new line here.

setup.cfg
gaborbernat: Separate PR where we add 3.9 to the CI too.

PR #29, opened 2020-06-20
Add .gitignore to ignore noise

gaborbernat:


gaborbernat:
Let's agree to disagree. Requesting people to maintain a global gitignore is not contribution friendly.

FFY00:
I don't think we should be maintaining a gitignore for all editors or other environment-specific stuff. We add idea and vscode today, tomorrow we'll have someone asking for vim swap files, eclipse, etc. Proving a `.gitignore` is still a great idea!

I went through the list again, where does `*$py.class` come from?

I have a more extensive list:
```
.nox/
.mypy_cache/
.coverage/
__pycache__/
htmlcov/
dist/
docs/srcbuild/

coverage.xml
*.py[cod]
```

What do you think?

gaborbernat:
I'll defer the question. I don't have the time to fight this battle for every repository, so I'll chicken out of this.

PR #30, opened 2020-06-20
ci: upload coverage reports for all test environments

FFY00:


FFY00:
Yup, it works. Last time I tried something like this I had issues.

PR #32, opened 2020-06-21
Add .gitignore

pganssle:
Picking up from #29, I think we definitely need a `.gitignore`. I agree with @FFY00 about IDE-related stuff going in global `.gitignore`, though I think my list of stuff we should ignore is longer than his.

I've started with something based on what gets generated when I run `nox` in a clean build.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,21 @@
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+
+# Distribution / packaging
+build/lib/
+dist/
+*.egg-info/
+.eggs
+
+# Documentation
+docs/srcbuild/
+
+# Testing and coverage
+.cache
+.nox/
+.mypy_cache/
+.pytest_cache
+htmlcov/
+.coverage
+coverage.xml


pganssle:
For some reason `pytest` doesn't generate any cache files for me when I run it, but I do have other projects that ignore both `.cache` and `.pytest_cache`, so that seems fine as well.

I'm also inclined to settle on one or more names for virtual environments, like `venv` and `.venv`. This is especially useful if we adopt `black`, because `black` won't try to reformat things in your `.gitignore` file, and if they aren't in there, it will try to recursively reformat all the libraries you've installed. A global ban on `venv/` wouldn't work very well if you ever try to work on the `venv` module of CPython, which at least Bernat and I are both likely to do :stuck_out_tongue:

FFY00:
I think we should run `black`, or any code formatter we might choose, on the folder in question, just like we do for `flake8`, `isort`, etc.

But feel free to add other common venv paths to `.gitignore`. I just want to put on the record I think this _might_ be a slippery slope. I am not going to block it, however.

pganssle:
Let's start with the subset of things we can obviously agree on. We can discuss the question of virtualenv names later.

PR #33, opened 2020-06-21
nox: add pytest-mock dependency

pganssle:
Without this, `nox` fails.

If we're going to have a `noxfile.py`, I think it should be used in the CI as well, so that we can expect contributors to be able to just run `nox` or `nox -s test-3.8` or whatever and it should always work.

Here's how I typically do it when I use `tox` in my projects: https://github.com/pganssle/zoneinfo/blob/4299bfd7623934acc6f27b32744b02a147912120/.github/workflows/python-tests.yml#L5-L32

I think that you'd do something similar by parameterizing over `NOXSESSION`.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -34,6 +34,6 @@ def test(session):
     htmlcov_output = os.path.join(session.virtualenv.location, 'htmlcov')
 
     session.install('.', 'importlib_metadata')
-    session.install('pytest', 'pytest-cov')
+    session.install('pytest', 'pytest-cov', 'pytest-mock')
 
     session.run('pytest', '--cov', f'--cov-report=html:{htmlcov_output}', 'tests/', *session.posargs)


FFY00:
I think I would prefeer to still have parallel jobs, which makes the CI much faster. We can add a matrix on the nox jobs.

pganssle:
The CI configuration I linked does run the tox jobs in parallel. The CI creates a separate job for each interpreter and runs the equivalent job in tox. It's easier to do with tox AFAICT, but I think I can make it work with nox easily enough.

gaborbernat:
Maybe we just use tox 😅 instead

FFY00:
With nox should be pretty easy too, I will open a PR.

Also note https://github.com/theacodes/nox/issues/334, if it gets resolved it should allow us to cache nox and make everything much faster.

pganssle:
Hm.. This might be more complicated than I would have hoped if we're going to support earlier versions of Python, since `noxfile.py` has an f-string in it.

We can either drop the f-string so that `nox` can be run from earlier interpreters or install the target Python and Python 3.8, then always invoke `nox` from Python 3.8. I feel like the second one will be harder to accomplish.

FFY00:
Just setup the standard python installation as the latest python and then set up an extra one with the target version.

pganssle:
Ugh, looks like the last version of `nox` that supported Python 2.7 did not have `python_requires`.

@FFY00 I don't think I have time to play with the CI stuff for much longer. Feel free to push stuff directly to this branch or make your own fork if you want to take over.

FFY00:
Dropped the last 2 commits in favor of #35.

gaborbernat:
IMHO these dependencies should be lower bound as it's not true will work with any version, what you're implying in the current format.

gaborbernat:
> Ugh, looks like the last version of `nox` that supported Python 2.7 did not have `python_requires`.
> 
> @FFY00 I don't think I have time to play with the CI stuff for much longer. Feel free to push stuff directly to this branch or make your own fork if you want to take over.

The more I use nox the less I like it 😅 I know I'm not impartial here being the tox maintainer but still 😟

gaborbernat:
> I think I would prefer to still have parallel jobs, which makes the CI much faster. We can add a matrix on the nox jobs.

This project CI/test suite is fast enough that IMHO at this point CI time optimization feels premature optimization, and IMHO worth optimizing for convenience and ease of use instead.

FFY00:
> IMHO these dependencies should be lower bound as it's not true will work with any version, what you're implying in the current format.

That is true, I was just have been lazy to figure out the minimum compatible version. Since pip's default behavior is to install the latest one, it hasn't been a problem.

> The more I use nox the less I like it sweat_smile I know I'm not impartial here being the tox maintainer but still worried

I like the configuration approach but I have found some sharp edges. Would https://github.com/theacodes/nox/issues/334 still happen in tox?

> This project CI/test suite is fast enough that IMHO at this point CI time optimization feels premature optimization, and IMHO worth optimizing for convenience and ease of use instead.

Well, it's just CI configuration, and I am more than happy to do it so it's not convenient. I don't think the configuration is more complex because of the parallel jobs, it's just different. The problem Paul was hitting would have happened in both scenarios, the trick here is that the lastest python installation will overwrite `/usr/bin/python`, so the python version with which we want to run nox needs to be installed last.

gaborbernat:
> I like the configuration approach but I have found some sharp edges. Would [theacodes/nox#334](https://github.com/theacodes/nox/issues/334) still happen in tox?

No idea why nox takes so slow so would be hard to tell (maybe hits pip build slow?). In general, my tox projects are fast-ish (<3s) to install build dependencies. But given how pyproject.toml works depend on your network speed. Furthermore, a better fix is to install wheels, to further speed things up. tox uses a cached environment to build the sdist, so should be faster that part than nox.

PR #36, opened 2020-06-21
codecov: disable the bot comment

### LINKED ISSUE #34: Turn off codecov comment?
Can we turn off the codecov comment bot? I find it spammy and annoying, and I can't figure out a way to prevent it from sending me e-mails every time someone pushes a commit. The status hook should be enough.

FFY00:
Closes #34

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,2 +1,1 @@
-comment:
-  layout: 'diff'
+comment: false


gaborbernat:
I'm personally happy to keep it on, but no strong feelings.

pganssle:
If y'all really like it feel free to keep it. I cannot find a way to block it from sending e-mails, so I find it very irritating; considering we have 100% coverage on this project, if we make 100% coverage required for all PRs, the coverage comment will be kinda useless anyway — you'll either have 100% coverage and the status hook will pass, or you won't have 100% coverage and you need to look at the full coverage details _anyway_ to see exactly what lines need coverage.

But then again, I never use this thing, so I don't know how people use it.

FFY00:
The comment is just for information, it does not impact the review in any way. It is a bit annoying receiving emails every PR, so I think it's probably best to disable it to avoid annoying future contributors, right?

PR #47, opened 2020-06-24
pre-commit: use remotes and add extra checks

gaborbernat:


.github/workflows/checks.yml
```diff
@@ -3,10 +3,10 @@ name: checks
 on:
   push:
     branches:
-      - master
+    - master
   pull_request:
     branches:
-      - master
+    - master
```
FFY00, line 9: Can we keep the style? If we want to change it, it can be done in a separate PR.

.github/workflows/checks.yml
```diff
@@ -35,38 +35,9 @@ jobs:
...
-      run: |
-        pip install --upgrade pip
-        pip install isort
-        isort --version
-
-    - name: Run isort
-      run: isort --check-only --diff --recursive build
+    - uses: actions/checkout@v1
+    - uses: actions/setup-python@v1
+    - uses: pre-commit/action@v2.0.0
```
FFY00, line 43: The current style declares human-readable names for each step, and has a new line between steps.

.github/workflows/checks.yml
```diff
@@ -35,38 +35,9 @@ jobs:
       run: mypy --python-version ${{ matrix.python }} -p build
 
 
-  flake8:
+  pre-commit: # per https://github.com/pre-commit/action
```
FFY00, line 38: I do not think this comment is needed.

setup.cfg
```diff
@@ -56,3 +56,5 @@ strict = True
 line_length = 127
 lines_between_types = 1
 lines_after_imports = 2
+known_first_part = build
+known_third_party = nox,pallets_sphinx_themes,pep517,pytest,setuptools,toml
```
FFY00, line 60: Do we need to explicitly declare this? isort hasn't been giving any issues.

.github/workflows/checks.yml
gaborbernat: Just called the YAML formatter on this file and this was collateral.

.github/workflows/checks.yml
gaborbernat: I just used the suggestion of pre-commit, not sure how much value the human-readable names add, but if you really want can add them 👍

setup.cfg
gaborbernat: This makes running isort much faster because avoids the need to install these packages to discover, and makes explicit a list of used packages, so I'm personally strongly in favor it.

.github/workflows/checks.yml
gaborbernat: I think it helps with where this setup recommendation comes from 🤔

.github/workflows/checks.yml
FFY00: I never `git add file`, I always use `git add -p`.

Given this is from the yaml formatter introduced here, it is okay to have it, but please do it for the whole codebase. And preferably, in a different commit.

.github/workflows/checks.yml
FFY00: The recommendation is a simple example of how to use a github action, I don't think it adds any extra value.

.github/workflows/checks.yml
gaborbernat: > I never `git add file`, I always use `git add -p`.

Not sure how this is relevant here, but 👍 for you. I'll remove this change, no worries. I'm not of fan of different commits, so my changes will always be one-commit 🤷 per feature/bugfix.

setup.cfg
FFY00: Wouldn't `known_first_part` be enough then? `known_third_party` is just another extra place to maintain dependencies.

gaborbernat: Not it's not enough, see https://github.com/asottile/seed-isort-config#seed-isort-config for more details, @asottile can shed more light too

FFY00: Why doesn't isort have a config (possibly `first_part`) where I can say declare our packages? It feels very silly to me to add all modules we use in the config file when I can just say "threat anything other than `build` as an external module".

gaborbernat: You're free to raise the issue with isort, but that's not a thing today.

gaborbernat:
@FFY00 addressed your current comments, want to review again? Note some required checks need to be altered to merge (I don't have rights to do that).

.github/workflows/checks.yml
```diff
@@ -35,9 +35,12 @@ jobs:
...
 
-  pre-commit: # per https://github.com/pre-commit/action
+  pre-commit:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v1
-    - uses: actions/setup-python@v1
-    - uses: pre-commit/action@v2.0.0
+    - name: Checkout
+      uses: actions/checkout@v2
```
FFY00, line 42: Missing the new lines between steps.

gaborbernat: https://github.com/FFY00/python-build/issues/51 to address this

PR #53, opened 2020-06-27
Use src layout for package

gaborbernat:
Without this the build name clashes with setuptools build folder, rendering source distributions broken.


gaborbernat:
> Also, maybe I should be the one doing the renaming part of the commit, to keep the git blame. It is basically me everywhere.

I offer almost no value to git blame within OSS. The fact I touched some code two years ago does not mean I'm still willing to invest time in it today. If anything needs to be blamed with OSS your best guess is the current maintainer.

noxfile.py
```diff
@@ -23,7 +23,7 @@ def test(session):
     session.install('.', 'importlib_metadata')
     session.install('pytest', 'pytest-cov', 'pytest-mock')
 
-    session.run('pytest', '--cov',
+    session.run('pytest', '--cov=src',
```
FFY00, line 26: Can we do this on a config instead?

gaborbernat: Considering we'll replace this with tox I'd rather do it there.

FFY00: Why? As I explained, tox is not desirable for distro packagers at the moment. This configuration is something we want to always have, and as such makes more sense to be in a config file rather than passed as an argument.

gaborbernat: I don't think that adds any value considering the coverage will migrate to be driven by tox, so no, not for now.

gaborbernat: Because wasn't before this PR so out of scope.

FFY00: But tox can make use of coverage configuration file, no? If not, why?

FFY00: > Because wasn't before this PR so out of scope.

It is not out of scope, I am asking you to add a configuration introduced by this PR in a config file instead of hardcoding it in nox/tox.

Same thing we do for flake8, mypy and isort.

PR #54, opened 2020-06-27
build: move module source to src directory

FFY00:


.github/workflows/build.yml
```diff
@@ -35,6 +35,8 @@ jobs:
 
     - name: Build python-build (without isolation)
       run: python -m build -x -n
+      env:
+        PYTHONPATH: src
```
pganssle, line 39: This defeats half the purpose of using a `src/` environment. We should only be testing _installed_ versions of the library, since that's what end users will be using.

If we want to test the ability to bootstrap `build` from the local directory, we should do that separately.

docs/Makefile
```diff
@@ -35,4 +36,4 @@ coverage: Makefile apidoc
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
 %: Makefile apidoc
 	cp -r "$(SOURCEDIR)"/* "$(SRCBUILDDIR)"
-	@PYTHONPATH=.. $(SPHINXBUILD) -M $@ "$(SRCBUILDDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+	@PYTHONPATH=$(SRCDIR) $(SPHINXBUILD) -M $@ "$(SRCBUILDDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
```
pganssle, line 39: I guess this is already how things work, but I don't think we should modify `PYTHONPATH` in this make file.

I usually never manipulate `PYTHONPATH` or `sys.path`, but for people who do, the more common idiom is to add this to `sys.path` in your `conf.py`.

When I have a project where Sphinx requires the project to be on the `PATH`, I usually just make "install the project into the relevant virtualenv" a prerequisite for building the docs.

tests/__init__.py
```diff
@@ -5,7 +5,7 @@
 
 
 f = os.readlink(__file__) if os.path.islink(__file__) else __file__
-path = os.path.realpath(os.path.join(f, '..', '..'))
+path = os.path.realpath(os.path.join(f, '..', '..', 'src'))
```
pganssle, line 8: This entire file should be empty. It is inappropriate to modify `sys.path` as part of the tests.

As with the integration tests, we should be testing *as installed*.

.github/workflows/build.yml
FFY00: I definitely want to test running from a local directory, it will be required for bootstrapping after `setup.py install` gets removed. The question is which behavior we should keep in this PR. Both behaviors will be tested in the integration tests, so I don't have a strong opinion here, though I am inclined to just set `PYTHONPATH`.

docs/Makefile
FFY00: I was modifying `sys.path` in `conf.py` previously, but was asked to stop. I can reintroduce this.

pganssle: Giving it more thought, I suppose this _is_ better than the `sys.path` modification in `conf.py`, because this only affects people building using the `Makefile` and allows us to use more strict build criteria for things like a `tox` environment or in the ReadTheDocs configuration, so I withdraw the objection.

.github/workflows/build.yml
pganssle: I think the situations in which we support running from a local directory should be *extremely* limited, and we should have custom, explicit tests for them if we test them at all.

You should never be installing `Solaar` or `python-dateutil` using `build` installed in a local directory, so we definitely shouldn't be testing it.

We should drop *all* path manipulation in all of our tests right now, and if we want to introduce a separate "can you bootstrap `build`" (or some other legitimate use for running it from a local directory), we should do that in a different PR.

FFY00:
@pganssle can you review the changes here? So that we can move forward with #56.

setup.cfg
```diff
@@ -28,6 +28,8 @@ install_requires =
     importlib-metadata;python_version < "3.8"
     typing;python_version<"3"
 python_requires = >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*
+package_dir =
```
pganssle, line 31: I usually use `packages = find:` and then add another table:

```ini
[options.packages.find]
where = src
```

In this case it doesn't matter, but *not* doing that is one of the leading causes of the problems that get caught when you test in tree rather than testing the installed package. :wink:

We can leave that for later, though.

FFY00: Sure, I can add that. It would need to be tested by installing the package either way, `find` has some non-obvious behaviors that we need to keep in mind, for eg. I am pretty sure last time I tried it didn't install submodules if the "super"module did not have an `__init__.py`.

PR #56, opened 2020-06-30
Add integration tests and refactor tests a bit

FFY00:
On top of #54.


tests/test_integration.py
```diff
@@ -0,0 +1,52 @@
...
+    assert filter(_WHEEL.match, os.listdir(tmpdir))
+
+
+def test_isolation(tmpdir, test_flit_path, mocker):
+    try:
+        import flit_core  # noqa: F401
+    except:  # noqa: E722
+        pass
+    else:
+        pytest.xfail('flit_core is available')
```
merwok, line 47: simpler:

```python
with pytest.raises(ImportError):
    import flit_core
```

FFY00: I want to skip the test if `flit_core` is available, not fail.

FFY00:
I have no clue why coverage is not filtering the module on windows, I checked, the path is correct.

@pganssle do we have to run the tests as installed? Can't we just run from source in nox?

pganssle:
I just moved yesterday so I have extremely limited time / access to my computer, but I would say yes, 100% we must test as installed. We already have seen problems that we failed to catch because we weren't testing as installed.

FFY00:
>  just moved yesterday so I have extremely limited time / access to my computer

No worries.

> We already have seen problems that we failed to catch because we weren't testing as installed.

Well, sort of, but not really. The only issue we had is that the `build` module was not being included in the sdist, that wouldn't have been caught by installing the module. pip would build the with PEP517 and install the wheel, which is not broken, so everything would work as expected. And this isn't even really our issue, it's the backend's fault.

Anyway, I have spent a lot of time trying to get the coverage to work properly and I can't. It is not design with this use-case in mind. It is designed to capture coverage from modules run in-tree. The filtering functionality does not permit me to isolate the `build` module from the rest. 079264eb604315dbc4755a095cd7fa13337ffb2b almost did it, but it inexplicably fails on Windows. While the almost works, it is not a great solution.

What kind of issues are you expecting to find by running `build` from the site-packages directory instead of in-tree?

I had a quick look at other projects to see how they are doing it and it seems https://github.com/pypa/pep517, https://github.com/dateutil/dateutil and https://github.com/pypa/packaging are not installing the module for the tests.

While I agree it would be great to tests with the module installed, it just does not seem feasible in this case. I propose changing the tests to run against the module in-tree but keeping the integrations tests running with it installed. That way we solve the coverage issue but still make sure the project works when installed.

If you have a solution for the coverage issue, please let me know. I have already put in a very significant amount of time trying to make it to work, I can't keep spending time on this.

pganssle:
> Well, sort of, but not really. The only issue we had is that the `build` module was not being included in the sdist, that wouldn't have been caught by installing the module. pip would build the with PEP517 and install the wheel, which is not broken, so everything would work as expected. And this isn't even really our issue, it's the backend's fault.

I assumed that `nox` works the same way as `tox` in this regard, but that might be wrong. The way `tox` works, is that it always starts builds from an sdist, then installs from the sdist (not the package root) into each environment. If `nox` doesn't work that way, it's another reason to switch to `tox`, because `tox` always exercises the full package root → sdist → wheel → installation pathway. However, without testing against the installed module, you can have serious problems with the sdist and not notice it, since anything that doesn't actively raise an error will just be ignored.

> What kind of issues are you expecting to find by running `build` from the site-packages directory instead of in-tree?

I go into more details on this in [this blog post](https://blog.ganssle.io/articles/2019/08/test-as-installed.html), but essentially the issue is that there are significant differences from the two modes of use and the installed module is the one that we have guarantees about. The most common problem this tests against is modules failing to be installed, but there are other potential failure modes.

> I had a quick look at other projects to see how they are doing it and it seems https://github.com/pypa/pep517, https://github.com/dateutil/dateutil and https://github.com/pypa/packaging are not installing the module for the tests.

I'm kinda shocked that this is the case for `dateutil`, that's absolutely a mistake. I could have sworn that I learned the `changedir=` trick specifically to make it work properly with `dateutil`. `pep517` *is* testing as installed, because they invoke `pytest` rather than `python -m pytest` — when you invoke `pytest`, the current directory is not included in the test suite.

Here is the coverage configuration for [`backports.zoneinfo`](https://github.com/pganssle/zoneinfo/blob/4299bfd7623934acc6f27b32744b02a147912120/pyproject.toml#L8-L16), [`attrs`](https://github.com/python-attrs/attrs/blob/3e3d84216c5c7b6b451610ebbe8f5a7ed9c3209b/pyproject.toml#L6-L15) and [`pytz-deprecation-shim`](https://github.com/pganssle/pytz-deprecation-shim/blob/47bd4bdd9346cafa6c6d66817082ccce099890ad/pyproject.toml#L8-L13) and [`tox`](https://github.com/tox-dev/tox/blob/9e2cb4476c33ac93f09157d55a7ccc65e6497de4/tox.ini#L120-L141), all of which test as installed.

[This blog post by Mark R Williams](https://web.archive.org/web/20170615032800/https://enotuniq.org/) goes into some detail about how to configure coverage when running the tests from somewhere other than the directory root. I don't think it applies in this case, but it might be helpful.

If you're not willing to fix the coverage issue, I can take a look at it later, but this isn't a terribly complicated project, so I don't think there should be anything especially complicated about getting the coverage working compared to all the other projects using a `src/` layout.

.gitmodules
```diff
@@ -0,0 +1,9 @@
+[submodule "tests/integration/pip"]
```
pganssle, line 1: I highly recommend _against_ using submodules. They are a complete PITA for end users.

FFY00:
> `pep517` _is_ testing as installed, because they invoke `pytest` rather than `python -m pytest` — when you invoke `pytest`, the current directory is not included in the test suite.

I don't think that's true.

```
$ cat tests/test_origin.py
import os.path

import pep517


def test_origin():
    assert pep517.__file__ != os.path.abspath(os.path.join('pep517', '__init__.py'))
```

```
$ pytest tests/test_origin.py
============================= test session starts ==============================
platform linux -- Python 3.8.3, pytest-5.4.3, py-1.9.0, pluggy-0.13.1
rootdir: /home/anubis/git/pep517, inifile: pytest.ini
plugins: twisted-1.12, lazy-fixture-0.6.3, mock-2.0.0, cov-2.10.0, asyncio-0.14.0, hypothesis-5.19.0, flake8-1.0.6
collected 2 items

tests/test_origin.py FF                                                  [100%]

=================================== FAILURES ===================================
_________________________________ FLAKE8-check _________________________________
/home/anubis/git/pep517/tests/test_origin.py:7:80: E501 line too long (84 > 79 characters)

------------------------------ Captured log call -------------------------------
WARNING  flake8.options.manager:manager.py:207 option --max-complexity: please update from optparse string `type=` to argparse callable `type=` -- this will be an error in the future
WARNING  flake8.options.manager:manager.py:207 option --min-python-version: please update from optparse string `type=` to argparse callable `type=` -- this will be an error in the future
_________________________________ test_origin __________________________________

    def test_origin():
>       assert pep517.__file__ != os.path.abspath(os.path.join('pep517', '__init__.py'))
E       AssertionError: assert '/home/anubis/git/pep517/pep517/__init__.py' != '/home/anubis/git/pep517/pep517/__init__.py'
E        +  where '/home/anubis/git/pep517/pep517/__init__.py' = pep517.__file__
E        +  and   '/home/anubis/git/pep517/pep517/__init__.py' = <function abspath at 0x7f9cd43a4ca0>('pep517/__init__.py')
E        +    where <function abspath at 0x7f9cd43a4ca0> = <module 'posixpath' from '/usr/lib/python3.8/posixpath.py'>.abspath
E        +      where <module 'posixpath' from '/usr/lib/python3.8/posixpath.py'> = os.path
E        +    and   'pep517/__init__.py' = <function join at 0x7f9cd43a45e0>('pep517', '__init__.py')
E        +      where <function join at 0x7f9cd43a45e0> = <module 'posixpath' from '/usr/lib/python3.8/posixpath.py'>.join
E        +        where <module 'posixpath' from '/usr/lib/python3.8/posixpath.py'> = os.path

tests/test_origin.py:7: AssertionError
=========================== short test summary info ============================
FAILED tests/test_origin.py::FLAKE8
FAILED tests/test_origin.py::test_origin - AssertionError: assert '/home/anub...
============================== 2 failed in 0.22s ===============================
```

```
$ python -m pytest tests/test_origin.py
============================= test session starts ==============================
platform linux -- Python 3.8.3, pytest-5.4.3, py-1.9.0, pluggy-0.13.1
rootdir: /home/anubis/git/pep517, inifile: pytest.ini
plugins: twisted-1.12, lazy-fixture-0.6.3, mock-2.0.0, cov-2.10.0, asyncio-0.14.0, hypothesis-5.19.0, flake8-1.0.6
collected 2 items

tests/test_origin.py FF                                                  [100%]

=================================== FAILURES ===================================
_________________________________ FLAKE8-check _________________________________
/home/anubis/git/pep517/tests/test_origin.py:7:80: E501 line too long (84 > 79 characters)

------------------------------ Captured log call -------------------------------
WARNING  flake8.options.manager:manager.py:207 option --max-complexity: please update from optparse string `type=` to argparse callable `type=` -- this will be an error in the future
WARNING  flake8.options.manager:manager.py:207 option --min-python-version: please update from optparse string `type=` to argparse callable `type=` -- this will be an error in the future
_________________________________ test_origin __________________________________

    def test_origin():
>       assert pep517.__file__ != os.path.abspath(os.path.join('pep517', '__init__.py'))
E       AssertionError: assert '/home/anubis/git/pep517/pep517/__init__.py' != '/home/anubis/git/pep517/pep517/__init__.py'
E        +  where '/home/anubis/git/pep517/pep517/__init__.py' = pep517.__file__
E        +  and   '/home/anubis/git/pep517/pep517/__init__.py' = <function abspath at 0x7f16c9f4aca0>('pep517/__init__.py')
E        +    where <function abspath at 0x7f16c9f4aca0> = <module 'posixpath' from '/usr/lib/python3.8/posixpath.py'>.abspath
E        +      where <module 'posixpath' from '/usr/lib/python3.8/posixpath.py'> = os.path
E        +    and   'pep517/__init__.py' = <function join at 0x7f16c9f4a5e0>('pep517', '__init__.py')
E        +      where <function join at 0x7f16c9f4a5e0> = <module 'posixpath' from '/usr/lib/python3.8/posixpath.py'>.join
E        +        where <module 'posixpath' from '/usr/lib/python3.8/posixpath.py'> = os.path

tests/test_origin.py:7: AssertionError
=========================== short test summary info ============================
FAILED tests/test_origin.py::FLAKE8
FAILED tests/test_origin.py::test_origin - AssertionError: assert '/home/anub...
============================== 2 failed in 0.16s ===============================
```

And running tox confirms pep517 is indeed being tested from source:
```
$ tox -e py38
py38 installed: attrs==19.3.0,flake8==3.8.3,mccabe==0.6.1,mock==4.0.2,more-itertools==8.4.0,packaging==20.4,pluggy==0.13.1,py==1.9.0,pycodestyle==2.6.0,pyflakes==2.2.0,pyparsing==2.4.7,pytest==5.4.3,pytest-flake8==1.0.6,six==1.15.0,testpath==0.4.4,toml==0.10.1,wcwidth==0.2.5
py38 run-test-pre: PYTHONHASHSEED='3270814318'
py38 run-test: commands[0] | pytest
============================= test session starts ==============================
platform linux -- Python 3.8.3, pytest-5.4.3, py-1.9.0, pluggy-0.13.1
cachedir: .tox/py38/.pytest_cache
rootdir: /home/anubis/git/pep517, inifile: pytest.ini
plugins: flake8-1.0.6
collected 48 items

pep517/__init__.py s                                                     [  2%]
pep517/_in_process.py s                                                  [  4%]
pep517/build.py s                                                        [  6%]
pep517/check.py s                                                        [  8%]
pep517/colorlog.py s                                                     [ 10%]
pep517/compat.py s                                                       [ 12%]
pep517/dirtools.py s                                                     [ 14%]
pep517/envbuild.py s                                                     [ 16%]
pep517/meta.py s                                                         [ 18%]
pep517/wrappers.py s                                                     [ 20%]
tests/__init__.py s                                                      [ 22%]
tests/test_build.py s...                                                 [ 31%]
tests/test_call_hooks.py s.........                                      [ 52%]
tests/test_envbuild.py s..                                               [ 58%]
tests/test_hook_fallbacks.py s....                                       [ 68%]
tests/test_inplace_hooks.py s...                                         [ 77%]
tests/test_meta.py s...                                                  [ 85%]
tests/test_origin.py FF                                                  [ 89%]
tests/samples/buildsys_pkgs/buildsys.py s                                [ 91%]
tests/samples/buildsys_pkgs/buildsys_minimal.py s                        [ 93%]
tests/samples/pkg1/pkg1.py s                                             [ 95%]
tests/samples/pkg2/pkg2.py s                                             [ 97%]
tests/samples/pkg_intree/backend/intree_backend.py s                     [100%]

=================================== FAILURES ===================================
_________________________________ FLAKE8-check _________________________________
/home/anubis/git/pep517/tests/test_origin.py:7:80: E501 line too long (84 > 79 characters)

------------------------------ Captured log call -------------------------------
WARNING  flake8.options.manager:manager.py:207 option --max-complexity: please update from optparse string `type=` to argparse callable `type=` -- this will be an error in the future
_________________________________ test_origin __________________________________

    def test_origin():
>       assert pep517.__file__ != os.path.abspath(os.path.join('pep517', '__init__.py'))
E       AssertionError: assert '/home/anubis/git/pep517/pep517/__init__.py' != '/home/anubis/git/pep517/pep517/__init__.py'
E        +  where '/home/anubis/git/pep517/pep517/__init__.py' = pep517.__file__
E        +  and   '/home/anubis/git/pep517/pep517/__init__.py' = <function abspath at 0x7f44b9dc9c10>('pep517/__init__.py')
E        +    where <function abspath at 0x7f44b9dc9c10> = <module 'posixpath' from '/usr/lib/python3.8/posixpath.py'>.abspath
E        +      where <module 'posixpath' from '/usr/lib/python3.8/posixpath.py'> = os.path
E        +    and   'pep517/__init__.py' = <function join at 0x7f44b9dc9550>('pep517', '__init__.py')
E        +      where <function join at 0x7f44b9dc9550> = <module 'posixpath' from '/usr/lib/python3.8/posixpath.py'>.join
E        +        where <module 'posixpath' from '/usr/lib/python3.8/posixpath.py'> = os.path

tests/test_origin.py:7: AssertionError
=========================== short test summary info ============================
FAILED tests/test_origin.py::FLAKE8
FAILED tests/test_origin.py::test_origin - AssertionError: assert '/home/anub...
=================== 2 failed, 24 passed, 22 skipped in 5.74s ===================
ERROR: InvocationError for command /home/anubis/git/pep517/.tox/py38/bin/pytest (exited with code 1)
___________________________________ summary ____________________________________
ERROR:   py38: commands failed
```

Anyway, after another debugging session, I managed to get it to work. The configuration is derived from attrs.

They have the following:
```
[coverage:run]
source = attr

[coverage:paths]
source =
    src
    .tox/*/site-packages
```

So I tried
```
[coverage:run]
source = build

[coverage:paths]
source =
    src
    .nox/*/site-packages
```
But that doesn't work, coverage does not find anything. I am not sure what kind of filter weirdness is happening here.

If I remove the run source and just leave
```
[coverage:paths]
source =
    src
    .nox/*/site-packages
```
it works...

As that also includes the `tests` directory, I added an `omit` entry. Our final configuration:
```
[coverage:run]
omit =
    tests/*
    tests\*

[coverage:paths]
source =
    src
    */site-packages
    *\site-packages
```

The PR is finally ready for review, uff...

.gitmodules
FFY00: I agree, but this should only affect people wanting to run the tests, I would not consider them users. Cloning each repo during the tests is wasteful, depending on how you do it, and potentially error prune. If we don't want to throw the cloned directories away, it is very difficult to manage it a thread safe manner. Running tests in parallel, either in pytest, tox or nox, would cause issues. I think using submodules would be reasonable here, but I understand if you don't agree.

I can add a check in the tests with message with command to clone submodules. What do you think?

FFY00:
@pganssle sorry to bother, could you please review?

.gitmodules
gaborbernat: > Cloning each repo during the tests is wasteful, depending on how you do it, and potentially error prune.

I personally would prefer doing a shallow copy, .tar.gz download of master branch instead of this. I think that would be more robust. I'd personally would hide it behind a pytest marker, that could be disabled by Linux distributions when validating the package, or they could inject the source package via an environment variable. Submodules are going to be a PITA to maintain.

setup.cfg
```diff
@@ -49,6 +49,9 @@ test =
 [options.packages.find]
 where = src
 
+[tool:pytest]
+norecursedirs = tests/integration/*
```
gaborbernat, line 53: If you'd download these into a tempdir on test run you would not need this 🤔

setup.cfg
```diff
@@ -63,3 +66,14 @@ lines_between_types = 1
 lines_after_imports = 2
 known_first_party = build
 known_third_party = nox,pallets_sphinx_themes,pep517,pytest,setuptools,toml
+
+[coverage:run]
+omit =
```
gaborbernat, line 71: Why do we need to omit tests? I think we do care that test have good coverage. To ensure that we don't have tests that do nothing.

src/build/env.py
```diff
@@ -11,6 +11,10 @@
     from typing import Dict, Optional, Iterable, Type
 
 
+if sys.version_info[0] == 2:  # pragma: no cover
+    FileExistsError = OSError
```
gaborbernat, line 15: This is kinda suprising way to handle this, though I suppose is correct 👍

src/build/env.py
```diff
@@ -38,6 +42,24 @@ def _restore_env(self):  # type: () -> None
...
+        if prefix and path and path.startswith(prefix):
+            new_path = os.path.join(self._path, path[len(prefix + os.pathsep):])
+            if not os.path.exists(new_path):
+                try:
+                    os.makedirs(os.path.dirname(new_path))
+                except FileExistsError:
+                    pass
+                if os.name == 'nt':
+                    import shutil
+                    shutil.copytree(path, new_path)
```
gaborbernat, line 59: Windows does support symlinks for a while now https://blogs.windows.com/windowsdeveloper/2016/12/02/symlinks-windows-10/ 🤔 This is kinda presemptious it does not. Here's what virtualenv does https://github.com/pypa/virtualenv/blob/master/src/virtualenv/info.py#L32-L51 to check for this.

src/build/env.py
```diff
@@ -38,6 +42,24 @@ def _restore_env(self):  # type: () -> None
...
+    def _symlink_relative(self, path):  # type: (Optional[str]) -> None
+        if not path:  # pragma: no cover
+            return
+
+        prefix = sysconfig.get_config_var('prefix')
+        if prefix and path and path.startswith(prefix):
+            new_path = os.path.join(self._path, path[len(prefix + os.pathsep):])
+            if not os.path.exists(new_path):
+                try:
+                    os.makedirs(os.path.dirname(new_path))
```
gaborbernat, line 54: Why not just use exists_ok flag here?

gaborbernat: Could you need here ``parents=True``?

tests/conftest.py
```diff
@@ -1,42 +1,83 @@
...
+    parser.addoption('--run-integration', action='store_true', help='run the integration tests')
+    parser.addoption('--only-integration', action='store_true', help='only run the integration tests')
+
+
+def pytest_collection_modifyitems(config, items):
+    skip_int = pytest.mark.skip(reason='integration tests not run')
+    skip_other = pytest.mark.skip(reason='only integration tests are run')
+
+    for item in items:
+        if os.path.basename(item.location[0]) == 'test_integration.py':
```
gaborbernat, line 22: maybe better to have an integration folder rather than a single file 🤔

tests/conftest.py
```diff
@@ -1,42 +1,83 @@
...
 import pytest
 
 
-DUMMY_PYPROJECT = '''
-[build-system]
-requires = ['flit_core']
-build-backend = 'flit_core.buildapi'
-'''.strip()
+def pytest_addoption(parser):
+    parser.addoption('--run-integration', action='store_true', help='run the integration tests')
```
gaborbernat, line 13: I find a better approach is to create an integration folder. That way users can do integration only via ``-k integration``. Then we just need to skip by default all tests in that folder, but enable them if user passed a ```--integration`` flag.

tests/conftest.py
```diff
@@ -1,42 +1,83 @@
...
+        if os.path.basename(item.location[0]) == 'test_integration.py':
+            if not config.getoption('--run-integration') and not config.getoption('--only-integration'):
+                item.add_marker(skip_int)
+        elif config.getoption('--only-integration'):
+            item.add_marker(skip_other)
+
+
+@pytest.fixture
+def packages_path():
+    return os.path.realpath(os.path.join(__file__, '..', 'packages'))
```
gaborbernat, line 31: there's some duplication here, maybe define a ``HERE`` global viarable that contains this folder.

tests/test_integration.py
```diff
@@ -0,0 +1,57 @@
...
+)
+@pytest.mark.parametrize(
+    ('args'),
+    [
+        [],
+        ['-x', '--no-isolation'],
+    ]
+)
+def test_build(tmpdir, integration_path, project, args):
+    os.environ['SETUPTOOLS_SCM_PRETEND_VERSION'] = 'dummy'  # for the projects that use setuptools_scm
```
gaborbernat, line 33: use pytests monkeypatch.setenv instead of atlering your os environ. This is needed to ensure that subsequent tests are not affected by changes to a global state in this test.

tests/test_integration.py
```diff
@@ -0,0 +1,57 @@
...
+
+    if project == 'python-build':  # windows does not support symlinks
```
gaborbernat, line 35: yes it does, see my comments from above

setup.cfg
pganssle: Agreed. I have found more than a few bugs where I was shadowing a test or something because my test coverage was less than 100%.

src/build/env.py
pganssle: I think `exists_ok=True` was added in Python 3.2. As long as this package still supports 2.7, we can't use it (though maybe it could be moved into a compatibility function) to make it easier to replace later. I think `parents` is not a valid argument to this function (it always creates the parent directories, which is the difference between `makedirs` and `mkdir`.

.gitmodules
pganssle: Yeah, I agree about a shallow git clone. I think cloning everything into a tempdir is the right place to start and if we decide that's taking up a lot of resources, we can move it into a private directory at the repo root or something.

That said, I also don't love the idea of a bunch of mocks and a bunch of integration tests as our testing strategies. I think we should just check in a bunch of very simple projects with various different configurations directly.

In the long run, I think that integration testing against real-world projects should probably be handled at a different layer similar to CPython's buildbots. Maybe that's https://github.com/pypa/integration-test, or maybe one day those of us who maintain systematically important projects like this can get our shit together and make something that tests your change against the top N projects on PyPI.

setup.cfg
FFY00: But cloning to a tempdir is wastefull, unless you want to go for a semi-temp directory, which would avoid re-downloading between runs.

src/build/env.py
FFY00: Well, I tried to unbundle from the rest of the code.

src/build/env.py
FFY00: IIRC the github actions failed when using symlinks.

> This new symlinks support first shipped in Windows 10 Insiders Build 14972, and will be formally delivered in Windows 10 Creators Update.

For older versions, they need to be manually enabled. I want to make sure it works for everyone, so I went with the Windows approach of just copying everything everywhere.

FFY00: We can add support for checking if symlinks are available, but I'd prefer to do it in another PR as this is already huge as is. And that would only be an optimization.

src/build/env.py
FFY00: Correct, Python 2.7 does not have `exists_ok`.

tests/conftest.py
FFY00: There is nothing right now warranting an integration folder, when there is, this can be change. Making a `integration` folder for these tests would be an premature optimization.

setup.cfg
gaborbernat: It's only integration tests, so their not expected to be run at every invocation; so it's not that bad. Though they could be semi-temp. Alternatively, consider commiting the ``.tar.gz`` and extracting that into a temp folder. And yes you should be recreating the structure of these test projects at every run to ensure that build happens on a clean start, so you don't pass because of some left over files from a previous run.

src/build/env.py
gaborbernat: If it's only an optimization drop it for Linux too, always copy. There's no reason to start discrimnating on behaviour between OSs.

setup.cfg
pganssle: I don't think it's any more wasteful on CI than using submodules, since you do a clone of the submodules on each run anyway, but it is true that it might be annoying locally if you have no cache. I think a semi-temporary directory is fine as well.

src/build/env.py
gaborbernat: > For older versions, they need to be manually enabled. I want to make sure it works for everyone, so I went with the Windows approach of just copying everything everywhere.

What older versions? Older Windows? That's the point of my test, I actively check if it works or not. And use copy if symlink does not work.

gaborbernat:
Oh and yeah there's a merge conflict on ``setup.cfg``, you'd want to resolve that too.

setup.cfg
gaborbernat: > I don't think it's any more wasteful on CI than using submodules, since you do a clone of the submodules on each run anyway, but it is true that it might be annoying locally if you have no cache. I think a semi-temporary directory is fine as well.

Note you'd need to implement clean functionality on semi-temporary directories, which adds code overhead. You don't need to worry about this if you e.g. just commit a .tar.gz and extract that on every run.

tests/conftest.py
FFY00: `pytest ... tests/test_integration.py` would also work. All these are possible options, I have a slight preference for the current approach since everything is handled the same way. Is there any benefit of handling it as you propose?

gaborbernat: Reduces the size of API needed to be maintained within ``conftest.py``, and keeps the API more towards what pytest offers, rather than adding new flags/behaviours by a bit.

src/build/env.py
FFY00: In Linux we can safely assume it works, Windows is problematic. The check is pretty small and reasonable, I think. It is already an optimization, it is not perfect but gives us good result for a small complexity hit. I can remove the behavior but that feels like moving backwards to me, let me know if you want me to do it.

> What older versions? Older Windows? That's the point of my test, I actively check if it works or not. And use copy if symlink does not work.

Yes, older Windows. I am not saying we shouldn't add a test like yours, we totally should, I am just saying that this PR is already big as is and is severely blocking future progress. As this is just a small optimization with small impact on the rest of the project, I just suggested doing it later, once we sort out the more pressing issues. What is your position on this? In your opinion, does this make sense?

tests/conftest.py
FFY00: We can do that. I usually avoid doing small optimizations like that when it's like two instances. If you notice here, that de-duplication would actually result in a bigger file. It also adds a complexity factor of having to check, or remember, what `HERE` is when we read the code.

Btw, I am not saying this just to be picky or something. I am establishing a precedent for these cases, this is also what I am doing for most other comments. This way we can come up to a conclusion and be on the same page for similar things in the future.

src/build/env.py
gaborbernat: > I can remove the behavior but that feels like moving backwards to me, let me know if you want me to do it.

I'd prefer to just pull in the code as is from virtualenv and do the prefer symlink use copy.

> I am just saying that this PR is already big as is and is severely blocking future progress.

Part of the reason it's blocking progress is because we're arguing about the scope. Rather than to do this, let's just fix the raised issues and merge it, and not have follow-up issues.

tests/conftest.py
gaborbernat: > or remember, what `HERE` is when we read the code.

It really would be just 3-5 lines above, so not that hard to find out; and in all projects I've been seing HERE used as the folder of the file, so it's kinda self-documenting. So I see a lot of precedence on going down this path.

setup.cfg
FFY00: > It's only integration tests, so their not expected to be run at every invocation; so it's not that bad. Though they could be semi-temp.

They aren't, but when they run, I want them to run fast. This is not a deal breaker but has 

> Alternatively, consider commiting the `.tar.gz` and extracting that into a temp folder.

Commit a `.tar.gz` to the repo? Is this what you mean?

> And yes you should be recreating the structure of these test projects at every run to ensure that build happens on a clean start, so you don't pass because of some left over files from a previous run.

Agreed.

> I don't think it's any more wasteful on CI than using submodules, since you do a clone of the submodules on each run anyway, but it is true that it might be annoying locally if you have no cache. I think a semi-temporary directory is fine as well.

It isn't, but it is wasteful on my machine. CI time is free, mine isn't. I don't want to download tarballs for all projects every time I want to run the integration tests, that will take a long time, and won't let me run the tests offline. Committing a tarball to the repo receives a big fat no from me, but I am not sure if this is what Bernát means. There are countless resources on the internet explaining why this is bad, but if anyone wants, I can expand on it. I guess we could use git lfs, but that is severely more PITA that submodules and Github doesn't give us much space. Which leaves us with the semi-temporary directory solution.

Personally I don't understand the issue of using submodules, those files are only needed by the developers, and we can have pytest detect if they are not present and print the clone command ready for you to copy+paste.

Anyway... it seems like having semi-temporary directory is an acceptable solution for everyone so lets go with it.

We can pull [`filelock`](https://github.com/benediktschmitt/py-filelock) as a test dependency and download the files to a semi-temporary directory. `filelock` is needed to handle test runs in parallel.

This semi-temporary directory could be cached by the CI to speed up things in the future.

tests/conftest.py
FFY00: Alright, that makes sense.

tests/conftest.py
FFY00: Yeah, I understand, but that is not the main point I am arguing against. That being that we only have two instances and with the code needed to declare the variable, the file would actually be bigger :sweat_smile:. The complexity point is just something I always consider in these situations, it might be pretty small, but it's there. I have also seen simple things like this accumulate to a point where it does make a difference.

With that said, if you really want, I can change it.

setup.cfg
gaborbernat: > Personally I don't understand the issue of using submodules

I'm suprised how you haven't been bitten/annoyed at submodules yet. But whenever I did use them quickly got to regret it. And feels to me @pganssle had the same experience.

> This semi-temporary directory could be cached by the CI to speed up things in the future.

Ok, but make sure to add some clean-feature so we can ensure that builds always start with clean semi-temporary folders. This is critical to ensure things work by design and not by leftovers from earlier runs.

FFY00: > I'm suprised how you haven't been bitten/annoyed at submodules yet. But whenever I did use them quickly got to regret it. And feels to me @pganssle had the same experience.

I have, but that is mitigated by having checks on the code.

> Ok, but make sure to add some clean-feature so we can ensure that builds always start with clean semi-temporary folders. This is critical to ensure things work by design and not by leftovers from earlier runs.

Of course, we would cache the semi-temporary directory with the tarballs, each tarball would be extracted to a temporary directory for the test run. We only keep the tarballs, so everything is definitely clean :grin:.

pganssle: > Ok, but make sure to add some clean-feature so we can ensure that builds always start with clean semi-temporary folders. This is critical to ensure things work by design and not by leftovers from earlier runs.

A cleaning feature should be easy enough, but I think we might be fine on this front anyway because nothing will be cached between CI runs, and end users can run `git clean` to clean all local directory caches.

Also, I think it makes sense to pin to an exact commit for each of these things to start with, for the purposes of reproducibility and stability. I don't think integration tests like this are a great fit for this current workflow, they're just considerably better than what we have now. Longer term I think the workflow should be revamped in such a way that we won't care as much about how stable the integration tests are, and hopefully by then we won't be worried about using any sort of semi-temporary directory or any strategy like that.

FFY00: The current tests are actually pinned to a commit. See https://github.com/FFY00/python-build/tree/tests-rework/tests/integration.

FFY00: Implemented the semi-temporary dir approach.

tests/conftest.py
FFY00: I didn't change this as I think the maintenance burden here is pretty low and the options do provide a nicer UX for contributors. The tradeoff seems pretty reasonable to me, but if you disagree I can change, let me know :grinning:

.github/workflows/tests.yml
```diff
@@ -43,9 +43,12 @@ jobs:
         pip install nox
         nox --version
 
-    - name: Run nox
+    - name: Run unit tests
       run: nox -s test-${{ matrix.python }}
 
+    - name: Run integration tests
+      run: nox -s integration-${{ matrix.python }}
```
gaborbernat, line 50: Wait, why do we need a new environment? Can't we just use the same, but have the ``--run-integration`` flag?

noxfile.py
```diff
@@ -20,10 +20,22 @@ def test(session):
...
-    session.run('pytest', '--cov',
+    session.run('pytest', '--cov', '--cov-config', 'setup.cfg',
                 f'--cov-report=html:{htmlcov_output}',
                 f'--cov-report=xml:{xmlcov_output}',
                 'tests/', *session.posargs)
+
+
+@nox.session(python=['2.7', '3.5', '3.6', '3.7', '3.8', 'pypy2', 'pypy3'])
+def integration(session):
+    htmlcov_output = os.path.join(session.virtualenv.location, 'htmlcov')
```
gaborbernat, line 33: I'm not on board with having integration tests under their own nox env, they should live in the same namespace as the unit ones, just have a flag to run/not run them 🤷

src/build/env.py
```diff
@@ -38,6 +69,24 @@ def _restore_env(self):  # type: () -> None
...
+                except FileExistsError:
+                    pass
+                if fs_supports_symlink():
+                    os.symlink(path, new_path)
+                else:
+                    import shutil
+                    shutil.copytree(path, new_path)
+
```
gaborbernat, line 89: In an ideal world would be nice to separate the test change and this functionality change here 🤔  As in separate PRs but 🤷 👍

tests/conftest.py
```diff
@@ -1,42 +1,139 @@
...
+}
 
 
-DUMMY_PYPROJECT_NO_BACKEND = '''
-[build-system]
-requires = []
-'''.strip()
+def pytest_addoption(parser):
+    parser.addoption('--run-integration', action='store_true', help='run the integration tests')
+    parser.addoption('--only-integration', action='store_true', help='only run the integration tests')
```
gaborbernat, line 32: These two should be put within an exclusive group, otherwise passing them at the same time will cause confusion.

tests/conftest.py
```diff
@@ -1,42 +1,139 @@
...
+    self_source = os.path.abspath(os.path.join(__file__, '..', '..'))
+    self_dest = os.path.join(dest, 'python-build')
+    if not os.path.exists(self_dest):
+        if build.env.fs_supports_symlink():
+            os.symlink(self_source, self_dest)
+        else:  # pragma: no cover
+            shutil.copytree(self_source, self_dest)
+
+    for target, (repo, version) in INTEGRATION_SOURCES.items():
+        with filelock.FileLock(os.path.join(srcdir, '{}.lock'.format(target))):
```
gaborbernat, line 70: pep-8 would suggest to use ``src_dir``

tests/conftest.py
```diff
@@ -1,42 +1,139 @@
...
+
+@pytest.fixture(scope='session')
+def integration_path():
+    srcdir = os.path.realpath('.integration-sources')
+    dest = tempfile.mkdtemp(prefix='python-build-integration-')
+
+    if not os.path.exists(srcdir):
+        os.makedirs(srcdir)
+
+    # self
```
gaborbernat, line 60: I'm failing to interpret this comment.

tests/conftest.py
```diff
@@ -1,42 +1,139 @@
...
+
+    for target, (repo, version) in INTEGRATION_SOURCES.items():
+        with filelock.FileLock(os.path.join(srcdir, '{}.lock'.format(target))):
+            target_dest = os.path.join(dest, target)
+            if os.path.exists(target_dest):  # pragma: no cover
+                continue
+
+            tarball = os.path.join(srcdir, '{}.tar.gz'.format(target))
+            data = urlopen('https://github.com/{}/archive/{}.tar.gz'.format(repo, version)).read()
+            with open(tarball, 'wb') as f:
```
gaborbernat, line 77: instead of ``f`` use ``file_handler`` one character variables should be used only for list comprehension variables IMHO

tests/conftest.py
```diff
@@ -1,42 +1,139 @@
...
+    os.chmod(file, ~stat.S_IRWXU)
+
+    yield os.path.join(packages_path, 'test-no-permission')
+
+    os.chmod(file, orig_stat)
+
+
+@pytest.fixture
+def tmpdir():
+    path = tempfile.mkdtemp(prefix='build-test-')
```
gaborbernat, line 135: Are you overriding here pytest builtin ``tmp_dir``?

tests/test_integration.py
```diff
@@ -0,0 +1,54 @@
...
+
+    build.__main__.main([path, '-o', tmpdir] + args)
+
+    assert filter(_SDIST.match, os.listdir(tmpdir))
+    assert filter(_WHEEL.match, os.listdir(tmpdir))
+
+
+def test_isolation(tmpdir, test_flit_path, mocker):
+    try:
+        import flit_core  # noqa: F401
```
gaborbernat, line 45: Why would flit core not be always available?

src/build/env.py
FFY00: I'll do it in a separate PR.

gaborbernat: I'm ok for this to go in together now, but for future reference.

tests/conftest.py
FFY00: What do you mean exactly? They are already under a custom group in pytest ("custom options").

gaborbernat: https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_mutually_exclusive_group

src/build/env.py
FFY00: Yes, I would also like for the test to go with the code, but this PR is huge :grin:. And the code was copied from virtualenv and only slightly modified, so I am dencently confident with the implementation, but I still want a test.

tests/conftest.py
FFY00: We are handling the python-build source, hence `self`. We treat python-build differently since we use our own source instead of downloading the tarball.

tests/conftest.py
FFY00: Ah! That clarifies things, thanks!

tests/conftest.py
gaborbernat: This is totally not clear from that one line, maybe put a more long winded explanation of above? 👍

FFY00: Yes, I have changed the comment, I will push soon, I am just finishing handling the other comments.

tests/test_integration.py
FFY00: It's the opposite, if flit is available we can't properly test the isolation, so we skip it. The test won't run and show as missing coverage.

tests/conftest.py
FFY00: I disagree in this specific case, `f` is a pretty standard name for file context managers. And here we don't have any other file variable so it won't cause any confusion. But I can change if you want.

tests/conftest.py
FFY00:
```
  File "/home/anubis/git/python-build/tests/conftest.py", line 31, in pytest_addoption
    group = parser.add_mutually_exclusive_group()
AttributeError: 'Parser' object has no attribute 'add_mutually_exclusive_group'
```

Can I keep this out? The arguments actually work properly when used together.

tests/conftest.py
gaborbernat: I personally agree on this with PEP-8 and prefer it, but 🤷

tests/conftest.py
gaborbernat: Can you check why is that? Properly is a bit stretch as you can pass them at the same time and things will work.

FFY00: Why it still works when I pass them together or why the pytest parser is not a argparse instance?

They can be passed at the same time because the logic handles this.

```python
for item in items:
if os.path.basename(item.location[0]) == 'test_integration.py':
    if not config.getoption('--run-integration') and not config.getoption('--only-integration'):
        item.add_marker(skip_int)
    elif config.getoption('--only-integration'):
        item.add_marker(skip_other)
```

I would be okay making them mutually exclusive but also having them together. Having them together seems easier now.

tests/test_integration.py
FFY00: I put the xfail in the `try` instead of the `else` and added a comment. This should clarify things.

tests/conftest.py
gaborbernat: I get why they can be passed together, but I'd rather not allow that use case than make it ambigous what happens when you pass it. Both skip or run is valid in that case, so I'd rather raise an error than go ahead. So can you check how to make it exclusive flags with pytest?

FFY00: I added a manual check and raised `pytest.UsageError`.

tests/conftest.py
```diff
@@ -1,42 +1,142 @@
...
+def pytest_collection_modifyitems(config, items):
+    skip_int = pytest.mark.skip(reason='integration tests not run')
+    skip_other = pytest.mark.skip(reason='only integration tests are run')
+
+    if config.getoption('--run-integration') and config.getoption('--only-integration'):  # pragma: no cover
+        raise pytest.UsageError("--run-integration and --only-integration can't be used together, choose one")
+
+    for item in items:
+        if os.path.basename(item.location[0]) == 'test_integration.py':  # pragma: no cover
+            if not config.getoption('--run-integration') and not config.getoption('--only-integration'):
```
gaborbernat, line 44: Feels this can no longer happen so you can drop the second part, no?

FFY00: No, I still have to check both, both flags enable running the integration tests. I can only skip the tests if both are false.

gaborbernat: Oh, I'm not in love with having two flags for this, IMHO you're widening and complicating the UI needlessly 🤷

FFY00:
Ugh, why did you squash merge everything? This is a massive pull request and I added commit descriptions explaining certain things, to make it easier in the future.

FFY00:
I reverted the commit and added the single commits. Please be more careful in the future, or just leave the review approval and let me merge. I am sorry for being picky, but this one of the things I am most particular about.

gaborbernat:
Sorry for that, I'll dissapear then and leave the rest to you and Paul. TBF I only reviewed this because Paul asked me to, pushing the merge button was a mistake.

shrug

FFY00:
You don't need to disappear. This is something we had already talked about, that's why I was a bit annoyed. Getting used to other people and workflow is difficult, especially on open source where communication with other people is hard. I am sorry for the way I expressed my slight frustation.

PR #58, opened 2020-07-15
env: keep original PATH

FFY00:
This is required so that we can find programs such as GCC.

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -116,7 +116,15 @@ class IsolatedEnvironment(object):
             if path in sys_path:
                 sys_path.remove(path)
 
-        self._replace_env('PATH', self._get_env_path('scripts'))
+        env_scripts = self._get_env_path('scripts')
+        if env_scripts is None:  # pragma: no cover
+            raise RuntimeError('Missing scripts directory in sysconfig')
+
+        exe_path = [env_scripts]
+        if 'PATH' in os.environ:
+            exe_path.append(os.environ['PATH'])
+
+        self._replace_env('PATH', os.pathsep.join(exe_path))
         self._replace_env('PYTHONPATH', os.pathsep.join(sys_path))
         self._replace_env('PYTHONHOME', self._path)
 
--- a/tests/test_env.py
+++ b/tests/test_env.py
@@ -9,9 +9,10 @@ import build.env
 
 
 def test_isolated_environment(mocker):
+    old_path = os.environ['PATH']
     with build.env.IsolatedEnvironment() as env:
         if os.name != 'nt':
-            assert os.environ['PATH'] == os.path.join(env._path, 'bin')
+            assert os.environ['PATH'] == os.pathsep.join([os.path.join(env._path, 'bin'), old_path])
         assert os.environ['PYTHONHOME'] == env._path
 
         for path in ('purelib', 'platlib'):


src/build/env.py
```diff
@@ -67,7 +67,9 @@ def __enter__(self):  # type: () -> IsolatedEnvironment
             if path in sys_path:
                 sys_path.remove(path)
 
-        self._replace_env('PATH', self._get_env_path('scripts'))
+        self._replace_env('PATH', os.pathsep.join(
+            filter(None, [os.environ.get('PATH', None), self._get_env_path('scripts')]))
```
pganssle, line 71: Two things:

1. Would it make sense for `PATH` to come second here, so that anything installed in the build environment would take precedence over globally-installed tools?
2. `filter(None, ...)` will also filter out things like `''`, and apparently `PATH=''` is a valid setting (it makes it so that only the CWD is on the PATH). You can change the `None` to `lambda x: x is None`, or possibly it's cleaner to break it across multiple lines (can `self._get_env_path` ever return `None`?).

FFY00: 1) Oh yes, that was my intent, sorry.

2) Right, I forgot `filter` had this special handling for `None`. It is helpful most of the time, but not here.

tests/test_env.py
```diff
@@ -11,7 +11,7 @@
 def test_isolated_environment(mocker):
     with build.env.IsolatedEnvironment() as env:
         if os.name != 'nt':
-            assert os.environ['PATH'] == os.path.join(env._path, 'bin')
+            assert os.environ['PATH'].split(os.pathsep)[-1] == os.path.join(env._path, 'bin')
```
gaborbernat, line 14: This test should be altered to validate that the path entries before the last are exactly what they were when invoked. Otherwise if the code above deletes all original paths and just adds the scripts at the end thing still pass 🤔

src/build/env.py
gaborbernat: This change request here by @pganssle still needs addressed.

src/build/env.py
```diff
@@ -116,7 +116,15 @@ def __enter__(self):  # type: () -> IsolatedEnvironment
             if path in sys_path:
                 sys_path.remove(path)
 
-        self._replace_env('PATH', self._get_env_path('scripts'))
+        env_scripts = self._get_env_path('scripts')
+        if env_scripts is None:  # pragma: no cover
+            raise RuntimeError('Missing scripts directory in sysconfig')
```
gaborbernat, line 121: Can this be true? Do we need to raise for it?

FFY00: It should never be true, but I am making mypy happy. The possibility of a Python implementation not having a `scripts` directory is pretty slim, the only use cases I would imagine is something like MicroPython or Python for WebAssembly, but this package wouldn't be relevant there.

gaborbernat: you could also just use the typing.cast operation to cast away the None 🤷 but is good this way too 👍 the downside is no coverage for it though...

FFY00: I don't really care about the coverage of this particular check, hence the `# pragma: no cover`. I could use `typing.cast`, this just felt better and since it was such a simple implementation, I went for it. If it was more complex, I would probably have opted the other way :blush:.

PR #59, opened 2020-07-15
env: add include, platinclude and srcdir to the environment

FFY00:

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -11,6 +11,10 @@ if False:  # TYPE_CHECKING  # pragma: no cover
     from typing import Dict, Optional, Iterable, Type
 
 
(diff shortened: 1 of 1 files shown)


pganssle:
@FFY00 Do you have any background for this? What problem does it solve?

We should also probably set up [`towncrier`](https://pypi.org/project/towncrier/) or something similar for changelog management. Otherwise it will be very hard to write up our changelogs when it comes to release time.

FFY00:
It fixes building some native packages, which rely on these folders. It is part of #53.

I am aware of the changes, so when it is time I can write a changelog. But we could definitely set up towncrier, it is just not my priority right now.

FFY00:
Closing because this got merged in #56. Tests are added in #70.

PR #60, opened 2020-07-15
build: fix adding requirements when using the default backend

FFY00:

### DIFF
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -7,9 +7,10 @@ __version__ = '0.0.3.1'
 
 import importlib
 import os
+import re
 import sys
 
-from typing import Dict, List, Optional, Set, Union
+from typing import Dict, List, Optional, Sequence, Set, Union
 
 import pep517.wrappers
 import toml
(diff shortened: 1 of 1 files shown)


FFY00:
An alternative would be to just not add any requirements if the user specify then, instead of trying to merge them.

gaborbernat:
> This has no tests to validate the change. Also is missing some explanation in the commit message of what use case it's actually trying to address, under what circumstance.

Ping on this.

FFY00:
The use case is a `pyproject.toml` that looks like this:

```toml
[build-system]
requires = ['setuptools', 'wheel']
```

Which I've found in the wild. If we add the requirements for the default backend, which include `setuptools >= 40.8.0`, we have duplicated requirements 

Example:
```
$ pip install 'setuptools >= 40.8.0' setuptools wheel
ERROR: Double requirement given: setuptools (already in setuptools>=40.8.0, name='setuptools')
```

However, this is solved with `--use-feature=2020-resolver`. So I think we should just wait for pip and in the meantime spit out a warning if that happens.

PR #61, opened 2020-07-15
env: fix installing requirements with markers

FFY00:

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -150,5 +150,12 @@ class IsolatedEnvironment(object):
 
         subprocess.check_call([sys.executable, '-m', 'ensurepip'], cwd=self._path)
 
-        cmd = [sys.executable, '-m', 'pip', 'install', '--ignore-installed', '--prefix', self._path] + list(requirements)
-        subprocess.check_call(cmd)
+        with tempfile.NamedTemporaryFile('w+', prefix='build-reqs-', suffix='.txt', delete=False) as req_file:
+            req_file.write(os.linesep.join(requirements))
+            req_file.close()
+            cmd = [
+                sys.executable, '-m', 'pip', 'install', '--prefix',
+                self._path, '-r', os.path.abspath(req_file.name)
+            ]
+            subprocess.check_call(cmd)
+            os.unlink(req_file.name)
--- a/tests/test_env.py
+++ b/tests/test_env.py
@@ -61,6 +61,7 @@ def test_isolated_environment_install(mocker):
         env.install(['some', 'requirements'])
         if sys.version_info[:2] != (3, 5):
             subprocess.check_call.assert_called()
-        assert subprocess.check_call.call_args[0][0] == [
-            sys.executable, '-m', 'pip', 'install', '--ignore-installed', '--prefix', env._path, 'some', 'requirements'
+        args = subprocess.check_call.call_args[0][0]
+        assert args[:7] == [
+            sys.executable, '-m', 'pip', 'install', '--prefix', env._path, '-r'
         ]


src/build/env.py
```diff
@@ -89,5 +89,12 @@ def install(self, requirements):  # type: (Iterable[str]) -> None
 
         subprocess.check_call([sys.executable, '-m', 'ensurepip'], cwd=self._path)
 
-        cmd = [sys.executable, '-m', 'pip', 'install', '--ignore-installed', '--prefix', self._path] + list(requirements)
-        subprocess.check_call(cmd)
+        with tempfile.NamedTemporaryFile('w+', prefix='build-reqs-', suffix='.txt') as req_file:
+            req_file.write(os.linesep.join(requirements))
+            req_file.flush()
+            cmd = [
+                sys.executable, '-m', 'pip',
```
gaborbernat, line 96: This implies pip as install dependency... but it does not show up here https://github.com/FFY00/python-build/blob/master/setup.cfg#L24

src/build/env.py
```diff
@@ -89,5 +89,12 @@ def install(self, requirements):  # type: (Iterable[str]) -> None
...
+                'install', '--ignore-installed', '--prefix',
```
gaborbernat, line 97: would help to explain here why we need ignore installed

src/build/env.py
```diff
@@ -89,5 +89,12 @@ def install(self, requirements):  # type: (Iterable[str]) -> None
...
+                self._path, '-r', os.path.abspath(req_file.name)
+            ]
+            subprocess.check_call(cmd)
```
gaborbernat, line 100: Why do we need here a requirements.txt file and not just pass the dependencies as arguments?

src/build/env.py
FFY00: No? This is run inside the isolated environment, and we use ensurepip to get pip there.

src/build/env.py
FFY00: I guess I should have this in the comment message. Because pip install something does not support environment markers, such as `python_version`, which is exactly what I am fixing here.

src/build/env.py
FFY00: Hum, I guess this is leftover from when we didn't have full isolation.

src/build/env.py
gaborbernat: Can you comment that here, or link to where we use ensurepip as such?

FFY00: It's literally just up there. It is included in the preview.

```python
subprocess.check_call([sys.executable, '-m', 'ensurepip'], cwd=self._path)
```

src/build/env.py
gaborbernat: > `pip install something` does not support environment markers

I never experienced this, @pf_moore do we have a pip issue for this?

src/build/env.py
gaborbernat: So can we then remove it now?

src/build/env.py
pfmoore: I'm not sure why we should. I wouldn't expect to support environment markers on the command line, it seems like unnecessary generalisation to me. But I guess someone could raise a pip feature request if they felt like it (and had a compelling enough use case - I don't think this usage would count, using a requirements file seems fine if you want to support a wider range of command line forms than pip does).

gaborbernat: How does pip handle environment markers in PEP-518? Does it passes it in a requirement file?

pfmoore: Honestly, I don't know for sure, I'd have to look at the code.

pfmoore: The code in `build_env.py` doesn't do anything special. If you can produce a test case that demonstrated the issue, maybe it should be raised as a bug... (I'd be inclined to fix it by using a requirements file, though 😉)

gaborbernat: Guess we can proceed with this as it is then @FFY00

src/build/env.py
gaborbernat: Let's remove this ignore installed then and we should be good to go @FFY00

PR #65, opened 2020-07-22
build: include build system dependencies in check_dependencies

### LINKED ISSUE #64: Build doesn't check build requirements
If `pyproject.toml` includes a build dependency that isn't present in the system site-packages, this should be reported and the build aborted.

To demonstrate the issue, create a simple project that depends only on setuptools and wheel, but add `numpy` to `build-system.requires`. The build will succeed even though the declared dependencies are not present.

FFY00:
Fixes #64

### DIFF
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -142,6 +142,7 @@ class ProjectBuilder(object):
         :param distribution: Distribution to build (sdist or wheel)
         '''
         dependencies = self.get_dependencies(distribution)
+        dependencies.update(self.build_dependencies)
 
         return {dep for dep in dependencies if not check_version(dep)}
 
--- a/tests/test_projectbuilder.py
+++ b/tests/test_projectbuilder.py
@@ -132,8 +132,8 @@ def test_check_dependencies(mocker, pyproject_mock):
     builder.hook.get_requires_for_build_wheel.side_effect = copy.copy(side_effects)
 
     # requires = []
-    assert not builder.check_dependencies('sdist')
-    assert not builder.check_dependencies('wheel')
+    assert builder.check_dependencies('sdist') == {'flit_core'}
+    assert builder.check_dependencies('wheel') == {'flit_core'}
 
     # requires = ['something']
     assert builder.check_dependencies('sdist')


FFY00:
@pfmoore can you confirm this solves your issue?

pfmoore:
No, it still builds:

```
>type .\pyproject.toml
[build_system]
requires = ["setuptools >= 40.9.0", "wheel", "numpy"]
build-backend = "setuptools.build_meta"
PS 08:35 (e41e1642ae7e7ec) {00:00.041} C:\Work\Scratch\fff
>py -m build --no-isolation
running egg_info
writing fff.egg-info\PKG-INFO
writing dependency_links to fff.egg-info\dependency_links.txt
writing top-level names to fff.egg-info\top_level.txt
reading manifest file 'fff.egg-info\SOURCES.txt'
writing manifest file 'fff.egg-info\SOURCES.txt'
running sdist
running egg_info
writing fff.egg-info\PKG-INFO
writing dependency_links to fff.egg-info\dependency_links.txt
writing top-level names to fff.egg-info\top_level.txt
reading manifest file 'fff.egg-info\SOURCES.txt'
writing manifest file 'fff.egg-info\SOURCES.txt'
warning: sdist: standard file not found: should have one of README, README.rst, README.txt, README.md

running check
warning: check: missing required meta-data: url

warning: check: missing meta-data: either (author and author_email) or (maintainer and maintainer_email) must be supplied

creating fff-0.1
creating fff-0.1\fff.egg-info
copying files to fff-0.1...
copying pyproject.toml -> fff-0.1
copying setup.cfg -> fff-0.1
copying fff.egg-info\PKG-INFO -> fff-0.1\fff.egg-info
copying fff.egg-info\SOURCES.txt -> fff-0.1\fff.egg-info
copying fff.egg-info\dependency_links.txt -> fff-0.1\fff.egg-info
copying fff.egg-info\top_level.txt -> fff-0.1\fff.egg-info
Writing fff-0.1\setup.cfg
Creating tar archive
removing 'fff-0.1' (and everything under it)
running egg_info
writing fff.egg-info\PKG-INFO
writing dependency_links to fff.egg-info\dependency_links.txt
writing top-level names to fff.egg-info\top_level.txt
reading manifest file 'fff.egg-info\SOURCES.txt'
writing manifest file 'fff.egg-info\SOURCES.txt'
running bdist_wheel
running build
installing to build\bdist.win-amd64\wheel
running install
running install_egg_info
running egg_info
writing fff.egg-info\PKG-INFO
writing dependency_links to fff.egg-info\dependency_links.txt
writing top-level names to fff.egg-info\top_level.txt
reading manifest file 'fff.egg-info\SOURCES.txt'
writing manifest file 'fff.egg-info\SOURCES.txt'
Copying fff.egg-info to build\bdist.win-amd64\wheel\.\fff-0.1-py3.8.egg-info
running install_scripts
creating build\bdist.win-amd64\wheel\fff-0.1.dist-info\WHEEL
creating 'C:\Work\Scratch\fff\dist\tmpojzd5i_d\fff-0.1-py3-none-any.whl' and adding 'build\bdist.win-amd64\wheel' to it
adding 'fff-0.1.dist-info/METADATA'
adding 'fff-0.1.dist-info/WHEEL'
adding 'fff-0.1.dist-info/top_level.txt'
adding 'fff-0.1.dist-info/RECORD'
removing build\bdist.win-amd64\wheel
>pip list
Package    Version
---------- -------
build      0.0.3.1
packaging  20.4
pep517     0.8.2
pip        20.1.1
pyparsing  2.4.7
setuptools 46.4.0
six        1.15.0
toml       0.10.1
wheel      0.34.2
```

FFY00:
Is `build_system` a correct spelling for the build system entry? PEP 517 only refers `build-system` ([here](https://www.python.org/dev/peps/pep-0517/#source-trees)). What is happening here is that python-build is not finding a `build-system` entry and defaulting to the default build system table. If you apply the following patch it works.

```patch
--- a/pyproject.toml	2020-07-23 14:39:18.747501208 +0100
+++ b/pyproject.toml	2020-07-23 14:39:13.667470128 +0100
@@ -1,3 +1,3 @@
-[build_system]
+[build-system]
 requires = ["setuptools >= 40.9.0", "wheel", "numpy"]
 build-backend = "setuptools.build_meta"
```

Perhaps we could add a message telling that we are using the default build system table to help with issues like this.

pfmoore:
OMG, I'm sorry for not spotting that! No, `build_system` is not a correct spelling, it was just me being dumb.

I did wonder, because my first attempt with the patched version used isolate mode (which is new compared to the release version I'd tried first) and that didn't install numpy either, which is far more strange. But rather than thinking "what have I done wrong", I simply concluded the issue hadn't been fixed. Doh 🙁 

With this fix, everything is now working perfectly.

A message saying "build-system not specified, using the default value" would have helped, yes - that's a good suggestion.

Thanks for your patience here, I'm feeling suitably embarrassed now 😊

FFY00:
> I did wonder, because my first attempt with the patched version used isolate mode (which is new compared to the release version I'd tried first) and that didn't install numpy either, which is far more strange.

Yeah, I really want to do a release to avoid people expecting the current behavior.

> A message saying "build-system not specified, using the default value" would have helped, yes - that's a good suggestion.

I am gonna add that to my list, we would need to setup a logging system in `ProjectBuilder` for that.

src/build/__init__.py
```diff
@@ -142,6 +142,7 @@ def check_dependencies(self, distribution):  # type: (str) -> Set[str]
         :param distribution: Distribution to build (sdist or wheel)
         '''
         dependencies = self.get_dependencies(distribution)
+        dependencies.update(self.build_dependencies)
```
gaborbernat, line 145: Do we still need this considering @pfmoore issues was a typo?

pfmoore: It is still needed. Without it, there's no warning/error in `--no-isolation` mode if a build requirement is missing (the build will probably fail, of course, but the point of my report was that I think an up-front check is better).

pganssle, line 145: This will fix issue #64, but it occurs to me that `self.get_dependencies(distribution)` will actually *fail* in the case where a dependency required to execute the PEP 517 hooks is missing.

I think this may be a fundamental problem with the `check_dependencies` API, though. You can't actually get the full set of missing dependencies if a subset of those dependencies are missing. It might be worth splitting this into two functions like `check_build_system_requires` and `check_build_artifact_dependencies`.

Another option would be to document that this doesn't return *all* the missing dependencies if the build dependencies aren't satisfied, but the problem there is that the caller has no way of knowing where it failed.

pganssle: I'll note that this is a downside of the mocking approach you've taken here. We would have caught this issue if you were running a real build, because the mock is returning an empty set even when in reality it would raise a missing backend error.

tests/test_projectbuilder.py
```diff
@@ -132,8 +132,8 @@ def test_check_dependencies(mocker, pyproject_mock):
     builder.hook.get_requires_for_build_wheel.side_effect = copy.copy(side_effects)
 
     # requires = []
```
pganssle, line 134: This comment appears to be inaccurate — it seems that `pyproject_mock` has `requires=['flit-core']`.

PR #67, opened 2020-08-04
Detach interpreter introspection from other logic

uranusjr:
This makes it easier to reuse IsolatedEnvironment on environments other than the current intrepreter, and likely also would make it easier to test since the introspection rules can not be more easily mocked.

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -8,7 +8,7 @@ import types
 
 
 if False:  # TYPE_CHECKING  # pragma: no cover
-    from typing import Dict, Optional, Iterable, Type
+    from typing import Dict, Optional, Iterable, Sequence, Type
 
 
 if sys.version_info[0] == 2:  # pragma: no cover
@@ -49,9 +49,12 @@ class IsolatedEnvironment(object):
     Non-standard paths injected directly to sys.path still be passed to the environment.
     '''
 
-    def __init__(self):  # type: () -> None
+    MANIPULATE_PATHS = ('purelib', 'platlib')
+
+    def __init__(self, remove_paths):  # type: (Sequence[str]) -> None
         self._env = {}  # type: Dict[str, Optional[str]]
         self._path = None  # type: Optional[str]
+        self._remove_paths = remove_paths
 
     @property
     def path(self):  # type: () -> str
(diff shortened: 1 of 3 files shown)


uranusjr:
The `remove_paths` argument is required, so any instantiation that results in an invalid `IsolatedEnvironment` is due to the user passing in incorrect parameters, which is as likely a mistake as can happen with subclassing. So IMO the two approaches do not really differ in this regard. But I don’t really care to judge API design preferences, and the `init_env()` approach is good enough for me.

src/build/env.py
```diff
@@ -18,8 +18,27 @@ class IsolatedEnvironment(object):
     Non-standard paths injected directly to sys.path still be passed to the environment.
     '''
 
-    def __init__(self):  # type: () -> None
+    MANIPULATE_PATHS = ('purelib', 'platlib')
+
+    def __init__(self, remove_paths):  # type: (Collection[str]) -> None
```
FFY00, line 23: Maybe change from `Collection` to `Iterable` as `Collection` is not available in Python 3.5.

uranusjr:
My use case is that I want to run isolated build on another interpreter completely unrelated to the current one (that is importing this package). I’d do something like

```python
sysconfig_paths = subprocess.check_output([path_to_interpreter, "-c", "import json, sysconfig; print(json.dumps(sysconfig.get_paths()))").decode("ascii")

with IsolatedEnvironment(remove_paths=[sysconfig_paths["purelib"], sysconfig_paths["platlib"]]):
    custom_builder.build()
```

The same would be entirely possible with the `init_env()` approach, but less obvious since it is more difficult to identify where the parent class is doing environment introspection and override them. It would be a lot easier to know that the `IsolatedEnvironment` *won’t* do environment introspection, and I can use it without worrying about the parent class as long as I pass in the right values.

Another way to approach this would be to have two classes:

```python
class AgnosticIsolatedEnvironment:
    def __init__(self, remove_paths):
        ...

class IsolatedEnvironment(AgnosticIsolatedEnvironment):
    def __init__(self):
        remove_paths = ...  # Do introspections here.
        super().__init__(remove_paths)
```

> Forcing users to subclass to change this behavior would be my way of saying "you shouldn't really be messing up with this unless you are certain it is what you want to do". Does this make sense? I don't know, maybe I am being too picky, I'll sleep on it.

Composition vs. inheritance is an API design choice, and there is no right or wrong answer here IMO. The subclass design makes things easier (and less possible to get wrong) when users do no need to subclass, by trading the maintainability and flexibility when users do need to subclass. Composition creates more boilerplate (and more chance to write incorrect code), but is better at localsing logic, so users can understand by reading the final usage what is happening and when, without needing to trace the inheritance chain into the library.

FFY00:
I am not sure what to choose here. I have a slight preference for the inheritance approach, but I might be biased since that was my suggestion. @pganssle @gaborbernat what do you think we should do here?

FFY00:
Alright, @uranusjr can you rebase and fix the typing issue then :blush:?

gaborbernat:
ping @uranusjr

gaborbernat (approved):

PR #70, opened 2020-08-11
Better tests for `IsolatedEnvironment` and some minor fixes

FFY00:
This tests if the required environment directories are provided by
IsolatedEnvironment.

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -130,7 +130,7 @@ class IsolatedEnvironment(object):
 
         self._symlink_relative(sysconfig.get_path('include'))
         self._symlink_relative(sysconfig.get_path('platinclude'))
-        self._symlink_relative(sysconfig.get_config_var('srcdir'))
+        self._symlink_relative(sysconfig.get_config_var('LIBPL'))
 
         return self
 
--- a/tests/test_env.py
+++ b/tests/test_env.py
@@ -26,11 +26,19 @@ def test_isolated_environment_setup(mocker):
                 }
             ) in python_path
 
-        copy_path = (
+        copy_path = [
             sysconfig.get_path('include'),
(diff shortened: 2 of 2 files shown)


src/build/env.py
```diff
@@ -73,7 +73,7 @@ def _symlink_relative(self, path):  # type: (Optional[str]) -> None
         if not path:  # pragma: no cover
             return
 
-        prefix = sysconfig.get_config_var('prefix')
+        prefix = sysconfig.get_config_var('base' if sys.version_info[0] == 2 else 'prefix')
```
gaborbernat, line 76: Could be worth documenting why the discrepency?

src/build/env.py
```diff
@@ -130,7 +130,7 @@ def __enter__(self):  # type: () -> IsolatedEnvironment
 
         self._symlink_relative(sysconfig.get_path('include'))
         self._symlink_relative(sysconfig.get_path('platinclude'))
-        self._symlink_relative(sysconfig.get_config_var('srcdir'))
+        self._symlink_relative(sysconfig.get_config_var('LIBPL'))
```
gaborbernat, line 133: also we could document why this is needed 🤷

FFY00: Commit description

> LIBPL contains the files necessary for embedding Python in C, I wrongly
assumed srcdir was used for this. srcdir will point to LIBPL in Python
3, but not in Python2.

src/build/env.py
FFY00: Oh, it was leftover from my initial tries. I just checked and this is actually not needed. Thanks :blush:

src/build/env.py
gaborbernat: Ah, I see. Often I make it a code comment, as the commit history gets lost easier.

PR #74, opened 2020-08-18
env: export path as read-only

FFY00:

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -51,6 +51,14 @@ class IsolatedEnvironment(object):
 
     def __init__(self):  # type: () -> None
         self._env = {}  # type: Dict[str, Optional[str]]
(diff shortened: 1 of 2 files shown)


src/build/env.py
```diff
@@ -70,6 +75,7 @@ def _get_env_path(self, path):  # type: (str) -> Optional[str]
         return sysconfig.get_path(path, vars=self._env_vars)
 
     def _symlink_relative(self, path):  # type: (Optional[str]) -> None
+        assert self._path
```
gaborbernat, line 78: Why not raise an error? Assertions can be disabled, raising an error cannot?

FFY00: Right, sorry! Asserts shouldn't be used here.

gaborbernat (approved):

PR #81, opened 2020-09-05
build: do not check for the backend in when using in-tree backends

FFY00:
Right now we are trying to check for in-tree backends, which obviously
won't work.

We only have this check to provide nicer error messages to users, I
don't see the need to provide the same "nicer" error messages when using
in-tree backends as it is quite niche.

In these cases, pep517 will raise an error and we will print it.

### DIFF
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -133,10 +133,11 @@ class ProjectBuilder(object):
 
         self._backend = self._build_system['build-backend']
 
-        try:
-            importlib.import_module(self._backend.split(':')[0])
-        except ImportError:  # can't mock importlib.import_module  # pragma: no cover
-            raise BuildException("Backend '{}' is not available".format(self._backend))
+        if 'backend-path' not in self._build_system:
+            try:
+                importlib.import_module(self._backend.split(':')[0])
+            except ImportError:  # can't mock importlib.import_module  # pragma: no cover
+                raise BuildException("Backend '{}' is not available".format(self._backend))
 
         self.hook = pep517.wrappers.Pep517HookCaller(self.srcdir, self._backend,
                                                      backend_path=self._build_system.get('backend-path'))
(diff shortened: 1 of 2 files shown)


src/build/__init__.py
```diff
@@ -133,10 +133,11 @@ def __init__(self, srcdir='.', config_settings=None):  # type: (str, Optional[Co
...
         self._backend = self._build_system['build-backend']
 
-        try:
-            importlib.import_module(self._backend.split(':')[0])
-        except ImportError:  # can't mock importlib.import_module  # pragma: no cover
-            raise BuildException("Backend '{}' is not available".format(self._backend))
+        if 'backend-path' not in self._build_system:
+            try:
+                importlib.import_module(self._backend.split(':')[0])
+            except ImportError:  # can't mock importlib.import_module  # pragma: no cover
```
gaborbernat, line 139: Why can't we mock it?

FFY00: I couldn't get it to work. Maybe it was just be not doing something wrong, I remember that I spent a fair bit of time here, but couldn't make it work.

I gave it another spin today and it was actually pretty straight forward.

src/build/__init__.py
```diff
@@ -133,10 +133,11 @@ def __init__(self, srcdir='.', config_settings=None):  # type: (str, Optional[Co
 
         self._backend = self._build_system['build-backend']
 
-        try:
-            importlib.import_module(self._backend.split(':')[0])
-        except ImportError:  # can't mock importlib.import_module  # pragma: no cover
-            raise BuildException("Backend '{}' is not available".format(self._backend))
+        if 'backend-path' not in self._build_system:
+            try:
+                importlib.import_module(self._backend.split(':')[0])
```
gaborbernat, line 138: Arguably you could just do ``self._backend[self._backend.find(':')]`` 🤔 Which is more in spirit of what you want 🤔

gaborbernat (approved):

src/build/__init__.py
FFY00: What if there is no `:`? `find` will return `-1`, so `self._backend[-1]`, which will return the last character of the backend string :grin:. `split` will always return a list, if it doesn't find anything, it will return a list with the original string.

Issue #25, opened 2020-06-15
check_version does not verify extras


eli-schwartz:
https://github.com/FFY00/python-build/blob/360966b687c15d8357ba98f9e194ff0f877ce3f2/build/__init__.py#L62-L64

This doesn't do what it looks like it does. metadata.get_all merely checks to see if the project's PKG-INFO declares that such an extra exists. It doesn't check to see if the extra is enabled due to having all the required dependencies.

Instead, distribution(''....').requires should be used to recursively retrieve dependencies one level deeper, and all dependencies which are associated with the desired extra should be checked as well.

layday:
We don't check transitive dependencies at all, do we?  Extras just specify more dependencies that we don't check - the main issue is that `check_version` does not verify dependencies of declared build dependencies, though exras pose a problem in themselves.

PR #88, opened 2020-09-08
build: raise warning when we verify extras

FFY00:
See #25

### DIFF
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -54,6 +54,12 @@ class TypoWarning(Warning):
     '''
 
 
+class IncompleteCheckWarning(Warning):
+    '''
+    Warning raised when we have an incomplete check
+    '''
+
+
 def check_version(requirement_string, extra=''):  # type: (str, str) -> bool
     '''
     :param requirement_string: Requirement string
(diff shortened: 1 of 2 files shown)


src/build/__init__.py
```diff
@@ -83,6 +89,11 @@ def check_version(requirement_string, extra=''):  # type: (str, str) -> bool
     for extra in req.extras:
         if extra not in (metadata.get_all('Provides-Extra') or []):
             return False
+        warnings.warn(
+            "Verified that the '{}[{}]' extra is present but did not verify that it is active "
```
gaborbernat, line 93: What does active mean in this context?

gaborbernat: Shouldn't we actually validate instead as that ticket alludes to?

FFY00: We verify that the distribution provides the extra but do not verify if the dependencies required to make the extra active are met. This will require us to recursively verify extras.

I wanted to a proper dependency checker into https://github.com/pypa/packaging and use it here. See https://github.com/pypa/packaging/issues/317.

For more details, look at the issue in the commit message.

gaborbernat (approved):
Peding fixing that failing environment.

PR #91, opened 2020-09-09
env: pop PIP_REQUIRE_VIRTUALENV from isolated environment

### LINKED ISSUE #90: Unset PIP_REQUIRE_VIRTUALENV when running 'pip'
This is the same issue as in pep517: https://github.com/pypa/pep517/pull/84

If `PIP_REQUIRE_VIRTUALENV` is set (to a non-empty value) when running `python -m build` , `pip` errors with its "ERROR: Could not find an activated virtualenv (required)." message:

```
$ python -m build
Looking in links: /var/folders/kq/02p9h16n7zv_z7bhzmg6y_pm0000gn/T/tmpo1xlgb08
Processing /private/var/folders/kq/02p9h16n7zv_z7bhzmg6y_pm0000gn/T/tmpo1xlgb08/setuptools-47.1.0-py3-none-any.whl
Processing /private/var/folders/kq/02p9h16n7zv_z7bhzmg6y_pm0000gn/T/tmpo1xlgb08/pip-20.1.1-py2.py3-none-any.whl
Installing collected packages: setuptools, pip
Successfully installed pip setuptools
ERROR: Could not find an activated virtualenv (required).
  File "/Users/chainz/.pyenv/versions/3.8.5/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/Users/chainz/.pyenv/versions/3.8.5/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/Users/chainz/.pyenv/versions/3.8.5/lib/python3.8/site-packages/build/__main__.py", line 176, in <module>
    main(sys.argv[1:], 'python -m build')
  File "/Users/chainz/.pyenv/versions/3.8.5/lib/python3.8/site-packages/build/__main__.py", line 168, in main
    build(args.srcdir, args.outdir, distributions, config_settings, not args.no_isolation, args.skip_dependencies)
  File "/Users/chainz/.pyenv/versions/3.8.5/lib/python3.8/site-packages/build/__main__.py", line 80, in build
    _build_in_isolated_env(builder, outdir, distributions)
  File "/Users/chainz/.pyenv/versions/3.8.5/lib/python3.8/site-packages/build/__main__.py", line 45, in _build_in_isolated_env
    env.install(builder.build_dependencies)
  File "/Users/chainz/.pyenv/versions/3.8.5/lib/python3.8/site-packages/build/env.py", line 176, in install
    subprocess.check_call(cmd)
  File "/Users/chainz/.pyenv/versions/3.8.5/lib/python3.8/subprocess.py", line 364, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['/Users/chainz/.pyenv/versions/3.8.5/bin/python', '-m', 'pip', 'install', '--prefix', '/var/folders/kq/02p9h16n7zv_z7bhzmg6y_pm0000gn/T/build-env-9j1lejv3', '-r', '/var/folders/kq/02p9h16n7zv_z7bhzmg6y_pm0000gn/T/build-reqs-cc8vednw.txt']' returned non-zero exit status 3.
```

The solution is to unset the environment variable for the subprocess.

adamchainz:
Fixes #90.

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -86,6 +86,9 @@ class IsolatedEnvironment(object):
         self._env[key] = os.environ.get(key, None)
         os.environ[key] = new
 
+    def _pop_env(self, key):  # type: (str) -> None
+        self._env[key] = os.environ.pop(key, None)
+
     def _restore_env(self):  # type: () -> None
         for key, val in self._env.items():
             if val is None:
@@ -143,6 +146,7 @@ class IsolatedEnvironment(object):
         self._replace_env('PATH', os.pathsep.join(exe_path))
         self._replace_env('PYTHONPATH', os.pathsep.join(sys_path))
         self._replace_env('PYTHONHOME', self.path)
+        self._pop_env('PIP_REQUIRE_VIRTUALENV')
 
         self._symlink_relative(sysconfig.get_path('include'))
         self._symlink_relative(sysconfig.get_path('platinclude'))
--- a/tests/test_env.py
+++ b/tests/test_env.py
@@ -10,7 +10,7 @@ import pytest
 import build.env
 
 
-def test_isolated_environment_setup(mocker):
+def test_isolated_environment_setup():
     old_path = os.environ['PATH']
     with build.env.IsolatedEnvironment.for_current() as env:
         if os.name != 'nt':
@@ -51,6 +51,13 @@ def test_isolated_environment_setup(mocker):
                 relative_path = path[len(prefix + os.pathsep):]
                 path = os.path.join(env.path, relative_path)
             assert os.path.exists(path)
+
+
+def test_isolated_environment_setup_require_virtualenv(mocker):
+    mocker.patch.dict(os.environ, {"PIP_REQUIRE_VIRTUALENV": "true"})
+    with build.env.IsolatedEnvironment.for_current():
+        assert "PIP_REQUIRE_VIRTUALENV" not in os.environ
+    assert os.environ["PIP_REQUIRE_VIRTUALENV"] == "true"
 
 
 def test_isolated_environment_install(mocker):


adamchainz:
If this is found to break anything, the approach from https://github.com/pypa/pep517/pull/84/commits/a03b4e846281d6d6141f29d20650522743e5b82f could be copied

FFY00:
Hum, will `--isolated` break `PIP_USE_FEATURE=2020-resolver`? Maybe we should be setting that all ourselves when it is required.

@pradyunsg could you recommend anything here? In some cases we might end up with duplicated requirements and the dependency resolver would be able to handle that. Are any possible drawbacks to using the dependency resolver in all cases?

What I've been doing is.

```
PIP_USE_FEATURE=2020-resolver python -m build
```

But I agree here that passing `--isolated` would be good.

I think we should remove `PIP_REQUIRE_VIRTUALENV` regardless, can you please add that to the PR? Thanks!

adamchainz:
> Hum, will --isolated break PIP_USE_FEATURE=2020-resolver?

I imagine so

> I think we should remove PIP_REQUIRE_VIRTUALENV regardless, can you please add that to the PR? Thanks!

I don't see why we'd do both fixes? `PIP_REQUIRE_VIRTUALENV` has no effect when `--isolated` is passed.

pganssle:
I think we should go with the approach where we pop `PIP_REQUIRE_VIRTUALENV`.

To me, the whole purpose of the `PIP_*` environment variables is to allow you to configure `pip` invocations nested within other programs. For `PIP_REQUIRE_VIRTUALENV`, it makes sense to remove it because we _are_ installing to a virtual environment of sorts (I actually wonder if there's something we can do to make `pip` recognize this), so the actual criterion is satisfied.

For other things like the resolver, setting `--no-binary :all:`, etc, we don't want to be overriding people's `pip` settings.

FFY00:
Okay, so you can add a `_pop_env` method to `IsolatedEnvironment` that will pop the env if present and add it to `self._env` to be restored later. Please also add a test for this if you can :blush:

pradyunsg:
I see I've been mentioned here, but I don't have anything to add beyond what @pganssle said (all of which I agree with). :)

FFY00:
@pradyunsg my question for you was if it was a good idea to set `PIP_USE_FEATURE=2020-resolver` as we need it in some cases.

pradyunsg:
Unconditionally in `build`, I'd say don't.

pip's going to flip the switch soon, and we'll remove that flag eventually as well. Sticking to what pip does by default will be easier to explain and justify. It'd be a much nicer idea to let environment variables "percolate" into the pip call TBH.

FFY00:
@adamchainz you are just missing the type hint, `  # type: (str) -> None`, if you could add it it would be great! Thanks :grin:

adamchainz:
Use [github suggestions](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/commenting-on-a-pull-request) for such minutiae.

FFY00 (approved):

FFY00:
Thank you :blush:

Issue #69, opened 2020-08-08
Command does not make sure the output directory exists


uranusjr:
And it crashes when it doesn’t. This is particularly problematic since the default output directory (``$PWD/dist``) is usually absent in a newly checked out repository.

This works:

```
$ python -m build -s -o .
```

But this fails if ``./dist`` does not exist, with a cryptic message:

```
$ python -m build -s
Traceback (most recent call last):
  File "lib\subprocess.py", line 1307, in _execute_child
    hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
NotADirectoryError: [WinError 267] The directory name is invalid

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "build\__init__.py", line 143, in build
    raise BuildBackendException('Backend operation failed: {}'.format(e))
build.BuildBackendException: Backend operation failed: [WinError 267] The directory name is invalid

ERROR Backend operation failed: [WinError 267] The directory name is invalid
```

FFY00:
Can you confirm this is working on `master`?

uranusjr:
It works if I provide an absolute path to `-o`, but creates if the path is relative; the directory is created at the wrong location.

```
$ python -m build ./example-project/ -s -o output-directory
Traceback (most recent call last):
...
FileNotFoundError: [Errno 2] No such file or directory: '/.../example-project/output-directory/my_package-2.tar.gz'
...
$ ls
example-project
output-directory
```

FFY00:
This should be fixed now. Feel free to reopen if you are still having issues.

uranusjr:
I’m still running into problems.

(Start in the root directory of a valid project.)

```
$ cd (project-root)
$ mkdir foo
$ cd foo  # Now we are in (project-root)/foo
$ python -m build .. -s
...
$ ls
dist
$ ls dist
pkg-1.tar.gz
```

The distribution got created in the nested `(project-root)/foo/dist` (i.e. `dist` resolved against the working directory) instead of `(project-root)/dist` (`dist` resolved against the project root).

However:

```
$ cd (project-root)/foo
$ python -m build .. -s -d bar
$ ls dist
bar
$ ls bar  # Empty!
$ ls ..
bar   ...(other project files)
```

A custom `-d` value is resolved against the project root, instead of the working directory.

The behaviour should be entirely the other way around. The default `dist` directory should be located in the project root no matter what the current working directory is; a custom destination directory should be located relative to the current working directory, no matter where the project is.

[...]

PR #93, opened 2020-09-09
build: fix calling hooks with relative output directory

FFY00:

### DIFF
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -201,6 +201,7 @@ class ProjectBuilder(object):
         :param outdir: Outpur directory
         '''
         build = getattr(self.hook, 'build_{}'.format(distribution))
+        outdir = os.path.abspath(outdir)
 
         if os.path.exists(outdir):
             if not os.path.isdir(outdir):
--- a/tests/test_projectbuilder.py
+++ b/tests/test_projectbuilder.py
@@ -160,7 +160,7 @@ def test_working_directory(tmp_dir):
         assert os.path.realpath(os.curdir) == os.path.realpath(tmp_dir)
 
 
-def test_build(mocker, test_flit_path):
+def test_build(mocker, test_flit_path, tmp_dir):
     mocker.patch('importlib.import_module', autospec=True)
     mocker.patch('pep517.wrappers.Pep517HookCaller', autospec=True)
     mocker.patch('build._working_directory', autospec=True)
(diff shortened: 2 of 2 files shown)


FFY00:
@uranusjr I did some local testing and it appears to be working. Can you confirm that it also resolves the issue for you?

src/build/__init__.py
```diff
@@ -201,6 +201,7 @@ def build(self, distribution, outdir):  # type: (str, str) -> None
         :param outdir: Outpur directory
         '''
         build = getattr(self.hook, 'build_{}'.format(distribution))
+        outdir = os.path.realpath(outdir)
```
pganssle, line 204: Should this be `realpath` or `abspath`?

Not sure what the relative merits are of resolving vs. not resolving symbolic links.

tests/test_projectbuilder.py
```diff
@@ -233,6 +233,17 @@ def test_missing_outdir(mocker, tmp_dir, test_flit_path):
...
 
+def test_relative_outdir(mocker, tmp_dir, test_flit_path):
+    mocker.patch('importlib.import_module', autospec=True)
+    mocker.patch('pep517.wrappers.Pep517HookCaller', autospec=True)
+
+    builder = build.ProjectBuilder(test_flit_path)
+
+    builder.build('sdist', '.')
+
+    builder.hook.build_sdist.assert_called_with(os.path.realpath('.'), {})
```
pganssle, line 244: You know what the path is already, right? Can you instead do:

`assert_called_with(tmp_dir, {})` or whatever `.` is intended to resolve to?

That would help to detect bugs from `os.realpath` doing something different than you expect.

src/build/__init__.py
FFY00: Not sure. I went with `realpath` because Windows has given me some trouble with symlinks in the past, I figured it didn't matter that much anyway.

tests/test_projectbuilder.py
FFY00: Sure.

src/build/__init__.py
FFY00: Do you want me to change to `abspath`?

pganssle: Hard to tell which is better or what the failure modes are — I wasn't really sure which is better and thought I'd flag it for consideration.

I don't think it will be that big a deal if we get it wrong though. Either someone will complain because something broke in some configuration that involves symlinks or they'll complain about something like the path being recorded incorrectly or something. I suppose we'll find out!

FFY00: Yeah, I changed it to `abspath`, we can merge that and see if anyone complains. It's a low risk change anyway, I think.

pganssle (approved):
Seems like a trivial enough change, though I'm not convinced that the tests here really cover the scenario in #69 (they just cover the root cause).

I think we don't have any non-integration tests that would be able to cover this sort of thing, but eventually it'd be nice to do something like build out a minimal PEP 517 backend that we can use to ensure that building simple projects in simple ways works.

Issue #98, opened 2020-09-10
Missing backend error when PEP 518 requirements are not already satisfied


pganssle:
Using `build==0.0.4`, it appears that the `python-build` command must be activated in an environment where the build backend is already installed! As an example, creating a project with a `pyproject.toml` like this:

```toml
[build-system]
requires = ["setuptools", "wheel"]
build-backend = "setuptools.build_meta"
```
Then `touch setup.cfg`, and create a virtualenv like so:

```
virtualenv venv --no-setuptools --no-wheel
source venv/bin/activate
pip install build==0.0.4
```

Now when you invoke `python-build -w .`, you'll get:

```
$ python-build .
ERROR Backend 'setuptools.build_meta' is not available
```

I think the issue is that [`ProjectBuilder` checks the existence of the build backend](https://github.com/FFY00/python-build/blob/3d2491fb33abbcfd8ce514939668823bf0235dac/src/build/__init__.py#L160-L164), but in `__main__` it is [constructed before the environment is built](https://github.com/FFY00/python-build/blob/3d2491fb33abbcfd8ce514939668823bf0235dac/src/build/__main__.py#L77-L82).

I'll note that you can build a wheel just fine from an environment where `setuptools` but not `wheel` is installed, but if `setuptools` isn't installed ahead of time, it will cause problems.

This problem is related to but distinct from #97.

I think this definitely needs an integration test.

FFY00:
I think we should move that check to `build`.

Issue #97, opened 2020-09-10
Isolated build environment has access to build's environment when run from a Python environment


pganssle:
It seems that when constructing the isolated environment, `python-build` still uses the packages in the outer environment if they are present, rather than installing into a "clean" environment.

One consequence of this is that if you have an older version of your build backend in the environment that contains `python-build`, `pip install` won't install an upgraded version of it. For example, construct a project with a `pyproject.toml` like so:

```toml
[build-system]
requires = ["setuptools", "wheel"]
build-backend = "setuptools.build_meta"
```

And a `virtualenv` like so:

```python
virtualenv venv --no-wheel
source venv/bin/activate
pip install build==0.0.4
pip install setuptools==39.0.0
```

When you run `python-build -w .`, when satisfying `setuptools`, `pip` will not upgrade `setuptools`, since it is "already satisfied":

> Requirement already satisfied: setuptools in ./venv/lib/python3.8/site-packages (from -r /tmp/build-reqs-r6nyhzz2.txt (line 2)) (39.0.0)

Preferably, `pip` would consider the isolated environment a blank slate, and you'd get the latest `setuptools`, unless `--no-isolation` is passed.

This is not only a problem with `pip` — the build environment used contains all the packages from the environment `python-build` was invoked in, as you can see if you create a `setup.py` like so:

```python
import setuptools

try:
    import dateutil
    raise Exception("dateutil should not be present during isolated builds!")
except ImportError
    print("dateutil not found")

setuptools.setup()
```

If you run this in the virtualenv we created above, it will succeed and print the "dateutil not found" message. If you then run `pip install python-dateutil` in the virtualenv, the `python-build -w .` will fail, indicating that the build is being carried out in an environment not isolated from the base environment.

These results on Arch Linux using `virtualenv==20.0.26`, `build==0.0.4` and Python version 3.8.2 (via pyenv). I have not attempted to replicate this without a virtualenv.

FFY00:
Hum, I am pretty sure I was passing `--ignore-installed` to pip, maybe that got removed in one of the reviews. Can you check if it fixes your issue?

pganssle:
> Hum, I am pretty sure I was passing `--ignore-installed` to pip, maybe that got removed in one of the reviews. Can you check if it fixes your issue?

[...]

PR #99, opened 2020-09-11
env: fix islation when pyvenv.cfg is present

FFY00:

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -50,6 +50,9 @@ class IsolatedEnvironment(object):
     MANIPULATE_PATHS = ('purelib', 'platlib')
 
     def __init__(self, remove_paths):  # type: (Sequence[str]) -> None
+        '''
+        :param remove_paths: Import paths that should be removed from the environment
+        '''
         self._env = {}  # type: Dict[str, Optional[str]]
         self._path = None  # type: Optional[str]
         self._remove_paths = remove_paths
@@ -63,6 +66,9 @@ class IsolatedEnvironment(object):
 
     @classmethod
     def for_current(cls):  # type: () -> IsolatedEnvironment
+        '''
+        Creates an isolated environment for the current interpreter
+        '''
         remove_paths = os.environ.get('PYTHONPATH', '').split(os.pathsep)
 
         for path in cls.MANIPULATE_PATHS:
@@ -78,6 +84,9 @@ class IsolatedEnvironment(object):
         return cls(remove_paths)
 
     def _replace_env(self, key, new):  # type: (str, Optional[str]) -> None
+        '''
+        Replaces an environment variable
+        '''
         if not new:  # pragma: no cover
             return
 
@@ -85,9 +94,15 @@ class IsolatedEnvironment(object):
         os.environ[key] = new
 
     def _pop_env(self, key):  # type: (str) -> None
+        '''
+        Removes an environment variable
+        '''
         self._env[key] = os.environ.pop(key, None)
 
     def _restore_env(self):  # type: () -> None
+        '''
+        Restores the initial environment variables
+        '''
         for key, val in self._env.items():
             if val is None:
                 os.environ.pop(key, None)
@@ -95,9 +110,18 @@ class IsolatedEnvironment(object):
                 os.environ[key] = val
 
     def _get_env_path(self, path):  # type: (str) -> Optional[str]
+        '''
+        Returns sysconfig path from our environment
+        '''
         return sysconfig.get_path(path, vars=self._env_vars)
 
     def _place_file(self, path, new_subpath):  # type: (str, str) -> None
+        '''
+        Places a file into our environment
+
+        :param path: File to be symlinked
+        :param new_subpath: Subpath of target destination
+        '''
         new_path = os.path.join(self.path, new_subpath)
         dir_path = os.path.dirname(new_path)
 
@@ -110,6 +134,13 @@ class IsolatedEnvironment(object):
             shutil.copyfile(path, new_path)
 
     def _place_path_relative(self, path):  # type: (Optional[str]) -> None
+        '''
+        Places a path into our environment
+
+        The original prefix will be removed and replaced with our environmenmt's
+
+        If the path is not valid, nothing will happen
+        '''
         if not path:  # pragma: no cover
             return
 
@@ -127,6 +158,11 @@ class IsolatedEnvironment(object):
                     shutil.copytree(path, new_path)
 
     def __enter__(self):  # type: () -> IsolatedEnvironment
+        '''
+        Set up the environment
+
+        The environment path should be empty
+        '''
         self._path = tempfile.mkdtemp(prefix='build-env-')
         self._env_vars = {
             'base': self.path,
@@ -152,29 +188,54 @@ class IsolatedEnvironment(object):
         if 'PATH' in os.environ:
             exe_path.append(os.environ['PATH'])
 
+        '''
+        Place include, platinclude and LIBPL into our environment. They should
+        be present and may be used to compile native packages or similar.
+        '''
         self._place_path_relative(sysconfig.get_path('include'))
         self._place_path_relative(sysconfig.get_path('platinclude'))
         self._place_path_relative(sysconfig.get_config_var('LIBPL'))
 
+        '''
+        We use PYTHONHOME to relocate the Python installation to our environment,
+        but turns out this can't be blindely relied upon, *sigh*. If a pyvenv.cfg
+        file is present on the same directory than the binary or one level above,
+        data, platlib, platstdlib, purelib and scripts will be overritten. So we
+        need to make sure that doesn't happen, hence we symlink the Python
+        executable onto a path we control and use that binary for the enrionment.
+
+        Unfortunately, this forces us to override sys.executable. There is no
+        other way of doing it.
+        '''
         executable_dir = 'executable'
         executable_path = os.path.join(self.path, executable_dir)
         self._place_file(sys.executable, os.path.join(executable_dir, 'python'))
         exe_path.insert(0, executable_path)
-        
+
         self._executable = sys.executable
         sys.executable = os.path.join(executable_path, 'python')
 
+        # Replace PATH with our value for the environment
         self._replace_env('PATH', os.pathsep.join(sorted(
             set(exe_path), key=exe_path.index
         )))  # sorted is required for Python 2 and 3.5
+
+        # Inject the missing import paths via PYTHONPATH
         self._replace_env('PYTHONPATH', os.pathsep.join(sys_path))
+
+        # Point the Python interpreter to our environment
         self._replace_env('PYTHONHOME', self.path)
+
+        # Remove environment variables that intrefer with our use of pip
         self._pop_env('PIP_REQUIRE_VIRTUALENV')
 
         return self
 
     def __exit__(self, typ, value, traceback):
         # type: (Optional[Type[BaseException]], Optional[BaseException], Optional[types.TracebackType]) -> None
+        '''
+        Restores the everything to the original state
+        '''
         if self.path and os.path.isdir(self.path):
             shutil.rmtree(self.path)
 
@@ -183,7 +244,11 @@ class IsolatedEnvironment(object):
 
     def install(self, requirements):  # type: (Iterable[str]) -> None
         '''
-        Installs the specified requirements on the environment
+        Installs the specified PEP 508 requirements on the environment
+
+        Passing non PEP 508 strings will result in undefined behavior, you
+        *should not* rely on it. It is merely an implementation detail, it may
+        change any time without warning.
         '''
         if not requirements:
             return
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -115,7 +115,7 @@ class IsolatedEnvironment(object):
         '''
         return sysconfig.get_path(path, vars=self._env_vars)
 
-    def _place_file(self, path, new_subpath):  # type: (str, str) -> None
+    def _symlink_or_copy_file(self, path, new_subpath):  # type: (str, str) -> None
         '''
         Places a file into our environment
 
@@ -133,7 +133,7 @@ class IsolatedEnvironment(object):
         else:
             shutil.copyfile(path, new_path)
 
-    def _place_path_relative(self, path):  # type: (Optional[str]) -> None
+    def _symlink_or_copy_path_relative(self, path):  # type: (Optional[str]) -> None
         '''
         Places a path into our environment
 
@@ -192,9 +192,9 @@ class IsolatedEnvironment(object):
         Place include, platinclude and LIBPL into our environment. They should
         be present and may be used to compile native packages or similar.
         '''
-        self._place_path_relative(sysconfig.get_path('include'))
-        self._place_path_relative(sysconfig.get_path('platinclude'))
-        self._place_path_relative(sysconfig.get_config_var('LIBPL'))
+        self._symlink_or_copy_path_relative(sysconfig.get_path('include'))
+        self._symlink_or_copy_path_relative(sysconfig.get_path('platinclude'))
+        self._symlink_or_copy_path_relative(sysconfig.get_config_var('LIBPL'))
 
         '''
         We use PYTHONHOME to relocate the Python installation to our environment,
@@ -209,7 +209,7 @@ class IsolatedEnvironment(object):
         '''
         executable_dir = 'executable'
         executable_path = os.path.join(self.path, executable_dir)
-        self._place_file(sys.executable, os.path.join(executable_dir, 'python'))
+        self._symlink_or_copy_file(sys.executable, os.path.join(executable_dir, 'python'))
         exe_path.insert(0, executable_path)
 
         self._executable = sys.executable
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -1,3 +1,4 @@
+import collections
 import os
 import shutil
 import subprocess
@@ -216,9 +217,9 @@ class IsolatedEnvironment(object):
         sys.executable = os.path.join(executable_path, 'python')
 
         # Replace PATH with our value for the environment
-        self._replace_env('PATH', os.pathsep.join(sorted(
-            set(exe_path), key=exe_path.index
-        )))  # sorted is required for Python 2 and 3.5
+        self._replace_env('PATH', os.pathsep.join(
+            list(collections.OrderedDict.fromkeys(exe_path))  # remove duplicates, need OrderedDict on Python <3.7
+        ))
 
         # Inject the missing import paths via PYTHONPATH
         self._replace_env('PYTHONPATH', os.pathsep.join(sys_path))
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -97,6 +97,18 @@ class IsolatedEnvironment(object):
     def _get_env_path(self, path):  # type: (str) -> Optional[str]
         return sysconfig.get_path(path, vars=self._env_vars)
 
+    def _place_file(self, path, new_subpath):  # type: (str, str) -> None
+        new_path = os.path.join(self.path, new_subpath)
+        dir_path = os.path.dirname(new_path)
+
+        if not os.path.exists(dir_path):
+            os.makedirs(dir_path)
+
+        if fs_supports_symlink():
+            os.symlink(path, new_path)
+        else:
+            shutil.copyfile(path, new_path)
+
     def _place_path_relative(self, path):  # type: (Optional[str]) -> None
         if not path:  # pragma: no cover
             return
@@ -112,7 +124,6 @@ class IsolatedEnvironment(object):
                 if fs_supports_symlink():
                     os.symlink(path, new_path)
                 else:
-                    import shutil
                     shutil.copytree(path, new_path)
 
     def __enter__(self):  # type: () -> IsolatedEnvironment
@@ -141,14 +152,24 @@ class IsolatedEnvironment(object):
         if 'PATH' in os.environ:
             exe_path.append(os.environ['PATH'])
 
-        self._replace_env('PATH', os.pathsep.join(exe_path))
-        self._replace_env('PYTHONPATH', os.pathsep.join(sys_path))
-        self._replace_env('PYTHONHOME', self.path)
-        self._pop_env('PIP_REQUIRE_VIRTUALENV')
-
         self._place_path_relative(sysconfig.get_path('include'))
         self._place_path_relative(sysconfig.get_path('platinclude'))
         self._place_path_relative(sysconfig.get_config_var('LIBPL'))
+
+        executable_dir = 'executable'
+        executable_path = os.path.join(self.path, executable_dir)
+        self._place_file(sys.executable, os.path.join(executable_dir, 'python'))
+        exe_path.insert(0, executable_path)
+        
+        self._executable = sys.executable
+        sys.executable = os.path.join(executable_path, 'python')
+
+        self._replace_env('PATH', os.pathsep.join(sorted(
+            set(exe_path), key=exe_path.index
+        )))  # sorted is required for Python 2 and 3.5
+        self._replace_env('PYTHONPATH', os.pathsep.join(sys_path))
+        self._replace_env('PYTHONHOME', self.path)
+        self._pop_env('PIP_REQUIRE_VIRTUALENV')
 
         return self
 
@@ -158,6 +179,7 @@ class IsolatedEnvironment(object):
             shutil.rmtree(self.path)
 
         self._restore_env()
+        sys.executable = self._executable
 
     def install(self, requirements):  # type: (Iterable[str]) -> None
         '''
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -97,7 +97,7 @@ class IsolatedEnvironment(object):
     def _get_env_path(self, path):  # type: (str) -> Optional[str]
         return sysconfig.get_path(path, vars=self._env_vars)
 
-    def _symlink_relative(self, path):  # type: (Optional[str]) -> None
+    def _place_path_relative(self, path):  # type: (Optional[str]) -> None
         if not path:  # pragma: no cover
             return
 
@@ -146,9 +146,9 @@ class IsolatedEnvironment(object):
         self._replace_env('PYTHONHOME', self.path)
         self._pop_env('PIP_REQUIRE_VIRTUALENV')
 
-        self._symlink_relative(sysconfig.get_path('include'))
-        self._symlink_relative(sysconfig.get_path('platinclude'))
-        self._symlink_relative(sysconfig.get_config_var('LIBPL'))
+        self._place_path_relative(sysconfig.get_path('include'))
+        self._place_path_relative(sysconfig.get_path('platinclude'))
+        self._place_path_relative(sysconfig.get_config_var('LIBPL'))
 
         return self
 
--- a/tests/test_env.py
+++ b/tests/test_env.py
@@ -14,7 +14,11 @@ def test_isolated_environment_setup():
     old_path = os.environ['PATH']
     with build.env.IsolatedEnvironment.for_current() as env:
         if os.name != 'nt':
-            assert os.environ['PATH'] == os.pathsep.join([os.path.join(env.path, 'bin'), old_path])
+            assert os.environ['PATH'] == os.pathsep.join([
+                os.path.join(env.path, 'executable'),
+                os.path.join(env.path, 'bin'),
+                old_path
+            ])
         assert os.environ['PYTHONHOME'] == env.path
 
         python_path = os.environ['PYTHONPATH'].split(os.pathsep)
@@ -58,6 +62,13 @@ def test_isolated_environment_setup_require_virtualenv(mocker):
     with build.env.IsolatedEnvironment.for_current():
         assert "PIP_REQUIRE_VIRTUALENV" not in os.environ
     assert os.environ["PIP_REQUIRE_VIRTUALENV"] == "true"
+
+
+def test_isolated_environment_executable():
+    with build.env.IsolatedEnvironment.for_current() as env:
+        executable = os.path.join(env.path, 'executable', 'python')
+        assert sys.executable == executable
+    assert sys.executable != executable
 
 
 def test_isolated_environment_install(mocker):


pganssle:
WRT the approach taken here, it'd be good if @gaborbernat could weigh in on this, because I am far from confident in my understanding of how virtual environment magic works.

I do think that it would be a good idea to add at least one test that actually covers the isolation _behavior_, though. The current test only ensures that the monkey-patch works (which I would expect to change if we can come up with a better solution), but won't detect regressions if something else breaks the behavior.

Creating a basically empty project and installing it into a virtualenv, then trying to build a project from within that `virtualenv` that has `import <projectname>` in the `setup.py` should work. There are conceptual simplifications (some of which don't involve pulling in other packages), but they involve more complicated test rigs (e.g. build a small project with an in-tree backend that executes a random file, have the "imported project" build itself without using `setuptools`, etc). Might be worth building out some of those test rigs _anyway_, so that we aren't leaning so heavily on integration tests, but I think #97 (and #98) are probably important enough that it's not worth delaying a solution just to avoid integration tests.

src/build/env.py
```diff
@@ -97,6 +97,19 @@ def _restore_env(self):  # type: () -> None
     def _get_env_path(self, path):  # type: (str) -> Optional[str]
         return sysconfig.get_path(path, vars=self._env_vars)
 
+    def _symlink_file(self, path, new_subpath):  # type: (str, str) -> None
```
gaborbernat, line 100: A better name for this would be _symlink_or_copy, because it falls back to copy.

src/build/env.py
```diff
@@ -97,6 +97,19 @@ def _restore_env(self):  # type: () -> None
...
+        new_path = os.path.join(self.path, new_subpath)
+        dir_path = os.path.dirname(new_path)
+
+        if not os.path.exists(dir_path):
+            os.makedirs(dir_path)
+
+        if fs_supports_symlink():
+            os.symlink(path, new_path)
+        else:
+            import shutil
```
gaborbernat, line 110: this import should be at top levle, not in here IMHO.

src/build/env.py
```diff
@@ -141,15 +154,23 @@ def __enter__(self):  # type: () -> IsolatedEnvironment
...
-        self._replace_env('PATH', os.pathsep.join(exe_path))
-        self._replace_env('PYTHONPATH', os.pathsep.join(sys_path))
-        self._replace_env('PYTHONHOME', self.path)
-        self._pop_env('PIP_REQUIRE_VIRTUALENV')
-
         self._symlink_relative(sysconfig.get_path('include'))
         self._symlink_relative(sysconfig.get_path('platinclude'))
         self._symlink_relative(sysconfig.get_config_var('LIBPL'))
 
+        self._symlink_file(sys.executable, os.path.join('bin', 'python'))
```
gaborbernat, line 161: IMHO we should be using here the platform specific folder name and executable, instead of bin/python. We don't want to introduce UNIX-ism on Windows.

src/build/env.py
```diff
@@ -141,15 +154,23 @@ def __enter__(self):  # type: () -> IsolatedEnvironment
...
+        exe_path.insert(0, os.path.join(self.path, 'bin'))
```
gaborbernat, line 162: this path worth saving a variable, as you should reuse it two lines below.

src/build/env.py
gaborbernat: What if bin/python already exists, can it? Should we overwrite it, I don't think so. Is there any guarantee that ``self.path`` is an empty folder at the start of this run?

src/build/env.py
```diff
@@ -141,15 +154,23 @@ def __enter__(self):  # type: () -> IsolatedEnvironment
...
+
+        self._executable = sys.executable
+        sys.executable = os.path.join(self.path, 'bin', 'python')
```
gaborbernat, line 165: I'm not in love with overwritting the sys.executable. Should detail with a comment here why.

src/build/env.py
```diff
@@ -141,15 +154,23 @@ def __enter__(self):  # type: () -> IsolatedEnvironment
...
+
+        self._replace_env('PATH', os.pathsep.join(sorted(
+            set(exe_path), key=exe_path.index
+        )))  # sorted is required for Python 2 and 3.5
```
gaborbernat, line 169: Can we make exe_path an ordered dict instead and then we don't have to do this dance?

src/build/env.py
```diff
@@ -141,15 +154,23 @@ def __enter__(self):  # type: () -> IsolatedEnvironment
...
+        self._replace_env('PYTHONPATH', os.pathsep.join(sys_path))
+        self._replace_env('PYTHONHOME', self.path)
+        self._pop_env('PIP_REQUIRE_VIRTUALENV')
```
gaborbernat, line 172: The isolation mechanism is complicated enough that we should have a section that explains how things bound together, and what we're doing to achieve it. (Beside code, architectural level). E.g. Why we need to create bin/python, why we need to alter python path and python home, why we need to pop PIP_REQUIRE_VIRTUALENV. These are all questions we should document within the code. How does pyvenv.cfg is affecting things.

gaborbernat (approved):
You have a typo in the title (``islation``) and IMHO I'd like a better description on why this is needed, what are the moving parts within python that allow it to happen and why we need to alter some of the knobs you do. The CI is failing also.

> WRT the approach taken here, it'd be good if @gaborbernat could weigh in on this, because I am far from confident in my understanding of how virtual environment magic works.

I'd wait with judgment for @FFY00 to explain more in-depth why he took this approach, and what problems he's trying to fix and how.

gaborbernat (changes requested):
Wrong button, sorry!

src/build/env.py
FFY00: The intent here is to symlink, copy is just a last resort fallback. I stand by the original naming but I can change if you feel strongly about it.

src/build/env.py
FFY00: Sure. I just kept it here as it's last resort and won't be needed in the vast majority of the cases.

src/build/env.py
FFY00: Shouldn't exist, we are creating a new environment, everything should be empty.

What is the Windows name? AFAIK the executable is also called `python` on Windows.
About the directory name, it is not supposed to be a UNIX-ism. It is merely implementation detail, `IsolcatedEnvironment` is not supposed to be interoperable with other Python environments, that is not the scope of the project, the directory that holds the binaries could be called `environment_binaries` or something else, it is not relevant. What is relevant is that `python` is in `PATH`.

src/build/env.py
FFY00: I would have to use `os.path.join` anyway, so I didn't feel the need. I can change that if you still want.

```python
bin_path = os.path.join(self.path, 'bin')
exe_path.insert(0, bin_path)
sys.executable = os.path.join(bin_path, 'python')
```

src/build/env.py
FFY00: Me neither :confused:. I will add a comment.

src/build/env.py
FFY00: We can, however I would prefer to keep it this way. Using `sorted` is only required in Python 3.5 and older, because the order isn't kept when converting to a set. It is used here as a temporary solution.

As a maintainer, I prefer to keep workarounds like this, as long as they are simple, so that we do not need to change the code in the future. When we drop Python 2 and 3.5, we can simply removed the `sorted` usage. Maybe this is another place where you and me disagree as maintainers :confused:

src/build/env.py
FFY00: You are totally right! That is long overdue. It started simpler but over time things got more complicated. I will document this :blush:

src/build/env.py
FFY00: Ops, turns out it was already up there :see_no_evil:

src/build/env.py
gaborbernat: The thing is it can happen that it doesn't symlink in which case the name is missleading, so let's address that.

src/build/env.py
gaborbernat: If that's the case I prefer a name such as build or something, so it cannot be confused with the Unix bin script folder name.

src/build/env.py
gaborbernat: I personally prefer this over that .

src/build/env.py
FFY00: What about `place_file`? And similarly rename `symlink_relative` to `place_path_relative`?

src/build/env.py
FFY00: What about `executable`?

src/build/env.py
```diff
@@ -141,27 +188,67 @@ def __enter__(self):  # type: () -> IsolatedEnvironment
...
+        '''
+        Place include, platinclude and LIBPL into our environment. They should
+        be present and may be used to compile native packages or similar.
+        '''
+        self._place_path_relative(sysconfig.get_path('include'))
+        self._place_path_relative(sysconfig.get_path('platinclude'))
+        self._place_path_relative(sysconfig.get_config_var('LIBPL'))
+
+        '''
+        We use PYTHONHOME to relocate the Python installation to our environment,
```
gaborbernat, line 200: You need the pyvenv.cfg only on python 3.4+. Copy on python 2 will not work. We should raise if that happens. Same on python 3 that's signed (macos+ windows store). We basically don't support those environments?

gaborbernat (commented):
You're isolation logic basically tries to create a virtual environment, and breaks in a lot of situations. That might be ok if this was a personal project, but for a pypa one you need to support all major interpreters. Maybe just use virtualenv instead?

src/build/env.py
FFY00: Can you elaborate? We are not using `pyvenv.cfg`.

FFY00:
Can you describe how it breaks?

gaborbernat:
For starters windows store python can't be copied for security reasons. So when symlink is off for the is os this will break. Furthermore as far as I read cpython code it checks all parent folders, not just the immediate parent for pyvenv.cfg. it's more - complicated but I'm on my phone now.

FFY00:
> For starters windows store python can't be copied for security reasons. So when symlink is off for the is os this will break.

The number of systems without symlinking support is already pretty small, so this won't affect many users. I am okay with not supporting this use case.

How does virtualenv deal with this anyway?

> Furthermore as far as I read cpython code it checks all parent folders, not just the immediate parent for pyvenv.cfg. it's more - complicated but I'm on my phone now.

That is not documented, nor is the behavior I see. This could change between os, but on Linux, CPython is only checking the same directory and one directory up.

I would be happy to discuss this further. virtualenv is kind of a heavy dependency, so I am reluctant to pull it.

gaborbernat:
If you say not supporting windows without symlink is ok I'll need to retract my support for this joining pypa.

FFY00:
> If you say not supporting windows without symlink is ok I'll need to retract my support for this joining pypa.

*supporting Windows without symlink support when the interpreter cannot be copied (which is the case of Python distributed by the Windows store)

The core functionality of the package still works, you just can't build in isolation.

Please let me know how virtualenv solves this issue, we may be able to apply the same approach here.

gaborbernat:
Per PEP 517 building in isolated environments is considered core functionality. Maybe you can get away with using venv instead of virtualenv to be fair here. And then have some hack like this for python 2, where copy only might be enough.

src/build/env.py
pganssle: I feel like `symlink_or_copy` is the best, most descriptive name for this, to be honest. It's a little more verbose, but it's immediately obvious what the function would do, and I think it best serves the two important audiences: people who know that symlinks are not universally available and people who don't know that.

For people who know that symlinks aren't available on all platforms:

- `symlink_or_copy` makes it pretty obvious that the intention is to symlink, but that it falls back to copying (why else would there be any doubt as to which one you're doing)
- If it's just named `symlink`, I am suddenly in doubt as to whether it throws deliberately, fails or falls back to something else.

For people who don't know that symlinks are available on all platforms:

 - `symlink_or_copy` alerts them that they may not understand exactly what's going on in the function (why is there any doubt?), and that in some cases the file may be _copied_ (helping them debug their code more easily as they try and figure out _why_ it would sometimes be copied).
 - If it's named `symlink`, they may not realize that there are edge cases where symlinking won't take place — they won't realize that they don't quite understand what's going on here.

I understand the desire to be more succinct, but I think `place` is also bad. To me it brings up uncertainty as to whether you are creating these files, moving them or copying them. I wouldn't think that `place` describes symlinking at all, much less that symlink is the _more likely_ behavior.

src/build/env.py
pganssle: Why is sorted only required for Python 2 and 3.5? `set` is unordered in all versions of Python.

pganssle:
It definitely would be nice to avoid pulling in `virtualenv` as a dependency (since it exposes console scripts and a bunch of stuff like that), but `venv` is in the standard library and AFAICT is basically a stripped-down version of `virtualenv` _anyway_ (that satisfies all of our use cases).

If it's significantly non-trivial to get the isolation logic right, maybe we should use `venv` on Python 3 and either add a Python 2-only dependency on `virtualenv` or a hacked-together "best effort" version for Python 2 support.

With regards to platform support, I don't think PyPA has any particular requirements for universal support for all features for all platforms — I'll note that `pep517` doesn't support this kind of isolation properly _at all_. If we're not going the `venv` route, do we have any numbers on how many users are on a platform without symlinking support and unable to copy their interpreter? This isn't a situation like `pip` where you _really_ need to support every random platform out there because it's a core workflow. This tool is intended to be used for building artifacts _for distribution_ (for personal consumption you can do `pip wheel` to create a wheelhouse). We should aim for as wide support as possible (since you need to be able to run it on all the platforms you would distribute wheels for), but if this would impact very few users and there's an easy enough workaround (install and run from a different interpreter, create your own isolated environment to run this in, use a github action, etc), I don't think it's the end of the world.

gaborbernat:
I think using venv for python 3, and some hacked together for python 2 would be the easiest and best investment.

FFY00:
virtualenv should have exactly the same issue we do, so what I am interested here is to understand if the way it solves the issue is complex enough that would make sense for us to pull it as a dependency. If it isn't complicated we could possibly do the same thing here.

Both venv and virtualenv work by giving you a custom Python interpreter. We have already established that overriding `sys.executable` is bad and we want to move away from it, so I see no future in us using them. Any effort put moving to those solutions would be better used on fixing the current workflow in the upstream IMHO. If we want this to be solved properly, CPython will have to gain a way to be reliably set up with custom paths without depending on the interpreter. I don't think there is any way around that, please let me know if I missed something.

Now, I think the use cases that wouldn't be supported are obscure enough for us to justify it. And if we (or more likely, I) do manage to fix the current workflow in the Python interpreter, they would start to be supported.

As a maintainer, I think it is very important to keep in mind the scope of the project and understand that you can't do everything. I think supporting the use case in question will block progress for the project. If I had unlimited time, I would maintain all backend approaches, maximizing the use cases we support, but I don't. If anyone does, please let me know, I would be more than glad to have that.
Right now I think I should optimize my time, and where I think it is best used is in trying to fix this issue properly.

To finish, I just wanted to note that this is not a run, it's a marathon. Not supporting use cases right now doesn't mean we can't support them in the future. We need to save our breath to try to pass the finish line, not waste it trying to go fast. Sorry for this corny metaphor :sweat_smile:, that is just how I feel.

FFY00:
Oh, I think it's import point this out, in case anyone is forgetting, python-build would still be able to build packages in the environments in question, it's *just* that it won't be able to build in an isolated environment.

gaborbernat:
> Both venv and virtualenv work by giving you a custom Python interpreter.

Can you ellaborate on this, what does custom covers here?

> We have already established that overriding `sys.executable` is bad and we want to move away from it, so I see no future in us using them

Why do you need to override?

> If we want this to be solved properly, CPython will have to gain a way to be reliably set up with custom paths without depending on the interpreter.

Not sure I follow, but we need to support Python in general, not just CPython. Please don't exclude other Python implementations.

> And if we (or more likely, I) do manage to fix the current workflow in the Python interpreter, they would start to be supported.

Unlikely without tens of lines of extra code for every release type (version constrained, etc). The main reason virtualenv moved away from single file architecture is because they are way too many variations, and you end up with hundereds lines of code.

> As a maintainer, I think it is very important to keep in mind the scope of the project and understand that you can't do everything

Exactly. The scope of this project is to build a python package. Creating isolated python environments is a goal for virtual envrionment creation projects. So let's not try to adddress that within this project, but instead delegate that job to a tool that does this (venv/virtualenv). This means we can spend more time on addressing the project goals, less trying to get support for various python releases.

> Right now I think I should optimize my time, and where I think it is best used is in trying to fix this issue properly.

This is exactly my issue raised. IMHO you're trying to fix a dumpster fire, and I'm just warning if you go down this path you'll end up with few more of this. I'd recommend instead avoiding them by delegating isolation part to tools who are dedicated to do this.

> Not supporting use cases right now doesn't mean we can't support them in the future.

If you just switch to venv/virtualenv you'll no longer need to revisit this issue at any point in the future. You solve this and all future isolation issues in one go. Seems the efficient thing to do.

src/build/env.py
```diff
@@ -141,27 +188,67 @@ def __enter__(self):  # type: () -> IsolatedEnvironment
...
         '''
-        Installs the specified requirements on the environment
+        Installs the specified PEP 508 requirements on the environment
+
+        Passing non PEP 508 strings will result in undefined behavior, you
+        *should not* rely on it. It is merely an implementation detail, it may
+        change any time without warning.
         '''
         if not requirements:
             return
```
gaborbernat, line 254: I just realized that you are calling:

```
subprocess.check_call([sys.executable, '-m', 'ensurepip'], cwd=self.path)
```

This will install pip and setuptools. IMHO this is bad because now setuptools is by default provisioned and not specifying it within your build requires will silently pass. We should ensure only pip is installed in there 🤔 . Perhaps we should uninstall if not specified within the build-requires list. On handle the easy to make error of forgetting setuptools as build-require but build still passes.

FFY00:
> Can you ellaborate on this, what does custom covers here?

They give you their own Python interpreter.

> Why do you need to override?

Because `sys.executable` might be used withing the code, in fact this is exactly what `pep517` does.

> Not sure I follow, but we need to support Python in general, not just CPython. Please don't exclude other Python implementations.

Sorry, in Python.

> Unlikely without tens of lines of extra code for every release type (version constrained, etc). The main reason virtualenv moved away from single file architecture is because they are way too many variations, and you end up with hundereds lines of code.

Can you elaborate, I don't follow? Unless Python implementations are not following the PEP, I see no reason to have custom code for each release.

> Exactly. The scope of this project is to build a python package. Creating isolated python environments is a goal for virtual envrionment creation projects. So let's not try to adddress that within this project, but instead delegate that job to a tool that does this (venv/virtualenv). This means we can spend more time on addressing the project goals, less trying to get support for various python releases.

The scope of the project is to provide a CLI tool to build Python packages, that is easy to bootstrap. I understand that other people may not care about bootstrapability, and just vendor dependencies, but I do and that is exactly the reason I created this project. I want to make sure that Python environments are easily bootstrapble, preferring following all that PEP 517 says. You might also notice the 'correct' in the description, which means that we should do things correctly, I've been burned many times by projects not caring about this, especially on such critical components as build systems.

We shouldn't be relying on hacks such as overriding `sys.executable`, the documentation does not say it is allowed, it could be read-only for all I know. In practice, both on CPython and PyPy it can be overriden, but that might change in the future. Even though I think it relatively low risk, I don't want to be relying on that. So I won't make any long term decisions that explicitly requires us to rely on that :confused:

The long term plan can't be to use venv and virtualenv. But we could use them venv and virtualenv short term, I am not arguing with that. The long term plan would be to fix this use-case in Python itself.

The current short term effects are that a few obscure use cases are not supported, and I don't think that warrants me taking the time to move to venv/virtualenv. I would rather take that time to work on fixing the workflow in Python, it takes time and as soon as we do that, as soon we will have this problem fixed.

With that say, if anyone does want to take the time to change the code to venv/virtualenv, **please** be my guest, I would really appreciate it. But that *will not* be the long term plan.

> This is exactly my issue raised. IMHO you're trying to fix a dumpster fire, and I'm just warning if you go down this path you'll end up with few more of this. I'd recommend instead avoiding them by delegating isolation part to tools who are dedicated to do this.

My position on this is to wait, if the maintenance burden becomes higher than the burden of moving the code to venv/virtualenv, I will consider doing that.

> If you just switch to venv/virtualenv (I'd say use venv on py3.4+, virtualenv otherwise - pulling in virtualenv on python 2 seems good enough compromise considering it's EOL) you'll no longer need to revisit this issue at any point in the future. You solve this and all future isolation issues in one go. Seems the efficient thing to do.

As I just said above, I think that would be a fine *short term* approach. You are free to do it.

> Violating the requirement of PEP-518 should not be taken lighly IMHO. Building in isolated environment is not optional per that PEP.

I do not take it lightly.

src/build/env.py
FFY00: You are right, I hadn't noticed before. We should be able to just delete `setuptools*` from `site-packages`.

FFY00:
Also, we are not on 1.0.0, we are on 0.0.4, having rough edges is fine.

gaborbernat:
> Because `sys.executable` might be used withing the code, in fact this is exactly what `pep517` does.

If we invoke the build operations in a subprocess of the virtual environment (isolated) this is not a problem. You already do a lot of subprocess calls, so see no reason why not do it here too.

> They give you their own Python interpreter.

Well it's not really their own. It's a python environment that tweaks enough parts of the global python to hide away globally installed packages. Which also sounds to me exactly what an isolated python environment is.

> I understand that other people may not care about bootstrapability

I've suggested to use venv, that's part of the standard library so should not need boostrapping. I did say virtualenv for python 2, but if you'd not want that we can use some hackish solution like here.


> You might also notice the 'correct' in the description, which means that we should do things correctly, I've been burned many times by projects not caring about this, especially on such critical components as build systems.

I'm complelty on board with you on this. This is why I'm suggesting a better isolation system than what you have here. Because what's here is not correct when symlinks are not available, or the python executable is not read-able (but can be executed).

> The long term plan can't be to use venv and virtualenv.

Why not? Those systems have been designed especially to create isolated build environments, and venv does not require any custom boostraping. 

> My position on this is to wait, if the maintenance burden becomes higher than the burden of moving the code to venv/virtualenv, I will consider doing that.

I guess we fundamentally disagree on how one should go about supporting python interpreters. I prefer designed to work upfront, while you're aiming at wait until enough people complain to make your effort worthwhile. From my POV people trying to use this tool, and then finding out it does not work because it does not support their use cases is hurtful to the whole packaging ecosystem, because another thing that they should use but they cannot.

src/build/env.py
gaborbernat: > We should be able to just delete `setuptools*` from `site-packages`.

I don't think that's wise, you'd be relying there on implementation details of what pip uninstall does. Probably better to just use pip uninstall directly.

pganssle: WTF? Why does `ensurepip` install `setuptools`?

gaborbernat: Because in a source only non PEP-518 world ``pip install`` would break without a local setuptools.. Is my guess.

pganssle:
> The scope of the project is to provide a CLI tool to build Python packages, that is easy to bootstrap. I understand that other people may not care about bootstrapability, and just vendor dependencies, but I do and that is exactly the reason I created this project. I want to make sure that Python environments are easily bootstrapble, preferring following all that PEP 517 says. You might also notice the 'correct' in the description, which means that we should do things correctly, I've been burned many times by projects not caring about this, especially on such critical components as build systems.

I can understand `virtualenv` being a problem for bootstrapping, but `venv` is in the standard library. The reasons for wanting to bootstrap and devendor everything have to do with making sure you can apply (security, mostly) patches in a central location, which is actually a major plus for using `venv` — CPython is way more likely to get security bug reports and fixes like that than this small project, which would argue in favor of delegating to `venv`.

> We shouldn't be relying on hacks such as overriding `sys.executable`, the documentation does not say it is allowed, it could be read-only for all I know.

Yeah, @gaborbernat to be clear the `sys.executable` hack was to make it so that we aren't blocked on [this issue in `pep517`](https://github.com/pypa/pep517/issues/92). None of us are happy with it and it cannot be the long-term plan.

> The long term plan can't be to use venv and virtualenv. But we could use them venv and virtualenv short term, I am not arguing with that. The long term plan would be to fix this use-case in Python itself.

Anything we fix in Python would only hit Python 3.10+. Assuming `python-build` continues to support anything not EOL upstream (and we're already actually way past EOL on a few Python versions we're supporting), I don't think we'd be able to take advantage of anything of this nature for ~4-5 years, so I think "long term" may be very long indeed.

> The current short term effects are that a few obscure use cases are not supported, and I don't think that warrants me taking the time to move to venv/virtualenv. I would rather take that time to work on fixing the workflow in Python, it takes time and as soon as we do that, as soon we will have this problem fixed.
> 
> With that said, if anyone does want to take the time to change the code to venv/virtualenv, **please** be my guest, I would really appreciate it. But that _will not_ be the long term plan.

I think we're already doing _significantly_ better than `pep517`, and that this PR is a net improvement for this package. We definitely must not let the perfect be the enemy of the good here. I wonder if this might be a good compromise:

1. For the easiest cases, we maintain our own lightweight isolating virtual environment — for people who care about bootstrapping, it's easy enough to make their use case the easy case, since they control their build environments (and are generally linux distros anyway — plus `build` can be built as a universal wheel, so you can always bootstrap `build` in a specific environment and then use it to build other stuff on your constrained environments).
2. For the tougher cases, we fall back to `venv` on Python 3 and either `virtualenv` or just no significant isolation support on Python 2 (we can raise a warning to that effect).

The upside of this is that it should help us avoid the complicated logic involved in supporting dozens of different platforms while also giving _most users_ the benefits of not using `venv` (we could also support an opt-in option to use `venv` unconditionally — which would certainly make it easier to test if it's hard to generate these obscure platforms). The downside here is that it's at least predicated on our ability to detect that we're in an unusual situation so we know when to fall back — and _detecting_ that may end up being almost as complicated as supporting all the other platforms.

Another option that is possibly more palatable would be to default to using `venv`, but add an option to use the lighter weight option for people who want to avoid it for whatever reason.

pganssle:
Actually, another thought occurs — @FFY00, wouldn't we be expecting all the people who care about ease of bootstrapping be using `--no-isolation` anyway? No point in avoiding a dependency on `venv` if the only people who would care about that anyway are Linux distros planning to create their own environments that don't rely on our isolation code.

FFY00:
> If we invoke the build operations in a subprocess of the virtual environment (isolated) this is not a problem. You already do a lot of subprocess calls, so see no reason why not do it here too.

I want to avoid using subprocess calls here as it makes it much harder to test and to provide good UI/UX. If that is ultimately required, I am not opposed, but would like to avopid them.

> Well it's not really their own. It's a python environment that tweaks enough parts of the global python to hide away globally installed packages. Which also sounds to me exactly what an isolated python environment is.

Let's not get wrapped up on semantics, what I am saying it that it is a different file that is being run. We cannot use `sys.executable`, we need to use the environment Python.

> I've suggested to use venv, that's part of the standard library so should not need boostrapping. I did say virtualenv for python 2, but if you'd not want that we can use some hackish solution like here.

Our current approach should work fine on Python 2 since there is no `pyvenv.cfg` to mess things up. If we do have problems, however, we can add a dependency on virtualenv.

> Why not? Those systems have been designed especially to create isolated build environments, and venv does not require any custom boostraping.

*sigh* because we have to run their Python, so we need to patch `sys.executable`. That is fine short term, but not long term.

> I guess we fundamentally disagree on how one should go about supporting python interpreters. I prefer designed to work upfront, while you're aiming at wait until enough people complain to make your effort worthwhile. From my POV people trying to use this tool, and then finding out it does not work because it does not support their use cases is hurtful to the whole packaging ecosystem, because another thing that they should use but they cannot.

Well, maybe the problem here is why they need to use a tool that is not mature yet. I think the PEP 517 move was not correctly handled, it should have never been pushed into people until there was good enough tooling for the new workflow. I think PEP 517 adoption was premature, and that is the cause of the issue, not that this fairly new tool is not good enough yet.

Furthermore, when I complain about the PEP 517 without proper tooling I was told that was a non-issue and that there were in fact tooling available. People should be using that tooling instead then until this is ready.

So, again, please open a PR to use venv instead. I am putting my efforts into fixing the workflow in Python itself first. I am not arguing against using venv, just that it is not my priority.

If you think that is a deal breaker for PyPA inclusion, fine, I don't really agree but understand. We can postpone that until you are happy with the current state.

> I can understand `virtualenv` being a problem for bootstrapping, but `venv` is in the standard library. The reasons for wanting to bootstrap and devendor everything have to do with making sure you can apply (security, mostly) patches in a central location, which is actually a major plus for using `venv` — CPython is way more likely to get security bug reports and fixes like that than this small project, which would argue in favor of delegating to `venv`.

I wrote that to clarify the scope. I agree that `venv` is not an issue for bootstrapping.

> Anything we fix in Python would only hit Python 3.10+. Assuming `python-build` continues to support anything not EOL upstream (and we're already actually way past EOL on a few Python versions we're supporting), I don't think we'd be able to take advantage of anything of this nature for ~4-5 years, so I think "long term" may be very long indeed.

Yes, my goal right now is that that doesn't change to Python 3.11+. I want to write a PEP proposing a way to fix the issue and try to get the changes in as soon as possible. After that I want to properly fix older versions, but I think the use cases that this PR would leave out are obscure enough that don't justify changing my priorities.

> I think we're already doing _significantly_ better than `pep517`, and that this PR is a net improvement for this package. We definitely must not let the perfect be the enemy of the good here. I wonder if this might be a good compromise:
> 
>     1. For the easiest cases, we maintain our own lightweight isolating virtual environment — for people who care about bootstrapping, it's easy enough to make their use case the easy case, since they control their build environments (and are generally linux distros anyway — plus `build` can be built as a universal wheel, so you can always bootstrap `build` in a specific environment and then use it to build other stuff on your constrained environments).
> 
>     2. For the tougher cases, we fall back to `venv` on Python 3 and either `virtualenv` or just no significant isolation support on Python 2 (we can raise a warning to that effect).
> 
> 
> The upside of this is that it should help us avoid the complicated logic involved in supporting dozens of different platforms while also giving _most users_ the benefits of not using `venv` (we could also support an opt-in option to use `venv` unconditionally — which would certainly make it easier to test if it's hard to generate these obscure platforms). The downside here is that it's at least predicated on our ability to detect that we're in an unusual situation so we know when to fall back — and _detecting_ that may end up being almost as complicated as supporting all the other platforms.
> 
> Another option that is possibly more palatable would be to default to using `venv`, but add an option to use the lighter weight option for people who want to avoid it for whatever reason.

I agree with the venv usage, virtualenv is more problematic for me so I would like to avoid it.

Again, I am not against any of this, it is just not my priority right now.

So, for now, I would merge the PR as is and work on drafting up a PEP. If someone has issues with the current implementation, I would rethink my priorities and consider doing the change to venv/virtualenv earlier.

pganssle:
Ok, I think we're finally mostly all on something close to the same page here. ☺

I think to unblock this PR, I think the most important thing here is adding some tests that would have caught this problem that don't rely on the specifics of the implementation — i.e. tests that will continue passing and acting as regression tests when we switch over to using `venv` as the isolated environment provider in a later PR. [My comments on a possible design for this test here](https://github.com/FFY00/python-build/pull/99#issuecomment-691155938).

I think also at least [this comment](https://github.com/FFY00/python-build/pull/99/files#r487735903), [this comment](https://github.com/FFY00/python-build/pull/99#discussion_r487451830) and [this comment](https://github.com/FFY00/python-build/pull/99#discussion_r487451943) still need to be addressed.

@gaborbernat Do you think we can agree that the standard for this particular PR's scope can/should be "makes `build` better than it currently is and doesn't make it worse" rather than "solves the problem perfectly in all environments" (we can spin off a new issue to switch to `venv` backend and/or fix the support for the platforms that this one doesn't cover)? I'd really prefer to get your eye on this, because I don't have any expertise on the details of virtual environment implementations.

gaborbernat:
> we can spin off a new issue to switch to `venv` backend and/or fix the support for the platforms that this one doesn't cover)?

I don't think you'll be able to fix without adding few hunders of lines of code, and even then you only support what you know. IMHO the only viable option here is to move to venv. Once you move to venv I'm not sure @FFY00 PEP is needed anymore. Also @FFY00 I think you're confusing PEP-517 and PEP-518. Build isolation comes with the later not the earlier. 

What would you propose in that future PEP to solve the problem?

I'm not saying we should not move ahead with this PR, but just poitining it out that this solution only addresses a fire, and more are ahead. If one would solve it with venv then probably you'd avoid those. If you don't want to deal with that now, no problem. But we should at least create an issue to track it.

> I'd really prefer to get your eye on this, because I don't have any expertise on the details of virtual environment implementations.

We discussed this a while ago, but the usage of nox over tox or not using black kinda discourages me to contribute to this project 🤦. I know it's dumb, but considering it's my free time I prefer to spend it on code bases that I find easy to use/read/write. 🤦 🤷

gaborbernat (approved):
Don't consider my comments as deal breakers, but just notting the project as it stands will (very likely) not work (correctly) with Windows Store or macOs framework python implementations.

FFY00:
> I don't think you'll be able to fix without adding few hunders of lines of code, and even then you only support what you know. IMHO the only viable option here is to move to venv. Once you move to venv I'm not sure @FFY00 PEP is needed anymore. Also @FFY00 I think you're confusing PEP-517 and PEP-518. Build isolation comes with the later not the earlier.

I agree with Bernát here, it would be very difficult to distinguish where to use venv and where not. I think we should just use venv.

> What would you propose in that future PEP to solve the problem?

Essentially I think we should have an environment variable to disable `pyvenv.cfg` so that it doesn't clash with `PYTHONHOME`, restoring the ability to rely on `PYTHONHOME`. But I would also like to propose to have the sysconfig variables configurable via environment variables. This would result in a weaker alternative to `pyvenv.cfg`.

The goal is to allow spinning up custom environments completely independently on the interpreter.

sysconfig paths:
```
$ python -m sysconfig
Platform: "linux-x86_64"
Python version: "3.8"
Current installation scheme: "posix_prefix"

Paths:
	data = "/usr"
	include = "/usr/include/python3.8"
	platinclude = "/usr/include/python3.8"
	platlib = "/usr/lib/python3.8/site-packages"
	platstdlib = "/usr/lib/python3.8"
	purelib = "/usr/lib/python3.8/site-packages"
	scripts = "/usr/bin"
	stdlib = "/usr/lib/python3.8"

...
```

With this we should be able to simply start a process from `sys.executable` providing something like `PYTHONIGNOREVENV=1`, `PYTHONPLATLIB=my_env/site-packages`, `PYTHONPURELIB=my_env/site-packages`, `PYTHONDATA=my_env/data` and `PYTHONSCRIPTS=my_env/bin`. `PYTHONIGNOREVENV` may not even be needed here, but I still think it's a good idea.

What do you think?

gaborbernat:
Not sure I understand how would it be useful. How would you provision pip for a such isolation, without making the entire system packages available🤔maybe would help to create a no packages installed venv via just env vars but creating virtual environments without pip is not expensive 🤷‍♂️so IMHO would cost more than benefit for interpreter maintenance.

FFY00:
> Not sure I understand how would it be useful. How would you provision pip for a such isolation, without making the entire system packages available :thinking: maybe would help to create a no packages installed venv via just env vars but creating virtual environments without pip is not expensive :man_shrugging: so IMHO would cost more than benefit for interpreter maintenance.

I don't see any reason why ensurepip would not work since I am not touching `stdlib` and `platstdlib`, just the site packages.

The benefit here would be that we can run a virtual environment with any interpreter, that is the whole point. It would also solve the Windows store Python issue, as we wouldn't need to copy the executable anywhere, no weird workarounds needed.

gaborbernat:
I guess could work... But arguably only useful for tools, as would be impractical to use by end users. I'm a bit skeptical core devs would accept this kind of fine tunability while you can already solve the problem with just invoking virtualenv. The benefits over direct virtualenv invocation overall are very low the way I see it.

FFY00:
Yes, only useful for tools. virtualenv would indeed be available but the problem I see here is that bootstraping virtual environments when we have some limitations regarding the interpreter, like your example of the Windows store, is *very* tricky. virtualenv is great and I am very glad it exists but I would like to see virtual environments being easily bootstrapable by anyone. The current status quo is use virtualenv or have pain.

Being able to create virtual environments without any dependency on what Python interpreter is being run would really help that.

Btw, I am still not clear on how virtualenv solves the issue of using Windows store Python when symlinks are not available. I would really appreciate if you could clarify that for me.

The main issue I see here is how most console entrypoints are generated. For tools like pipx that want to run an entrypoint from one specific environment, they would have to clear the environment before running the Python process. This is prefectly doable, but could be annoying.

I think it is at least worth exploring and getting feedback from other people. Do you agree?

gaborbernat:
> Being able to create virtual environments without any dependency on what Python interpreter is being run would really help that.

For Python 3 you can just use venv. For Python 2 you need virtualenv, but let's be realistic any PEP we'd come up would be python 3 only.

> Btw, I am still not clear on how virtualenv solves the issue of using Windows store Python when symlinks are not available. I would really appreciate if you could clarify that for me.

It doesn't solve it. Delegates the job to the window store pythons venv module, that knows how to handle himself. The venv module basically looks up a redirect script and uses that as python executable.

> I think it is at least worth exploring and getting feedback from other people. Do you agree?

If you have the bandwidth for it sure. From my side not worth the gain considering the effort one needs to put in, but feel free to pursue that path.

layday:
I'm not a packaging guru (by any measure!) but I don't understand what we're trying to do here.  As I understand it, build isolation means that the builder should not have access to system- and user-level packages (anything that is not part of the stdlib).  To my knowledge, you can accomplish this in two ways: minimally, by overriding `PYTHONPATH` in combination with the `-S` flag to disable the loading of the `site` module; or you can construct a virtual environment to take care of all of the minutiae for you.  Following that, you would invoke `pip`, after having installed it at a temporary location using `ensurepip` or `venv` (which wraps `ensurepip`) or the virtualenv package, to install the build requires, at another temporary location, on the `PYTHONPATH`.  Anything else would be outside the scope of build isolation.

FFY00:
We have no control over how the Python subprocess in `pep517` is called so we can't use the `-S` flag. Also, the backend or its dependencies might install needed console scripts or package data so we need to make sure the sysconfig paths are correct, or we might have trouble. So essentially, we need a virtual environment.

layday:
Thanks for explaining.  Using a virtual env sounds like a far better option than trying to account for what looks like a hotchpotch of edge cases in this package.

ofek:
@gaborbernat Are there plans to port your `virtualenv` refactor to `venv`? If not, I would strongly encourage us to not use the latter simply because of the abysmal performance, especially on Windows.

gaborbernat:
> @gaborbernat Are there plans to port your `virtualenv` refactor to `venv`? If not, I would strongly encourage us to not use the latter simply because of the abysmal performance, especially on Windows.

That's for the venv maintainers to decide onto. Note venv uses ensurepip to provision pip/setuptools, so in all likelyhood it's up to the ensurepip maintainers at the end of day. This seems to be @ncoghlan @pfmoore @dstufft. So if they agree that the path taken by virtualenv is preferable, can make a PR against CPython (and PyPy respectively).

pfmoore:
Can someone summarise the issue here, and what it is that is being suggested ensurepip (or venv) needs to do? I don't have time right now to read the whole of this thread, but I'm happy to comment on something specific.

gaborbernat:
off-topic for this discussion but currently ensurepip uses pip to install pip+setuptools. @ofek was suggesting it should do the same as virtualenv does: extract wheel himself (potentially via the install project), and cache it. So that venv virtual environment creation time is on par with virtualenv.

pfmoore:
OK, I'd say that's not likely to be something we'd add to ensurepip. Once installer is mature, *maybe* it would be worth using to do the installs - but why add an extra component to the core when we already have a wheel installer included in the form of pip?

More likely would be improvements to speed up pip. We currently use `--no-cache --no-index --find-links` so we should be grabbing the wheels direct and just unpacking them. Very recent versions of pip unpack wheels direct to the target, rather than unpacking to a temp location and copying into place, so that will help when that version goes into core. Otherwise, someone would need to come up with a specific change (probably around the flags we use to call pip) and demonstrate the improvement, in a PR, I guess. We wouldn't object to changes that improve venv creation time, but someone needs to do the work.

gaborbernat:
> why add an extra component to the core when we already have a wheel installer included in the form of pip?

@ofek answers this.

> simply because of the abysmal performance, especially on Windows.

> We wouldn't object to changes that improve venv creation time, but someone needs to do the work.

Your words above do object against this. And instead redirect people to make pip faster. Note this later is likely by a few order of magnitudes harder task, due to the legacy/complexity of pip compared to a simple wheel unistall/cache system.

pfmoore:
@gaborbernat OK, sorry if I'm missing things - as I say, I don't have time to read the whole thread, so I'm somewhat relying on the summary.

To repeat what I was trying to say - what is the proposed change, and do we have a demonstration that it'll increase venv creation speed sufficiently to justify any additional costs and overheads? I suspect a big proportion of the speed penalty is pip unpacking to a temp location and copying. Has anyone confirmed whether the new version of pip with unpack in place improves things here and by how much? That would be the first thing to check.

The proposal to add the installer module has a number of problems right now. First, the installer module isn't even complete. Let's look again when that's sorted out. Secondly, we would be adding a module which would be *solely* in support of ensurepip, and wouldn't be installed on the user's system. That's a bit of a weird thing to do - why not either bring installer into the stdlib (immediate response - it needs to mature a *lot* before that's worth considering) or use the tools that are already there, i.e., pip (the response being "pip's too slow", to which my counter is "why not improve pip and help everyone?") 

To directly answer your question:

> So if they agree that the path taken by virtualenv is preferable, can make a PR against CPython (and PyPy respectively).

I'm happy to see a PR, as long as it comes with benchmarks that demonstrate the improvement, and the code isn't more of a maintenance burden than the improvement warrants. One particular question I'd want to see addressed, is who would update the code to deal with possible future changes in the wheel format? Would that task fall on the core devs?

gaborbernat:
@pfmoore I'll defer for the foreseable future doing all that research you're requsting. From POV of this project @ofek our only choice is to go with venv, be that as fast or slow as it is. Mainly because we don't want this project to take dependencies for boostraping reasons.

gaborbernat:
@FFY00 shall we close this in favor of #112 ?

PR #112, opened 2020-10-04
env: use venv on Python 3

FFY00:

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -164,7 +164,11 @@ class IsolatedEnvironment(object):
         exe = 'python'
         if os.name == 'nt':
             pythonw = '{}w.exe'.format(exe)
-            if os.path.isfile(os.path.join(self.path, env_scripts, pythonw)):
+            if (
+                os.path.isfile(os.path.join(self.path, env_scripts, pythonw)) and not
+                # pythonw fails on Python 3.5 Windows
+                (os.name == 'nt' and sys.version_info[:2] == (3, 5))
+            ):
                 exe = pythonw
             else:
                 exe = '{}.exe'.format(exe)
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -6,7 +6,7 @@ import sysconfig
 import tempfile
 import types
 
-from typing import Dict, Iterable, Optional, Sequence, Type
+from typing import Dict, Iterable, List, Optional, Sequence, Type
 
 
 if sys.version_info[0] == 2:  # pragma: no cover
@@ -55,8 +55,12 @@ class IsolatedEnvironment(object):
         self._env = {}  # type: Dict[str, Optional[str]]
         self._env_vars = {}  # type: Dict[str, str]
         self._path = None  # type: Optional[str]
-        self._remove_paths = remove_paths
+        self._remove_paths = []  # type: List[str]
         self._executable = _executable
+
+        # normalize paths so that we can compare them -- required on case insensitive systems
+        for path in remove_paths:
+            self._remove_paths.append(os.path.normcase(path))
 
     @property
     def path(self):  # type: () -> str
@@ -129,7 +133,7 @@ class IsolatedEnvironment(object):
         for path in self.MANIPULATE_PATHS:
             env_path = self._get_env_path(path)
             if env_path:
-                sys_path.append(env_path)
+                sys_path.append(os.path.normcase(env_path))
 
         for path in self._remove_paths:
             if path in sys_path:
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -11,6 +11,8 @@ from typing import Dict, Iterable, Optional, Sequence, Type
 
 if sys.version_info[0] == 2:  # pragma: no cover
     FileExistsError = OSError
+else:
+    import venv
 
 
 _HAS_SYMLINK = None  # type: Optional[bool]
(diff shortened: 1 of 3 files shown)


FFY00:
As discussed in #99, we keep the current scheme for Python 2, which works fine because there is not `pyvenv.cfg` to mess things up, and use `venv` on Python 3.

FFY00:
@zooba dod you know what is going wrong on Windows?

src/build/env.py
```diff
@@ -117,38 +123,44 @@ def _symlink_relative(self, path):  # type: (Optional[str]) -> None
...
+            self._replace_env('PATH', os.pathsep.join(exe_path))
+            self._replace_env('PYTHONPATH', os.pathsep.join(sys_path))
+            self._replace_env('PYTHONHOME', self.path)
+
+            self._symlink_relative(sysconfig.get_path('include'))
+            self._symlink_relative(sysconfig.get_path('platinclude'))
+            self._symlink_relative(sysconfig.get_config_var('LIBPL'))
+        else:
+            subprocess.check_call([self.executable, '-m', 'venv', self.path])
+            self._executable = os.path.join(self.path, 'bin', 'python')
```
gaborbernat, line 161: This is incorrect on Windows. You should specify the ``.exe`` suffix. Also this is ``Scripts`` on Windows. You should call:

```
env_scripts = self._get_env_path('scripts')
```
To get the platform specific name.

src/build/env.py
```diff
@@ -117,38 +123,44 @@ def _symlink_relative(self, path):  # type: (Optional[str]) -> None
...
-        self._symlink_relative(sysconfig.get_config_var('LIBPL'))
+            self._replace_env('PATH', os.pathsep.join(exe_path))
+            self._replace_env('PYTHONPATH', os.pathsep.join(sys_path))
+            self._replace_env('PYTHONHOME', self.path)
+
+            self._symlink_relative(sysconfig.get_path('include'))
+            self._symlink_relative(sysconfig.get_path('platinclude'))
+            self._symlink_relative(sysconfig.get_config_var('LIBPL'))
+        else:
+            subprocess.check_call([self.executable, '-m', 'venv', self.path])
```
gaborbernat, line 160: There's no reason to make this a subprocess call. Those are expensive on Windows especially. Use the venv api instead, 

https://docs.python.org/3/library/venv.html#venv.EnvBuilder

src/build/env.py
```diff
@@ -166,13 +178,13 @@ def install(self, requirements):  # type: (Iterable[str]) -> None
         if not requirements:
             return
 
-        subprocess.check_call([sys.executable, '-m', 'ensurepip'], cwd=self.path)
+        subprocess.check_call([self.executable, '-m', 'ensurepip'], cwd=self.path)
```
gaborbernat, line 181: You only need to call this on Python 2. venv always comes with pip.

gaborbernat (changes requested):

src/build/env.py
FFY00: Right :facepalm:

tests/conftest.py
```diff
@@ -88,7 +68,14 @@ def integration_path():
         if build.env.fs_supports_symlink():
             os.symlink(self_source, self_dest)
         else:  # pragma: no cover
-            _copy_dir(self_source, self_dest, ignore=['.git', '.nox', '.integration-sources'])
+            os.makedirs(self_dest)
+            for target in ('pyproject.toml', 'setup.cfg', 'LICENSE', 'src'):
```
gaborbernat, line 72: Some documentation here would be helpful, stating why we need to single out these files, and not others.

tests/test_env.py
```diff
@@ -10,6 +10,7 @@
 import build.env
 
 
+@pytest.mark.skipif(sys.version_info[0] != 2, reason='Custom iolsated environment only available on Python 2')
```
gaborbernat, line 13: typo - iolsated -> isolated

gaborbernat (changes requested):

tests/conftest.py
FFY00: It was just Windows was having trouble copying the whole directory, so I am now copying the required files. For context, since it can be hard to track, this is the code to create the project folder for python-build in the integration tests. For the other projects we just download the tarball, but for python-build we copy the current source.

gaborbernat (approved):

FFY00:
Well, it looks like venv on Windows PyPy 3 is completely broken, from log it seems like the environment can't find the standard library. Not sure what is wrong on Windows CPython 2. Windows CPython 3.5 could be a fluke.

It would be really helpful if someone with Windows could have a look. Pinging @zooba and @pfmoore, if any of you have a few minutes to test and debug this, it would be greatly appreciated :blush:

zooba:
Does PyPy still lay out their library in multiple directories? If it's in any sort of sys.path isolation then it may require extra logic to set it up.

FFY00:
I am not sure. But shouldn't PyPy's `venv` just work?

FFY00:
So, looks like using `pythonw` fails on CPython 3.5 Windows.

gaborbernat:
> So, looks like using `pythonw` fails on CPython 3.5 Windows.

Weird, some bug? What about Windows 2.7?

FFY00:
I don't know. Just figured out Python 2, the isolation is broken because the paths never get removed due to case insensitive paths.

FFY00:
This was as far as I got today. PyPy in the CI seems to be borked (possibly has an older version?) but it still doesn't work locallly.

python-build-pypy-error

FFY00:
Okay, seems that ensurepip is what is borked. We can proceed and tell PyPy Windows users to bring their own pip.

gaborbernat:
> We can proceed and tell PyPy Windows users to bring their own pip.

How was it borked? How can they bring their own pip? Did you open an issue for the PyPy project about this?

PR #117, opened 2020-10-07
setup.cfg: rename entrpoint script to pyproject-build

### LINKED ISSUE #101: Clashes with pyenv
pyenv brings with it the [python-build](https://github.com/pyenv/pyenv/tree/master/plugins/python-build) tool, which it uses to build python versions from source.

Now I've done `pip install build`, pyenv invokes *this* python-build, and breaks 😢 

Maybe `python-build` could ship only for invoking as `python -m build`? This would have the advantage of always knowing which python binary you were invoking (the same reason [pip recommends using `python -m pip` everywhere](https://adamj.eu/tech/2020/02/25/use-python-m-pip-everywhere/)).

Or maybe pyenv could be convinced to change to `pyenv-python-build` or similar, though probably breaking some peoples' install scripts?

FFY00:
Fixes #101

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -14,7 +14,7 @@ It is a simple build tool and does not perform any dependency management.
 
 .. note::
 
-   A ``python-build`` CLI script is also available, so that tools such as pipx_
+   A ``pyproject-build`` CLI script is also available, so that tools such as pipx_
    can use it.
 
 By default python-build will build the package in a isolated
--- a/setup.cfg
+++ b/setup.cfg
@@ -35,7 +35,7 @@ package_dir =
 
 [options.entry_points]
 console_scripts =
-    python-build = build.__main__:entrypoint
+    pyproject-build = build.__main__:entrypoint
 
 [options.extras_require]
 doc =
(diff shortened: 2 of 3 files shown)


gaborbernat (approved):

gaborbernat (changes requested):
Let's go with ``pypa-build`` name 👍

gaborbernat (approved):

gaborbernat:
@FFY00 You have a merge conflict here you'll need to resolve.

pradyunsg:
Is there any interest in renaming this project to match the entry point name? It'll work better for discoverability as well as for SEO, since the search term "python build" will probably not show this project, like, ever.

gaborbernat:
Please read https://github.com/pypa/build/issues/101, there has been a long discussion there where we concluded that we're most comfortable with the current state of the world. Even if the project name and the entry point name does not match.

pradyunsg:
I have but you locked that topic, so I didn't want to comment there?

Anyway, at no point did anyone address the concerns around discoverability and, this was raised in the thread multiple times. If you want to lock threads to shut down someone raising a concern that they don't feel has not been addressed, so be it. I don't think it'd be worth my time to say anything else here. 🤷🏽‍♂️

gaborbernat:
Sorry I should have been clearer. This is one of those topics we can bikeshed endlessly. Both me and @FFY00 noted the issue with discoverability, but both feel strongly about keeping the ``-m`` name of ``build`` and ``pyproject-build`` as console script name. We noted this hinders discoverability but considered we can live with it.

PR #119, opened 2020-10-07
Use tox instead of nox

gaborbernat:
Example https://github.com/gaborbernat/python-build/actions/runs/293226462


gaborbernat:
pypy2 isolation is broken, can reproduce it locally. Not sure how/why it passes with nox, but it's a bug.

.github/workflows/check.yml
```diff
@@ -0,0 +1,130 @@
...
+          - windows
+        py:
+          - 3.8
+          - 3.7
+          - 3.6
+          - 3.5
+          - 2.7
+          - pypy3
+          - pypy2
+    steps:
```
FFY00, line 31: Empty line above.

.github/workflows/check.yml
```diff
@@ -0,0 +1,130 @@
...
+      - name: setup python for tox
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
+      - name: install tox
```
FFY00, line 36: Empty line above.

.github/workflows/check.yml
```diff
@@ -0,0 +1,130 @@
...
+        run: python -m pip install tox
+      - uses: actions/checkout@v2
```
FFY00, line 38: Add `name`.
Empty line above.

.github/workflows/check.yml
```diff
@@ -0,0 +1,130 @@
...
+      - name: setup python for test ${{ matrix.py }}
```
FFY00, line 39: Empty line above.

And so on...

gaborbernat:
pypy does not handles/honor/cares about PYTHONHOME, see https://foss.heptapod.net/pypy/pypy/-/blob/branch/default/pypy/interpreter/app_main.py#L38, hence the issues here; prefixes remain under original executable path, which then pulls in the site-packages via https://foss.heptapod.net/pypy/pypy/-/blob/branch/default/lib-python/2.7/site.py#L295-297 and https://foss.heptapod.net/pypy/pypy/-/blob/branch/default/lib-python/2.7/distutils/sysconfig_pypy.py#L38-56 respectively

FFY00 (changes requested):
Overall looks good, thanks :+1:
Not a big fan of the kind of handling we need to do in the CI to calculate the environment name, but I guess that's required.

Can you fix the style to match the rest/old. Please add empty lines between steps, always have a human-readable name for steps and please capitalize the first letter of the step name (to match the style of the mandatory steps).

Also, please drop the PyPI publish action, I want to start signing the packages there, not just the git tag. We will have a release script for this anyway. It also decreases attack surface because we will not need to have a token with PyPI upload permissions dangling on github.

gaborbernat:
> Also, please drop the PyPI publish action, I want to start signing the packages there, not just the git tag. We will have a release script for this anyway. It also decreases attack surface because we will not need to have a token with PyPI upload permissions dangling on github.

I personally consider the Github security teams solution more secure than our own machines. Because the github release machine has more security experts reviewing their solution, and we don't have to store security keys on our own machine. 🤔 Furthermore, I'd rather have release automated via CI, than end-user invoked, if possible.

FFY00:
> I personally consider the Github security teams solution more secure than our own machines. Because the github release machine has more security experts reviewing their solution, and we don't have to store security keys on our own machine. thinking Furthermore, I'd rather have release automated via CI, than end-user invoked, if possible.

How so? Your machine has push permissions to Github, if it compromised it can create releases that look valid in the Github side and there is no security precautions Github can take to prevent that. And I do trust my machine more than Github.

Anyway, as I said, I want to sign the PyPI packages, so that requires me to do it locally anyway. I am sorry but I will have to put the foot down in this one :confused: I am happy to keep discussing if you want.

gaborbernat:
> How so? Your machine has push permissions to Github, if is compromised it can create releases that look valid in the Github side and there is no security precautions Github can take to prevent that. And I do trust my machine more than Github.

I don't store security keys on my machine, if I can avoid. I create PyPi API keys that can publish to just 1 project, and store that under Github security tokens. Can't you store the security key for signing the package also in Github secrets store, if we want to sign it?

FFY00:
> I don't store security keys on my machine, if I can avoid. I create PyPi API keys that can publish to just 1 project, and store that under Github security tokens. Can't you store the security key for signing the package also in Github secrets store, if we want to sign it?

Right, but don't you have a ssh key with push permissions to Github? With that an attacker could push malicious code and make a release, Github would then upload it to PyPI. Even if you have your ssh key encrypted or even if you use plain credentials, if an attacker has access to your machine, they could easily install a keylogger to get your password.

gaborbernat:
> Right, but don't you have a ssh key with push permissions to Github?

No SSH, just HTTPS. However, my password is storred within macOs Keychain Access, that's password protected and fairly well securred.

FFY00:
For clarification, I have my ssh key as a gpg authenticaion key inside my yubikey. It's kind of a complicated setup, you can read about it here https://www.linode.com/docs/security/authentication/gpg-key-for-ssh-authentication/.

pganssle:
You want a local release script? That will be annoying, and if I understand correctly it means that either you'd be the only one authorized to do releases or we'd need to have a proliferation of people who have release privileges.

It's also a lot more of a PITA.

I think the only difference in threat models between "push signed tags to GH, verified on GH Actions" and "make signed releases from your local computer" is that a malicious actor with push permissions on GH could modify the action to not check the signature before publishing to PyPI, which would at least leave a public artifact. Otherwise you still need to be able to sign the git repository with the releaser's key.

Another option might be to have the GHA dump artifacts into a holding area and have a bot running on Heroku, GCP, AWS, a VPS or something check that they are properly signed, with the bot holding the token and doing the upload. I've got a project that lets you execute arbitrary bash scripts in response to Github webhooks: https://gitlab.com/pganssle/githooks

I use it to run my website and it's been working pretty well so far. Might be easy enough to use something like that to do the "release with intermediate verification" automation.

FFY00:
> You want a local release script? That will be annoying, and if I understand correctly it means that either you'd be the only one authorized to do releases or we'd need to have a proliferation of people who have release privileges.

To publish release you should meet the following criteria:
1) Have push permission to the repo
2) Have push permission on PyPI
3) Have your GPG key listed in the documentation
	3.1) Preferably have your GPG key signed by the original/main maintainer, in this case, me (I think we should probably make this a requirement, but I don't want to promote bad key signing practices)

Any maintainer, after given release permission, should be able to make a release.

The script could check for the requirements, at least the git and PyPI push permissions, before starting the process.

> It's also a lot more of a PITA.

Which part exactly would be PITA? The only thing I'd imagine would be a bit of PITA is the PyPI upload, where you need to either provide the credentials or have a token. But I think that is a perfectly reasonable for a rrelease workflow. You won't be releasing 

> I think the only difference in threat models between "push signed tags to GH, verified on GH Actions" and "make signed releases from your local computer" is that a malicious actor with push permissions on GH could modify the action to not check the signature before publishing to PyPI, which would at least leave a public artifact. Otherwise you still need to be able to sign the git repository with the releaser's key.

A token with push permission to PyPI would be saved as a secret on Github. There can't be such thing as Github Actions verifying your git tag, if I have push access to Github I can disable the check and just use the token to upload anything I want. The verification can't happen somewhere that could be compromized.

Furthermore, gaining push access to Github is an extremely low bar. If I can access your computer and you have a Github session open, I have push access. If I acess your computer and you have an unlocked SSH key (I think most people use a keyring, and the key is generally unlocked on a timeout or even for each boot cycle).

> Another option might be to have the GHA dump artifacts into a holding area and have a bot running on Heroku, GCP, AWS, a VPS or something check that they are properly signed, with the bot holding the token and doing the upload. I've got a project that lets you execute arbitrary bash scripts in response to Github webhooks: [gitlab.com/pganssle/githooks](https://gitlab.com/pganssle/githooks)

Basically the same attack surface as Github, could be a bit better depending on your provide security practices. Having a session open on the provide website (if they don't require reauthentication for sensitive operations) or an unlocked ssh key could compromise.

This also makes the setup more complex and can introduce financial cost.

> I use it to run my website and it's been working pretty well so far. Might be easy enough to use something like that to do the "release with intermediate verification" automation.

That is fine for a personal website. This project had the possibility of compromising the supply chain.

Even ignoring all that, I want to sign the artifacts that get pushed to PyPI, some distros prefeer to download the sdist from there instead of fetching the source from Github. I don't see how to do this remotely.

pganssle:
> Which part exactly would be PITA? The only thing I'd imagine would be a bit of PITA is the PyPI upload, where you need to either provide the credentials or have a token. But I think that is a perfectly reasonable for a release workflow. You won't be releasing very often.

I've been making releases manually on `dateutil` for years (and they are very infrequent), and I consider it a bit of a pain, because:

1. It happens in my local directory, which may have all kinds of weird build artifacts lying around.
2. There's no guarantee that the thing I'm uploading is the thing that just passed the tests on all platforms.
3. There's a lot of latency involved in pushing, waiting for tests, etc.
4. It's way harder to scale out and keep track of the people authorized to do releases.

By contrast, I can make sure that the builds all happen via a completely automated and reproducible process if I have CI do them, and it's very easy to scale out who has access. I just have to tag, wait a while, then check to make sure everything worked.

> Basically the same attack surface as Github, could be a bit better depending on your provide security practices. Having a session open on the provide website (if they don't require reauthentication for sensitive operations) or an unlocked ssh key could compromise.

This is not the same attack surface as Github really, because very few people need admin access to the uploader bot, so the permissions to trigger a release are separated from the permissions to alter the checks required for a release, and most people don't need permissions to alter the checks.

So the workflow would be something like GHA triggers a release and a build then makes the artifacts available via the little deployment zip file they have. This triggers a webhook that causes ReleaserBot to:

1. Download the zip file containing the built artifacts.
2. Check that they were built against a commit signed by someone on the allowed releaser list.
3. Sign the artifacts with a dedicated key or subkey signed by the maintainers.
4. Use a token stored locally to upload to PyPI.
5. Publish logs somewhere public.

Now the threat models are:

1. Someone compromises the releaser bot.
2. Someone can MITM the releaser bot and pretend to be Github.
3. Someone compromises the GPG key and Github account of anyone authorized to make releases.

Number 2 is not a realitsic concern since it would use HTTPS. Number 3 is already the threat model for "local script to make releases", so we're no worse off in that situation. The most likely avenue for compromising the releaser bot would be getting access to the credentials of someone who can access it — which is almost certainly a subset of the set of people with permissions to upload, in which case we're again no worse off than we would be in a "local release script" situation. We'd actually be slightly _better_ off in some respects, because the entire setup I'm describing is much more auditable than a local release script would be. If we set it up right, every event is automated and logged in public.

FFY00:
> I've been making releases manually on `dateutil` for years (and they are very infrequent), and I consider it a bit of a pain, because:
> 
>     1. It happens in my local directory, which may have all kinds of weird build artifacts lying around.
> 
>     2. There's no guarantee that the thing I'm uploading is the thing that just passed the tests on all platforms.
> 
>     3. There's a lot of latency involved in pushing, waiting for tests, etc.
> 
>     4. It's way harder to scale out and keep track of the people authorized to do releases.
> 
> 
> By contrast, I can make sure that the builds all happen via a completely automated and reproducible process if I have CI do them, and it's very easy to scale out who has access. I just have to tag, wait a while, then check to make sure everything worked.

All this can be fixed in a release script.

> Now the threat models are:
> 
>     1. Someone compromises the releaser bot.
> 
>     2. Someone can MITM the releaser bot and pretend to be Github.
> 
>     3. Someone compromises the GPG key and Github account of anyone authorized to make releases.
> 
> 
> Number 2 is not a realitsic concern since it would use HTTPS. Number 3 is already the threat model for "local script to make releases", so we're no worse off in that situation. The most likely avenue for compromising the releaser bot would be getting access to the credentials of someone who can access it — which is almost certainly a subset of the set of people with permissions to upload, in which case we're again no worse off than we would be in a "local release script" situation. We'd actually be slightly _better_ off in some respects, because the entire setup I'm describing is much more auditable than a local release script would be. If we set it up right, every event is automated and logged in public.

I am not comfortable with 1), which can be completely avoided with the release script. If I worked at the provider it could be fairly easy to simply fetch the PyPI token from the machine. Number 2) I agree with you. For number 3), what you said isn't necessarily true. If you simply have the keys on disk, yes, but you are not required to have the keys on the disk. For eg. in my setup, as I mentioned above, I use a yubikey, so my key is not on disk and it would be very difficult to steal the key, it would require a very advanced physical attack. For every operation, I am required to unlock the key with a pin and then physically touch it to sign-off, the key is never tranfered to the host computer, cryptographic operations happen on the key itself.

Having the release script locally simply has a smaller attack surface. And I will be making most, if not all, releases anyway, we can come back to this if that changes in the future. I think that is reasonable, no?

pganssle:
> All this can be fixed in a release script.

I mean, almost none of it can be. I have release scripts. You could theoretically spin up a docker container every time (now your release script has a dependency on docker, logs are local only if you bother to save them). I guess you could have your release script poll CI and then block until it succeeds, but then of course you have to leave the computer doing the release open and running during the entire CI run. A release script does not help with permissions management (revoking permissions in one place rather than revoking them in several places).

In any case, the threat model for 1 that you propose is a little far-fetched if we were to host on a reputable hosting platform. I would think we would do this in one of only a few ways:

1. Ask the PSF to host it on their infrastructure ­— in which case compromise of our infra is roughly equivalent to compromise of PyPI anyway.
2. Host it on AWS / GCP / Azure — in which case it is astoundingly unlikely that some random employee of the huge cloud companies who are fiercely competing with each other would be able to just compromise a key.
3. One of us runs a local internet-connected server on a raspberry pi or something.

I think you're probably over-thinking things and doing a bit of security theater. Yes there's some possibility that if you hand this over to a bot it could get compromised. On the other hand, anything I've proposed here is head and shoulders more secure than many _extremely popular Python libraries_. If someone compromises my computer and GPG keys, they could make arbitrary releases of `setuptools` for example. If they compromise Bernat's computer they could presumably make releases of `tox` and `virtualenv`. I doubt they would spend much effort on it, and it would be reasonably simple to add in an automatic audit — configure a server to watch PyPI and send you an e-mail any time a new release is made.

It doesn't affect me much anyway. If there is a way to cut a release by sending you my public key and then pushing signed tags, there's a decent chance that I would cut a release from time to time. If it's a manual process using a local script, I won't.

FFY00:
> I mean, almost none of it can be.

I agree it is not a perfect solution, but it fair IMO.

> * It happens in my local directory, which may have all kinds of weird build artifacts lying around.

The release script creates a temporary directory and git clones from out local .git to there, making sure it is a clean copy of the repo and everything is clean.

> * There's no guarantee that the thing I'm uploading is the thing that just passed the tests on all platforms.

The release script then query the CI for the check status. (more info in the next point)

> * There's a lot of latency involved in pushing, waiting for tests, etc.

You are tagging a release from a particular commit. There would be a requirement for this commit to have already been pushed and the Github CI finished. When tagging a release, we would query Github for the CI status and proceed if it succeeded.

> * It's way harder to scale out and keep track of the people authorized to do releases.

I don't agree much here, ultimately the people that have permission to push a release would be the people with access to the project on PyPI.

> Host it on AWS / GCP / Azure — in which case it is astoundingly unlikely that some random employee of the huge cloud companies who are fiercely competing with each other would be able to just compromise a key.

I wouldn't be *so* sure, but yeah.

> One of us runs a local internet-connected server on a raspberry pi or something.

I manage a server at my college, I could possibly host it there, although I don't know about the long-term reliability. I am gonna stay until next year to finish my bachelor's, not sure if I will stay for my master's, a lot of classes feel like a joke, we'll see.

> I think you're probably over-thinking things and doing a bit of security theater.

Isn't security always a theater :stuck_out_tongue:? Everything is fine until it isn't, and when it isn't, it can have catastrophic consequences.

Every attack we are discussing here is unlikely, but unlikely still means that it can happen. An attack like this can totally happen, and if it does, it had the possibility to hurt people. So, for me it is totally worth, the release process being slightly more annoying is totally worth possibly, even if unlikely, prevent people from getting hurt. But I do understand and accept if you do not feel the same.

> On the other hand, anything I've proposed here is head and shoulders more secure than what many _extremely popular Python libraries_ use. If someone compromises my computer and GPG keys, they could make arbitrary releases of `setuptools` for example. If they compromise Bernat's computer they could presumably make releases of `tox` and `virtualenv`. I doubt they would spend much effort on it, and it would be reasonably simple to add in an automatic audit — configure a server to watch PyPI and send you an e-mail any time a new release is made.

And I am unhappy about that, but there is nothing I can do besides annoying the maintainers, and I don't think they would appreciate that :confused:
I guess that will probably change when something bad happens.

> It doesn't affect me much anyway. If there is a way to cut a release by sending you my public key and then pushing signed tags, there's a decent chance that I would cut a release from time to time. If it's a manual process using a local script, I won't.

The point of the script is so that the process isn't manual...

FFY00:
But anyway, I think we can agree that for now there shouldn't be a release action :grin:

@pganssle if you want to keep discussing I am happy to, but please do not do it if I am annoying you. I know I am very opinionated about this, but I don't want to be annoying anyone! We can figure out a solution later.

gaborbernat:
@FFY00 this is also ready for merge, though you need to migrate the required CI checks.

FFY00 (approved):
LGTM

Can you just update the CI badges in the README?

Also, could you keep the default CI names? I prefeer them as default, automatically generated with the matrix configuration.

```
name: test ${{ matrix.py }} - ${{ matrix.os }}
```

gaborbernat:
> Can you just update the CI badges in the README?
> 
> Also, could you keep the default CI names? I prefeer them as default, automatically generated with the matrix configuration.

Done.

gaborbernat:
@FFY00 I don't have rights to change the required checks, so in order to merge it you'll need to do it 👍

.github/workflows/check.yml
```diff
@@ -0,0 +1,119 @@
...
+          PYTHON: ${{ matrix.python }}
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          file: ./.tox/coverage.xml
+          flags: tests
+          env_vars: PYTHON
+          name: ${{ matrix.py }} - ${{ matrix.os }}
+
+  check:
+    name: check ${{ matrix.tox_env }} - ${{ matrix.os }}
```
FFY00, line 93: Here too if you can.

tox.ini
```diff
@@ -0,0 +1,49 @@
...
+      --cov-report=xml:{toxworkdir}/coverage.{envname}.xml \
+      tests {posargs}
+
+[testenv:type]
+description = run type check on code base
+deps =
+    mypy==0.782
+commands =
+    mypy --strict --python-version 3.8 src/build
+    mypy --strict --python-version 2.7 src/build
```
FFY00, line 37: Also, just notice now, `--strict` could be removed as we already set that in the config.

FFY00 (approved):
Did you remove the pre-commit check? Can you please add it back?

tox.ini
gaborbernat: Oh you're right fixed.

.github/workflows/check.yml
gaborbernat: Oh you're right fixed.

gaborbernat:
@FFY00 ok, now it's actually all good.

FFY00:
Thanks!

PR #121, opened 2020-10-07
pre-commit: add double-quote-string-fixer

FFY00:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,6 +12,7 @@ repos:
   - id: debug-statements
   - id: end-of-file-fixer
   - id: trailing-whitespace
+  - id: double-quote-string-fixer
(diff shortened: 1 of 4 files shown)


gaborbernat (commented):
Before we merge this, let's settle the question of to black or not to black. Because if to black, then this commit is not needed...

FFY00:
If we are ever to move to black, I'd like it to be with `-S` and keep the single quotes.

gaborbernat:
> If we are ever to move to black, I'd like it to be with `-S` and keep the single quotes.

Ever the rebellious, I see. So the ever question is now? Yay or nay? From my side, and Paul's side +1 AFAIK.

gaborbernat (approved):
maser is broken because of https://foss.heptapod.net/pypy/pypy/-/issues/3321, we should probably work around the issue or disable the test for it...

PR #122, opened 2020-10-07
build: adopt black codestyle

FFY00:
On top of #121

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,6 +13,10 @@ repos:
   - id: end-of-file-fixer
   - id: trailing-whitespace
   - id: double-quote-string-fixer
+- repo: https://github.com/psf/black
+  rev: 20.8b1
+  hooks:
+  - id: black
 - repo: https://github.com/pre-commit/mirrors-isort
   rev: v4.3.21
   hooks:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,8 @@
 [build-system]
 requires = ['setuptools >= 40.8.0', 'wheel']
 build-backend = 'setuptools.build_meta'
+
+[tool.black]
+line-length = 127
+skip-string-normalization = true
+target-version = ['py38', 'py37', 'py36', 'py35', 'py27']
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -21,15 +21,12 @@ _WHEEL = re.compile('.*.whl')
         'pip',
         'dateutil',
         'Solaar',
-        os.path.join('flit', 'flit_core')
-    ]
+        os.path.join('flit', 'flit_core'),
+    ],
 )
 @pytest.mark.parametrize(
     ('args'),
-    [
-        [],
-        ['-x', '--no-isolation'],
-    ]
+    [[], ['-x', '--no-isolation']],
 )
 @pytest.mark.parametrize(
     ('call'),
@@ -37,7 +34,7 @@ _WHEEL = re.compile('.*.whl')
         None,  # via code
         ['python', '-m', 'build'],  # module
         ['python-build'],  # entrypoint
-    ]
+    ],
 )
 def test_build(tmp_dir, monkeypatch, integration_path, project, args, call):
     monkeypatch.setenv('SETUPTOOLS_SCM_PRETEND_VERSION', 'dummy')  # for the projects that use setuptools_scm
@@ -61,6 +58,7 @@ def test_isolation(tmp_dir, test_flit_path, mocker):
     try:
         # if flit is available, we can't properly test the isolation - skip the test in those cases
         import flit_core  # noqa: F401
+
         pytest.xfail('flit_core is available')  # pragma: no cover
     except:  # noqa: E722
         pass
--- a/tests/test_projectbuilder.py
+++ b/tests/test_projectbuilder.py
@@ -15,9 +15,11 @@ import build
 if sys.version_info >= (3, 8):  # pragma: no cover
     from importlib import metadata as importlib_metadata
     import email
+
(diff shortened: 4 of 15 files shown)


.pre-commit-config.yaml
```diff
@@ -12,6 +12,12 @@ repos:
   - id: debug-statements
   - id: end-of-file-fixer
   - id: trailing-whitespace
+  - id: double-quote-string-fixer
+- repo: https://github.com/psf/black
+  rev: 20.8b1
+  hooks:
+  - id: black
+    args: ["-S", "-l", "127"]
```
gaborbernat, line 20: Please add these arguments into the pyproject.toml config, so tools can pick it up autoamtically. Any reason you went with 127 instead of 120?

gaborbernat (changes requested):

.pre-commit-config.yaml
FFY00: 127 is the size of the Github editor.

gaborbernat: fair

pyproject.toml
```diff
@@ -1,3 +1,8 @@
 [build-system]
 requires = ['setuptools >= 40.8.0', 'wheel']
 build-backend = 'setuptools.build_meta'
+
+[tool.black]
+line-length = 127
+skip-string-normalization = true
+target-version = ['py27', 'py35', 'py36', 'py37', 'py38']
```
gaborbernat, line 8: Can reverse the order, I prefer to start with latest and go to smallest. Also, what do you think about making it merge friendly, and explode the list to be one version per line?

FFY00: I prefer it the other way :upside_down_face: but sure, doesn't bother me *much*.

tests/test_projectbuilder.py
```diff
@@ -28,19 +30,13 @@
...
 
 
-DEFAULT_BACKEND = {
-    'build-backend': 'setuptools.build_meta:__legacy__',
-    'requires': [
-        'setuptools >= 40.8.0',
-        'wheel'
-    ]
-}
+DEFAULT_BACKEND = {'build-backend': 'setuptools.build_meta:__legacy__', 'requires': ['setuptools >= 40.8.0', 'wheel']}
```
gaborbernat, line 39: This is the one place where having it in separate lines is better, just add the trailing within the array and dictionary ``,`` and black will format it on separate lines.

tests/test_integration.py
```diff
@@ -14,30 +14,14 @@
...
-    ('project'),
-    [
-        'python-build',
-        'pip',
-        'dateutil',
-        'Solaar',
-        os.path.join('flit', 'flit_core')
-    ]
-)
+@pytest.mark.parametrize(('project'), ['python-build', 'pip', 'dateutil', 'Solaar', os.path.join('flit', 'flit_core')])
```
gaborbernat, line 17: here to append the trailing comma to keep it on separate lines

tests/test_integration.py
```diff
@@ -14,30 +14,14 @@
...
+    [[], ['-x', '--no-isolation']],
 )
 @pytest.mark.parametrize(
     ('call'),
-    [
-        None,  # via code
-        ['python', '-m', 'build'],  # module
-        ['python-build'],  # entrypoint
-    ]
+    [None, ['python', '-m', 'build'], ['python-build']],  # via code  # module  # entrypoint
```
gaborbernat, line 24: here to append the trailing comma to keep it on separate lines

gaborbernat (approved):
Otherwise looks 👍

tests/test_projectbuilder.py
FFY00: I agree, and I dislike black doing this. Is there a workaround so that black doesn't do this?

FFY00: I had the `,` before but didn't work. Let me try again.

FFY00: Ah, it does work, sorry!

pyproject.toml
gaborbernat: I prefer to focus on the new fancy versions, and prefer to forget the old ones at the end of the line 😆

tests/test_projectbuilder.py
gaborbernat, line 39: Forgot this line?

tests/test_integration.py
```diff
@@ -21,23 +21,20 @@
...
     ('call'),
     [
-        None,  # via code
-        ['python', '-m', 'build'],  # module
-        ['python-build'],  # entrypoint
-    ]
+        None,
+        ['python', '-m', 'build'],
+        ['python-build'],
+    ],  # via code  # module  # entrypoint
```
gaborbernat, line 37: You probably want to re-explode these comments 🤔

FFY00:
Thanks for letting me know about the trailing `,` trick, it makes black much more usable for me! I went ahead and applied in some other places I though it made sense.

gaborbernat:
> Thanks for letting me know about the trailing `,` trick,

Yeah, adding them also means that when you add a new entry you only need to change the newline, so one-line diff, instead of the usual 2 (add coma, plus new line content).

gaborbernat (approved):

pganssle:
FWIW I support anything that makes it so that I can type my code in any syntactically valid way and have it automatically corrected to be in the project style. `black` or some variation thereof is perfectly acceptable to me.

I personally don't love the 127 character line length, because I generally code with 2 editors / terminal windows side-by-side, which is only ~100 characters with a reasonable font size, and it makes super wide lines annoyingly long. I'd rather either leave the default value or use 80.

That said, it doesn't really matter that much to me.

gaborbernat (approved):

Issue #123, opened 2020-10-07
Test suite breaks on pypy3.6 - 7.3.2 


gaborbernat:
It's because of https://foss.heptapod.net/pypy/pypy/-/issues/3321

FFY00:
We can temporarily remove it then.

gaborbernat:
Maybe you meant disable it?

FFY00:
I mean, remove `pypy3` from the run matrix. We can re-add it when the issue is solved.

gaborbernat:
I don't think we need to remove it from the matrix. We need to just mark one tests skipped.

FFY00:
Sure. Do you want to open a PR?

gaborbernat:
Resolved by https://github.com/FFY00/python-build/pull/125

PR #125, opened 2020-10-07
Fix failing CI

### LINKED ISSUE #124: pypy3 on Windows ensurepip is not borked
At the moment in our codebase we special handle it, but we shouldn't. See https://foss.heptapod.net/pypy/pypy/-/issues/3323 for the root issue. IMHO we can keep just always the with pip branch, and disable the isolation test until it's fixed together with https://foss.heptapod.net/pypy/pypy/-/issues/3321

gaborbernat:
Resolves #124 and #123.

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -1,4 +1,5 @@
 import os
+import platform
 import shutil
 import subprocess
 import sys
@@ -11,9 +12,6 @@ from typing import Dict, Iterable, List, Optional, Sequence, Type
 
 if sys.version_info[0] == 2:  # pragma: no cover
     FileExistsError = OSError
-else:
-    import venv
-
 
 _HAS_SYMLINK = None  # type: Optional[bool]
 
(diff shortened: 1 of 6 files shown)


src/build/env.py
```diff
@@ -185,18 +186,12 @@ def _create_env_pythonhome(self):  # type: () -> None
...
-            else:
-                venv.EnvBuilder(with_pip=True).create(self.path)
+        venv.EnvBuilder(with_pip=True).create(self.path)
 
         env_scripts = self._get_env_path('scripts')
         if not env_scripts:
             raise RuntimeError("Couldn't get environment scripts path")
-
-        exe = 'python'
+        exe = 'pypy3' if platform.python_implementation() == 'PyPy' else 'python'
```
gaborbernat, line 194: `python` is optional, and while virtualenv creates it, venv only does the minimal and creates only `pypy3.exe`.

gaborbernat:
@FFY00 want to approve? It's ready now.

FFY00 (approved):

PR #126, opened 2020-10-08
Add support for Python 3.9

gaborbernat:
future is here

PR #127, opened 2020-10-08
Upgrade to new setup-cfg-fmt

gaborbernat:


setup.cfg
```diff
@@ -39,15 +40,15 @@ console_scripts =
 
 [options.extras_require]
 doc =
+    Pallets-Sphinx-Themes
```
FFY00, line 43: (unrelated)

Oh, this is outdated. Should be `furo` instead.

If you want you can open a PR, otherwise I'll do it later.

setup.cfg
```diff
@@ -16,6 +16,7 @@ classifiers =
     Programming Language :: Python :: 3.6
     Programming Language :: Python :: 3.7
     Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
```
FFY00, line 19: Shouldn't this be in #126?

.pre-commit-config.yaml
```diff
@@ -22,9 +22,10 @@ repos:
   hooks:
   - id: isort
 - repo: https://github.com/asottile/setup-cfg-fmt
-  rev: v1.9.0
+  rev: v1.14.0
   hooks:
   - id: setup-cfg-fmt
+    args: ["--min-py3-version", "3.5"]
```
FFY00, line 28: Can you expand on this? What does it mean and why is it needed?

I want to keep the project buildable with python2, for distro workflows. Otherwise I'd have to have a dependency on Python 3 to build `build` for Python 2.

FFY00 (changes requested):

setup.cfg
gaborbernat: IMHO it's out of scope for this pr. Let's do it separately.

setup.cfg
gaborbernat: I've actually included py39 in tox.ini before so it picks it up...

setup.cfg
FFY00: Yes.

.pre-commit-config.yaml
gaborbernat: This is just metadata for setup-cfg-fmt, of what's the minimum python 3 version, default is python 3.6 for it. So we're overwriting that as we still support 3.5.

setup.cfg
FFY00: Ah, fair enough. If we can, I'd like to wait for #126. If it has problems materializing we'd have to remove this classifier.

.pre-commit-config.yaml
FFY00: Ah, okay :+1:

setup.cfg
gaborbernat: I don't think we need per se. Identified the issue and it's a virtualenv bug... Our test suite is created with virtualenv, hence showing up in the ci .. but as you think 👍not sure if we can limit the upper constraint here 🤔

gaborbernat:
@FFY00 your concerns has been addressed.

FFY00 (approved):

Issue #120, opened 2020-10-07
PyPy2 isolation broken


gaborbernat:
pypy does not handles/honor/cares about PYTHONHOME, see https://foss.heptapod.net/pypy/pypy/-/blob/branch/default/pypy/interpreter/app_main.py#L38, hence the issues here; prefixes remain under original executable path, which then pulls in the site-packages via https://foss.heptapod.net/pypy/pypy/-/blob/branch/default/lib-python/2.7/site.py#L295-297 and https://foss.heptapod.net/pypy/pypy/-/blob/branch/default/lib-python/2.7/distutils/sysconfig_pypy.py#L38-56 respectively

minimal reproducible:

```
env PYTHONHOME=this-is-missing pypy -E -m site
sys.path = [
    '/Users/bgabor8/git/github/python-build',
    '/Users/bgabor8/.pyenv/versions/pypy2.7-7.3.0/lib_pypy/__extensions__',
    '/Users/bgabor8/.pyenv/versions/pypy2.7-7.3.0/lib_pypy',
    '/Users/bgabor8/.pyenv/versions/pypy2.7-7.3.0/lib-python/2.7',
    '/Users/bgabor8/.pyenv/versions/pypy2.7-7.3.0/lib-python/2.7/lib-tk',
    '/Users/bgabor8/.pyenv/versions/pypy2.7-7.3.0/lib-python/2.7/plat-darwin',
    '/Users/bgabor8/.pyenv/versions/pypy2.7-7.3.0/lib-python/2.7/plat-mac',
    '/Users/bgabor8/.pyenv/versions/pypy2.7-7.3.0/lib-python/2.7/plat-mac/lib-scriptpackages',
    '/Users/bgabor8/.pyenv/versions/pypy2.7-7.3.0/site-packages',
]
USER_BASE: '/Users/bgabor8/.local' (exists)
USER_SITE: '/Users/bgabor8/.local/lib/pypy2.7/site-packages' (doesn't exist)
ENABLE_USER_SITE: True
```

vs

```
env PYTHONHOME=this-is-missing python2.7 -m site
ImportError: No module named site
```

FFY00:
What do you propose to do? If `PYTHONHOME` is ignored how do virtual environments work there?

gaborbernat:
> What do you propose to do? If `PYTHONHOME` is ignored how do virtual environments work there?

[...]

PR #140, opened 2020-10-14
Borrow pip from host python if available

gaborbernat:
Instead of provisioning via ensurepip. This speeds up isolated build environment creation.

While implementing this I had to also address https://github.com/pypa/build/issues/120 to make the CI pass. They're separate commits though. 

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -13,7 +13,7 @@ if sys.version_info[0] == 2:
 
 try:
     import pip
-except ImportError:
+except ImportError:  # pragma: no cover
     pip = None  # pragma: no cover
 
 _HAS_SYMLINK = None  # type: Optional[bool]
@@ -85,20 +85,10 @@ class IsolatedEnvironment(object):
         Creates an isolated environment for the current interpreter
         """
         remove_paths = os.environ.get('PYTHONPATH', '').split(os.pathsep)
-
-        for path in cls._MANIPULATE_PATHS:
-            our_path = sysconfig.get_path(path)
-            if our_path:
-                remove_paths.append(our_path)
-
-            for scheme in sysconfig.get_scheme_names():
-                our_path = sysconfig.get_path(path, scheme)
-                if our_path:
-                    remove_paths.append(our_path)
         # also remove the pth file adding this project (when installed via --develop install)
         for path in sys.path:
             if os.path.exists(os.path.join(path, 'build.egg-info')):
-                remove_paths.append(path)
+                remove_paths.append(path)  # pragma: no cover
 
         return cls(remove_paths)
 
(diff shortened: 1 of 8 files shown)


src/build/env.py
```diff
@@ -6,11 +6,16 @@
 import sysconfig
 import tempfile
 import types
-from typing import Dict, Iterable, List, Optional, Sequence, Type
+from typing import Dict, Iterable, List, Optional, Sequence, Type, cast
```
FFY00, line 9: I strongly prefer having `typing.cast` instead of `cast`. I generally only use partial imports for `typing` types.

src/build/env.py
```diff
@@ -245,19 +274,19 @@ def install(self, requirements):  # type: (Iterable[str]) -> None
...
 
-        if sys.version_info[0] == 2:
-            subprocess.check_call([self.executable, '-m', 'ensurepip'], cwd=self.path)
-
         with tempfile.NamedTemporaryFile('w+', prefix='build-reqs-', suffix='.txt', delete=False) as req_file:
             req_file.write(os.linesep.join(requirements))
             req_file.close()
             cmd = [
-                self.executable,
+                cast(str, self._pip_executable),
```
FFY00, line 226: Please raise an error if `self._pip_executable` is `None` instead. That approach is better at surviving refactors or other changes.

tests/conftest.py
```diff
@@ -35,6 +35,8 @@ def _setup():
 
 
 def pytest_addoption(parser):
+    os.environ['PYTHONWARNINGS'] = 'ignore:DEPRECATION::pip._internal.cli.base_command'  # for when not run within tox
```
FFY00, line 38: Can you provide some background on why this is needed?

FFY00 (approved):
Some nitpicks, otherwise LGTM

src/build/env.py
gaborbernat: Considering all other elements from typing are directly imported to me feels strange to have cast be imported separately. Strongly...

src/build/env.py
gaborbernat: The way the code it's written that use case cannot happen... feels redundant to have a branch we don't cover. IMHO the way to ensure we handle various use cases is by tests and coverage, rather than redundant defensive checks.

tests/conftest.py
gaborbernat: On python2.7 pip prints a warning that people should not use old python, which pollutes the stdout and causes issues when we parse the stdout to assert on content of it. This disables that warning. We know python2.7 is EOL, but we want to support it nonetheless, for us that warning is not needed.

src/build/env.py
FFY00: If for anything else, `mypyc` works fine right now, using `typing.cast` will break that. I generally try to keep of `typing.cast` to the absolute minimum.

Also remember that people can subclass this, so I don't think this check would be *that* defensive.

gaborbernat: >If for anything else, mypyc works fine right now, using typing.cast will break that.

How? It passes for me as is now here.

If they subclass it and break the contract of using the class it's on them. Also if this is None it will fail with an error that's self descriptive, not sure what's the value in raising a custom error.

FFY00: Alright. I still have a slight preference on having the check but go ahead.

gaborbernat: To be fair I want to move the isolation creation to the ``__init__`` rather than ``__enter__``, and keep the ``__enter__`` almost empty. That way the executable and pip executable are guaranteed to be not None. This class is not useful without the ``__enter__`` so maybe worth doing all the setup within init. But that's orthogonal to this change, so was thinking to do it in a follow-up PR?

FFY00: That seems fine.

FFY00 (approved):
All good, go ahead and merge once the CI finishes :blush:

src/build/env.py
pganssle: Sorry if I'm late to the game here, but a low cost option I've taken in the past for stuff like this would be just `assert self._pop is not None`.

PR #144, opened 2020-10-17
Refactor to simplify the env package

gaborbernat:
Create the isolated build environment upfront to avoid executable, pip executable or path not yet set. Remove unused code segments.

As promised in https://github.com/pypa/build/pull/140 this is the follow-up PR to clean up the env code a bit and avoid some redundant type casts. There have been many code segments we don't need actually anymore with the switch to venv/virtualenv. While doing it I've taken the liberty to add a few tests to get 100% coverage for the module.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -78,6 +78,12 @@ omit =
     *bin/pyproject-build
     *bin\pyproject-build.exe
 
+[coverage:report]
+exclude_lines =
+    \#\s*pragma: no cover
+    ^\s*raise NotImplementedError\b
+    ^if __name__ == ['"]__main__['"]:$
+
 [coverage:paths]
 source =
     src
--- a/setup.cfg
+++ b/setup.cfg
@@ -50,6 +50,9 @@ test =
     pytest
     pytest-cov
     pytest-mock
+typing =
+    mypy==0.790
+    typing-extensions>=3.7.4.3
 
 [options.packages.find]
 where = src
@@ -82,7 +85,6 @@ omit =
 exclude_lines =
     \#\s*pragma: no cover
     ^\s*raise NotImplementedError\b
-    ^if __name__ == ['"]__main__['"]:$
 
 [coverage:paths]
 source =
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,7 +26,7 @@ project_urls =
 packages = find:
 install_requires =
     packaging
-    pep517
+    pep517>=0.9.1
     toml
     importlib-metadata;python_version < "3.8"
     typing;python_version<"3"
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -117,6 +117,7 @@ def _working_directory(path):  # type: (str) -> Iterator[None]
 
 class ProjectBuilder(object):
     def __init__(self, srcdir='.', config_settings=None, python_executable=sys.executable):
+        # type: (str, Optional[ConfigSettings], str) -> None
         """
         Create a project builder.
 
@@ -124,7 +125,6 @@ class ProjectBuilder(object):
         :param config_settings: config settings for the build backend
         :param python_executable: the python executable where the backend lives
         """
-        # type: (str, Optional[ConfigSettings], str) -> None
         self.srcdir = os.path.abspath(srcdir)
         self.config_settings = config_settings if config_settings else {}
 
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -116,11 +116,15 @@ def _working_directory(path):  # type: (str) -> Iterator[None]
 
 
 class ProjectBuilder(object):
-    def __init__(self, srcdir='.', config_settings=None, executable=sys.executable):
+    def __init__(self, srcdir='.', config_settings=None, python_executable=sys.executable):
+        """
+        Create a project builder.
+
+        :param srcdir: the source directory
+        :param config_settings: config settings for the build backend
+        :param python_executable: the python executable where the backend lives
+        """
         # type: (str, Optional[ConfigSettings], str) -> None
-        """
-        :param srcdir: Source directory
-        """
         self.srcdir = os.path.abspath(srcdir)
         self.config_settings = config_settings if config_settings else {}
 
@@ -151,7 +155,10 @@ class ProjectBuilder(object):
         self._backend = self._build_system['build-backend']
 
         self.hook = pep517.wrappers.Pep517HookCaller(
-            self.srcdir, self._backend, backend_path=self._build_system.get('backend-path'), python_executable=executable
+            self.srcdir,
+            self._backend,
+            backend_path=self._build_system.get('backend-path'),
+            python_executable=python_executable,
         )
 
     @property
@@ -189,7 +196,6 @@ class ProjectBuilder(object):
 
         :param distribution: Distribution to build (sdist or wheel)
         :param outdir: Output directory
-        :param executable: executable
         """
         build = getattr(self.hook, 'build_{}'.format(distribution))
         outdir = os.path.abspath(outdir)
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -10,13 +10,11 @@ import difflib
 import os
 import sys
 import warnings
-
 from typing import Dict, Iterator, List, Mapping, Optional, Set, Union
 
 import pep517.wrappers
 import toml
 import toml.decoder
-
 
 if sys.version_info < (3,):
     FileNotFoundError = IOError
@@ -118,7 +116,8 @@ def _working_directory(path):  # type: (str) -> Iterator[None]
 
 
 class ProjectBuilder(object):
-    def __init__(self, srcdir='.', config_settings=None):  # type: (str, Optional[ConfigSettings]) -> None
+    def __init__(self, srcdir='.', config_settings=None, executable=sys.executable):
+        # type: (str, Optional[ConfigSettings], str) -> None
         """
         :param srcdir: Source directory
         """
@@ -152,7 +151,7 @@ class ProjectBuilder(object):
         self._backend = self._build_system['build-backend']
 
         self.hook = pep517.wrappers.Pep517HookCaller(
-            self.srcdir, self._backend, backend_path=self._build_system.get('backend-path')
+            self.srcdir, self._backend, backend_path=self._build_system.get('backend-path'), python_executable=executable
         )
 
     @property
@@ -190,6 +189,7 @@ class ProjectBuilder(object):
 
         :param distribution: Distribution to build (sdist or wheel)
         :param outdir: Output directory
+        :param executable: executable
         """
         build = getattr(self.hook, 'build_{}'.format(distribution))
         outdir = os.path.abspath(outdir)
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -12,20 +12,19 @@ import tempfile
 from types import TracebackType
 from typing import Iterable, Optional, Tuple, Type
 
+from ._compat import abstractproperty, add_metaclass
+
 try:
     import pip
 except ImportError:  # pragma: no cover
     pip = None  # pragma: no cover
 
 
-ABC = abc.ABCMeta('ABC', (object,), {'__slots__': ()})  # Python 2/3 compatible ABC
-
-
-class IsolatedEnvironment(ABC):
+@add_metaclass(abc.ABCMeta)
+class IsolatedEnv(object):
     """Abstract base of isolated build environments, as required by the build project."""
 
-    @property
-    @abc.abstractmethod
+    @abstractproperty
     def executable(self):  # type: () -> str
         """Return the executable of the isolated build environment."""
         raise NotImplementedError
@@ -41,11 +40,11 @@ class IsolatedEnvironment(ABC):
 
 
 class IsolatedEnvBuilder(object):
-    def __init__(self):
+    def __init__(self):  # type: () -> None
         """Builder object for isolated environment."""
         self._path = None  # type: Optional[str]
 
-    def __enter__(self):  # type: () -> IsolatedEnvironment
+    def __enter__(self):  # type: () -> IsolatedEnv
         """
         Creates an isolated build environment.
 
@@ -55,7 +54,7 @@ class IsolatedEnvBuilder(object):
         try:
             executable, pip_executable = _create_isolated_env(self._path)
             return _IsolatedEnvVenvPip(path=self._path, python_executable=executable, pip_executable=pip_executable)
-        except Exception:  # cleanup if creation fails
+        except Exception:  # cleanup folder if creation fails
             self.__exit__(*sys.exc_info())
             raise
 
@@ -72,7 +71,7 @@ class IsolatedEnvBuilder(object):
             shutil.rmtree(self._path)
 
 
-class _IsolatedEnvVenvPip(IsolatedEnvironment):
+class _IsolatedEnvVenvPip(IsolatedEnv):
     """
     Isolated build environment context manager
 
@@ -212,5 +211,5 @@ else:
 
 __all__ = (
     'IsolatedEnvBuilder',
-    'IsolatedEnvironment',
+    'IsolatedEnv',
 )
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -1,6 +1,7 @@
 """
 Creates and manages isolated build environments.
 """
+import abc
 import os
 import platform
 import shutil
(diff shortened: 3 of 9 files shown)


src/build/env.py
```diff
@@ -1,221 +1,158 @@
...
-def fs_supports_symlink():  # type: () -> bool
-    global _HAS_SYMLINK
-    if _HAS_SYMLINK is None:
-        _HAS_SYMLINK = _fs_supports_symlink()
-    return _HAS_SYMLINK
-
 
-class IsolatedEnvironment(object):
+@contextlib.contextmanager
+def isolated_env():  # type: () -> Generator[IsolatedEnvironment, None, None]
```
gaborbernat, line 21: We could put this to be a class method but wasn't sure if it's a good idea to annotate a class method with a ``contextmanager``. And I don't think this class makes sense without always using ``__enter__`` right after creation.

FFY00:
I am fine with the refactoring, as well as moving the environment creation to `__init__`, but what warrants breaking the API? Is there any good motivation behind that or is it just a nice to have? I'd greatly prefer to keep API compatibility.

src/build/env.py
FFY00: Can we keep the class? You can move the environment creation to `__init__`, which makes more sense. This way we could create and manage a virtual environment without ever entering the context manager, and then enter the context manager when we need it. This way we can create the environment one time and enter it multiple times.

gaborbernat:
> I am fine with the refactoring, as well as moving the environment creation to `__init__`, but what warrants breaking the API?

We haven't yet officially released the tool yet (hence version ``0.0.4``) and as such, I don't think we are in a state yet where we must ensure API compatibility with the earlier version. Seconds I feel like the API was created as optimized for the old way of isolation builds, and as such a bit incompatible with where we landed now. 

The biggest problem with the current API is my comment above:

> We could put this to be a class method but wasn't sure if it's a good idea to annotate a class method with a contextmanager. And I don't think this class makes sense without always using __enter__ right after creation.

The earlier API IMHO is not good with the constraint that this class must be entered to be used. It allowed people to do:

```
env = IsolatedBuildEnv.for_x()

# here basically any method call on env is a really bad idea and undefined behavior, and as such IMHO the API itself should not allow it rather than adding defensive error raises at runtime

with env:
	pass 
```

The new API achieves exactly this.

FFY00:
> The earlier API IMHO is not good with the constraint that this class must be entered to be used. It allowed people to do:
> 
> ```python
> env = IsolatedBuildEnv.for_x()
> 
> # here basically any method call on env is a really bad idea 
> # and undefined behavior, and as such IMHO the API itself should
> # not allow it rather than adding defensive error raises at runtime
> 
> with env:
> 	pass 
> ```
> 
> The new API achieves exactly this.
> 
> By the way how much of the API you consider public? I'd personally restrict it to just the main method, until people provide use cases where they'd need more.

I agree, and that is what I suggested in the code comment.

Overall I do not like the unbundling of the environment creation from `IsolatedEnvironment`. Let's keep it there, please :blush:

src/build/env.py
gaborbernat: But that allows:

```python
env = IsolatedBuildEnv.for_x()

# here basically any method call on env is a really bad idea 
# and undefined behavior, and as such IMHO the API itself should
# not allow it rather than adding defensive error raises at runtime

with env:
	pass 
```

I feel if people want to use the same environment for multiple builds they likely want:

```python

with isolated_env() as env:
     env.install()
     for a in (a, b):
        env.build()
```

That being said can you define a use case when people want to reuse isolated build environments. If you share them across packages they are no longer specific to the package 🤔 Do we really want that? Especially now that creating an environment is very low cost (<10ms, IMHO they are better off always recreating it).

FFY00:
Sorry, misread your comment.

gaborbernat:
> > The earlier API IMHO is not good with the constraint that this class must be entered to be used. It allowed people to do:
> > 
> > ```python
> > env = IsolatedBuildEnv.for_x()
> > 
> > # here basically any method call on env is a really bad idea 
> > # and undefined behavior, and as such IMHO the API itself should
> > # not allow it rather than adding defensive error raises at runtime
> > 
> > with env:
> > 	pass 
> > ```
> > 
> > The new API achieves exactly this.
> > 
> > By the way how much of the API you consider public? I'd personally restrict it to just the main method, until people provide use cases where they'd need more.
> 
> I agree, and that is what I suggested in the code comment.
> 
> Overall I do not like the unbundling of the environment creation from `IsolatedEnvironment`. Let's keep it there, please :blush:

But calling env.build without entering the context of the class is also broken behavior...

FFY00:
> here basically any method call on env is a really bad idea 
> and undefined behavior, and as such IMHO the API itself should
> not allow it rather than adding defensive error raises at runtime

The only public method in `IsolatedEnvironment` is `install`, which can be called outside the environment. Could you elaborate?

src/build/env.py
pganssle: I think we should probably remove the `for_current` functionality entirely, and replace anywhere that we use it ourselves with something that spawns a new interpreter.

Activating and then de-activating a virtual environment _during an interpreter run_ seems very likely to cause weird bugs. What happens if, for example, the environment where you've installed `build` has the latest version of `toml`, and the version in the isolated environment is pinned to an older version? If you've followed a code path where `build` (or anything using the context manager) has imported `toml`, the "isolated" environment will get the outer environment's `toml`.

You'll run into other problems if you try to make the context manager also manage the `sys.module` cache, since something like this will fail:

```python

def f() -> foo.Bar:
    import foo
    return foo.Bar(1)

with env:
    x : foo.Bar = execute_something()

if x < foo.Bar(3):
   ...
```

This will probably have references to two different `foo.Bar` classes and fail!

If there's a context manager involved, it should manage the lifetime of the environment's existence (á la `tempfile.TemporaryDirectory`), not a context in which the environment is _active_.

gaborbernat:
I'll amend the PR in the morning.

FFY00:
@gaborbernat I hope I am not being too pushy, we can keep discussing this if you want.

src/build/env.py
FFY00: We do not change anything in the running interpreter session. `for_current` sets up a virtual environment based on the current interpreter. It was added because @uranusjr wanted to create an environment from an external interpreter.

```python
x : foo.Bar = execute_something()
```
runs in the current interpreter session. If you want to do something in the environment you will have to spawn a subprocess. `pep517` does this, and that is why we do not run it in a subprocess, it runs the subprocess for us.

I do not see a different in allowing a `IsolatedEnvironment` instance to run outside the context manager. Having the environment "activated" should not have impact on the current interpreter session.

gaborbernat:
> What is this `env.build`?

My point was that achieving the workaround for https://github.com/pypa/pep517/pull/93 we patched the ``sys.executable`` within our context manager of the class. If one would call ``builder.build`` without using the context manager (but just creates the environment) things can still be misused. That being said with the merge and release of that PR this is no longer the case, so I've now adopted the release of https://github.com/pypa/pep517/pull/93 into this PR to avoid this problem.

src/build/env.py
```diff
@@ -51,191 +24,166 @@ class IsolatedEnvironment(object):
...
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        # type: (Optional[Type[BaseException]], Optional[BaseException], Optional[TracebackType]) -> None
+        """
+        Exit from the isolated build environment.
+
+        :param exc_type: the type of exception raised (if any)
+        :param exc_val: the value of exception raised (if any)
+        :param exc_tb: the traceback of exception raised (if any)
+        """
+        self.close()
```
gaborbernat, line 64: Users who want to reuse an isolated build environment multiple times should not invoke this class's context manager. However, they must invoke ``close`` at the end of their application to trigger the cleanup.

tests/test_main.py
```diff
@@ -87,39 +87,54 @@ def test_prog():
 
 
 @pytest.mark.isolated
-def test_build(mocker, test_flit_path):
```
gaborbernat, line 90: This test actually contained five different tests, and because was causing me some issues I've split it up into five separate tests within this PR.

src/build/env.py
```diff
@@ -51,191 +24,166 @@ class IsolatedEnvironment(object):
...
+
+        :return: the isolated build environment
         """
-        self._env = {}  # type: Dict[str, Optional[str]]
-        self._env_vars = {}  # type: Dict[str, str]
-        self._path = None  # type: Optional[str]
-        self._remove_paths = []  # type: List[str]
-        self._executable = _executable
-        self._old_executable = None  # type: Optional[str]
-        self._pip_executable = None  # type: Optional[str]
```
gaborbernat, line 66: I've realized that ``pip_executable`` is assuming too much. In the case of `conda` for example, the user would be most likely to use `conda`. To acknowledge this I've renamed this to `instal_executable`.

gaborbernat:
@FFY00 this is now ready for review.

src/build/__init__.py
```diff
@@ -118,7 +116,8 @@ def _working_directory(path):  # type: (str) -> Iterator[None]
 
 
 class ProjectBuilder(object):
-    def __init__(self, srcdir='.', config_settings=None):  # type: (str, Optional[ConfigSettings]) -> None
+    def __init__(self, srcdir='.', config_settings=None, executable=sys.executable):
```
FFY00, line 119: Can this be `python_executable` instead? I think that is a bit clearer.

src/build/env.py
```diff
@@ -51,191 +24,166 @@ class IsolatedEnvironment(object):
...
-        # normalize paths so that we can compare them -- required on case insensitive systems
-        for path in remove_paths:
-            self._remove_paths.append(os.path.normcase(path))
+        path = tempfile.mkdtemp(prefix='build-env-')
+        executable, pip_executable = _create_isolated_env(path)
+        return cls(path=path, executable=executable, install_executable=pip_executable)
+
+    def __enter__(self):  # type: () -> IsolatedEnvironment
+        """Enable the isolated build environment"""
+        return self
```
FFY00, line 53: We should adjust `PATH` here so that you can call scripts from the environment.

src/build/env.py
FFY00: I don't think we should call `close` here. This should deactivate the environment, not close it.

We should call `close` instead in `__del__`. I am not even sure if there is need for an explicit destructor. IMO we should just rename `close` to `__del__`.

FFY00 (changes requested):

src/build/env.py
FFY00: I would prefer to keep the original name. If you want to use conda you should subclass `IsolatedEnvironment`, where you can have a `_conda_executable` or something else.

src/build/env.py
gaborbernat: I'm strongly against that, but can you expand on why? 

My reasoning is IMHO mangling with the path is always a really bad idea. See https://bugs.python.org/issue42041 for a recent bug and discussion on why.

src/build/env.py
gaborbernat: Well, potentially. It's a question of design. But then the base should be an abstract class because within conda you should not have pip.

src/build/env.py
gaborbernat: That's very much against the language design and guarantees. Relying on ``__del__`` to be called is just not supported/guaranteed within the Python language. See here https://bugs.python.org/issue39360 for more information about a nasty bug this can cause, and where core developers express that in Python you should always explicitly cleanup and never rely on ``__del__`` to be called for resources.

src/build/env.py
FFY00: Sure. Then we should implement a `get_script` method. I can do this in a later PR.

gaborbernat: Why and where would you need it though? Can you expand on that?

src/build/env.py
FFY00: If you want to explicitly clean up the temporary directory, you can call the destructor. But you are not necessarily required to. We are not relying on `__del__` to be called for anything.
If you want to keep `env.close()` as an alternative to `del env`, sure.

The main point here is that I do not think we should be forcefully cleaning up the environment when we exit. Maybe we can pass an argument to opt in/out?

src/build/env.py
FFY00: I need it if I want to call a script that is provided by a package.

One could argue that you should be able to call entrypoints directly without need for a script, but they are not always public API. And regardless, per PEP 427, packages may bring their own scripts, so I have no other alternative.

src/build/env.py
gaborbernat: I think the only thing you should be cleaning up within a ``__del__`` method is memory. Not folders and files.

> The main point here is that I do not think we should be forcefully cleaning up the environment when we exit

Why? We created it just to be able to perform the build. Why would we not clean it up? At least within build. If we're talking about some custom application they can always use the API and not call the close /context amanger if they want to reuse across runs.

src/build/env.py
gaborbernat: > I need it if I want to call a script that is provided by a package.

> And regardless, per PEP 427, packages may bring their own scripts, so I have no other alternative than to call the script directly.

The above still doesn't answer the question of where and why would you want to call a script?

src/build/env.py
gaborbernat: Would you like an abstract base class `IsolatedEnvironment` that defines the API as expected by the project builder, and then we add implementation `IsolatedEnvironmentVenvPip`? Because then yes naming it directly ``pip_executable`` makes a lot of sense.

gaborbernat:
@FFY00 @pganssle made the requested changes 👍 please review again.

src/build/env.py
uranusjr: Why do we need this instead of `[self._executable, "-m", "pip"]`?

gaborbernat: Because there's no reason for us to have to provision pip inside the isolated environment, we can just use the pip install alongside the pip project and pass in ``--prefix`` to this path.

uranusjr: Would `'sys.executable', '-m', 'pip'` not work in that case?

gaborbernat: Not in general, hence why is a config value. Sometimes the host doesn't have pip, and in such cases, it's good to fall back to ensurepip. Why you don't like it?

uranusjr: I prefer `-m` as a rule since entry points add an additional layer of indirection that can easily go wrong in edge cases.

I also wanted to reduce the number of instance members since the more there are the more possibility things can go wrong, but from your explaination above, that’s no longer an issue now since there’s no way to avoid it.

gaborbernat: Not we're still using -m (if you read the install method implementation), the pip_executable is not actually the pip executable script, but the python executable having the pip inside 🤔 The name is not best, but it's an implementation detail (part of a private class).

uranusjr: I see, thanks for the clarification.

setup.cfg
```diff
@@ -50,6 +50,9 @@ test =
     pytest
     pytest-cov
     pytest-mock
+typing =
+    mypy==0.782
```
FFY00, line 54: Why pin the dependency?

setup.cfg
```diff
@@ -78,6 +81,12 @@ omit =
     *bin/pyproject-build
     *bin\pyproject-build.exe
 
+[coverage:report]
+exclude_lines =
+    \#\s*pragma: no cover
+    ^\s*raise NotImplementedError\b
+    ^if __name__ == ['"]__main__['"]:$
```
FFY00, line 88: Why? We test this.

src/build/env.py
```diff
@@ -1,241 +1,215 @@
...
-            if val is None:
-                os.environ.pop(key, None)
-            else:
-                os.environ[key] = val
+if sys.version_info[0] == 2:  # noqa: C901 # disable if too complex
 
-    def _get_env_path(self, path):  # type: (str) -> Optional[str]
-        """
-        Returns sysconfig path from our environment
+    def _create_isolated_env(path):  # type: (str) -> Tuple[str, str]
```
FFY00, line 140: Wouldn't make sense for this to be in `IsolatedEnvBuilder`?

setup.cfg
```diff
@@ -78,6 +81,12 @@ omit =
     *bin/pyproject-build
     *bin\pyproject-build.exe
 
+[coverage:report]
+exclude_lines =
+    \#\s*pragma: no cover
```
FFY00, line 86: Why?

FFY00 (approved):
LGTM :+1: Just some minor comments.

setup.cfg
gaborbernat: Because mypy is getting better with every release. We don't want the CI to break just because a new release was done, and now the checks are better and mypy detected a new bug in our app. We'll periodically upgrade.

setup.cfg
gaborbernat: Yes. Inside the ``__main__`` file. This makes that line covered.

src/build/env.py
gaborbernat: Well, maybe, however, these are implementation detail methods. If we move them, you'd need to make these static methods and increase the indent by one level. So overall, probably not worth the change.

setup.cfg
FFY00: I would definitely forget to bump, but if you got that covered :+1:

setup.cfg
FFY00: But we test that via integration tests, it should already be covered :thinking:

setup.cfg
gaborbernat: Here, the core goal is to add ``raise NotImplementedError`` to be automatically marked with no coverage needed for them (because the metaclass makes it impossible to hit it). By default, coverage only excludes `pragma no cover` lines, so we want to extend that. There's no extend the exclude method, though, override of the exclude lines matcher, so we need to set it again.

setup.cfg
gaborbernat: Yeah 👍

PR #149, opened 2020-10-20
main: run isolation build with the environment python interpreter

FFY00:


setup.cfg
```diff
@@ -26,7 +26,7 @@ project_urls =
 packages = find:
 install_requires =
     packaging
-    pep517
+    pep517>=0.9
```
gaborbernat, line 29: Well, this a dangerous form. The minimum version required is `0.9.1`. There hasn't been released a `0.9.0`, so this works, but is there any reason why you don't just fully specify the exact minimum version that's guaranteed not to break? Otherwise, an erroneous 0.9.0 release can break us.

src/build/__main__.py
```diff
@@ -40,15 +40,19 @@ def _error(msg, code=1):  # type: (str, int) -> None  # pragma: no cover
     exit(code)
 
 
-def _build_in_isolated_env(builder, outdir, distributions):  # type: (ProjectBuilder, str, List[str]) -> None
+def _build_in_isolated_env(srcdir, config_settings, outdir, distributions):
+    # type: (str, Optional[ConfigSettings], str, List[str]) -> None
     with IsolatedEnvironment.for_current() as env:
+        builder = ProjectBuilder(srcdir, config_settings, python_executable=env.executable)
```
gaborbernat, line 46: I'm not a fan of duplicating the project builder creation and moving it inside the respective methods. I prefer my solution from https://github.com/gaborbernat/python-build/blob/50c6a93891/src/build/__main__.py#L44

gaborbernat (changes requested):

setup.cfg
FFY00: There has, it was just not released to PyPI. People who get releases from Github may be using it. The fix was introduced in 0.9.0, not 0.9.1.

src/build/__main__.py
FFY00: Hum, sure, we can do that.

FFY00: I have made `python_executable` as property that mirrors the hook's `python_executable`. I think we should make the hook private, that can come in the next PR.

gaborbernat (approved):

src/build/__init__.py
```diff
@@ -21,6 +21,8 @@
 if sys.version_info < (3,):
     FileNotFoundError = IOError
     PermissionError = OSError
+else:
+    unicode = str
```
gaborbernat, line 25: Why not use typing.Text?

src/build/__init__.py
```diff
@@ -118,7 +120,8 @@ def _working_directory(path):  # type: (str) -> Iterator[None]
 
 
 class ProjectBuilder(object):
-    def __init__(self, srcdir='.', config_settings=None):  # type: (str, Optional[ConfigSettings]) -> None
+    def __init__(self, srcdir='.', config_settings=None, python_executable=sys.executable):
```
gaborbernat, line 123: Why not use typing.Text?

src/build/__init__.py
```diff
@@ -152,9 +155,22 @@ def __init__(self, srcdir='.', config_settings=None):  # type: (str, Optional[Co
...
         )
 
+    @property
+    def python_executable(self):  # type: () -> unicode
+        if not isinstance(self.hook.python_executable, unicode):
+            raise ValueError('python_executable is not a string')
+        return self.hook.python_executable
+
+    @python_executable.setter
+    def python_executable(self, value):  # type: (str) -> None
```
gaborbernat, line 171: Why not use typing.Text?

gaborbernat (changes requested):

src/build/__init__.py
pganssle: I don't think `typing.Text` is appropriate here, since `value` refers to a path of some sort, which is `str` on both Python 2 and 3.

Though I think it might be better if we did something like this:

```python
if sys.version_info[0] == 2:
    PathType = Union[str, os.Pathlike]
else:
    PathType = str
```

Since presumably we do want to accept `pathlib.Path` objects.

FFY00: We are doing a check.

```python
if not isinstance(self.hook.python_executable, unicode):
```

pganssle: Shouldn't that fail on Python 2.7 if the setter only accepts `str`?

And what is the rationale for that? In Python 2.7 the natural way to work with paths is with `bytes`.

gaborbernat: pep517 accepts only ``Text`` path for python executable (for some reason), hence IMHO we should mirror here that requirement (we could accept bytes on python 2, and decode it if we must)

layday: I might be wrong, but from what I can see, pep517 accepts whatever `subprocess.check_call` accepts, either Unicode or bytes will work as well as `Path`s in Python 3.6+.  Shouldn't the check be done in the setter anyway or is `ProjectBuilder.python_executable` used in some other fashion?

FFY00: @layday thanks for the input :blush:

I have modified it to accept either, `str`, `bytes` and `unicode` (on Python 2).

src/build/__init__.py
```diff
@@ -21,9 +21,12 @@
 if sys.version_info < (3,):
     FileNotFoundError = IOError
     PermissionError = OSError
+else:
+    unicode = str
 
 
 ConfigSettings = Dict[str, Union[str, List[str]]]
+Path = Union[str, bytes, unicode]
```
gaborbernat, line 29: Why not use ``Union[bytes, Text]``?

gaborbernat (changes requested):

src/build/__init__.py
FFY00: That doesn't accept normal strings on Python 2, does it?

gaborbernat: `Union[bytes, Text]` does.

layday: `Union[bytes, Text]` is synonymous with `Union[str, bytes, unicode]`.  According to the typeshed stub `subprocess.check_call` accepts both bytes and Unicode in Python 2: https://github.com/python/typeshed/blob/master/stdlib/2/subprocess.pyi#L8.

FFY00: > `Text` is an alias for `str`. It is provided to supply a forward compatible path for Python 2 code: in Python 2, `Text` is an alias for `unicode`.

That's not how I interpret the documentation.

gaborbernat: Exactly, so ``Union[bytes, Text]`` is essentially ``Union[str, bytes, unicode]``; what you have. On Python 3 IMHO we should only accept ``Text`` (which resolves to just ``str``).

layday: `subprocess` accepts `bytes` in Python 3 too so it shouldn't matter unless you have a preference for Unicode.

FFY00:
```
Python 2.7.18 (default, Sep  5 2020, 11:17:26)
[GCC 10.2.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import typing
>>> typing.Text
<type 'unicode'>
```

Am I missing something? Using `typing.Text` would not include `str` on Python 2, right?

layday: No, but `str is bytes` and the proposed format is `Union[bytes, Text]`.  In Python 3 this resolves into `Union[bytes, str]`; in Python 2 it resolves into `Union[bytes, unicode]`.

FFY00: Ah, nevermind. I am dumb :sweat_smile:

gaborbernat (changes requested):

tests/test_projectbuilder.py
```diff
@@ -111,6 +129,14 @@ def test_init(mocker, test_flit_path, legacy_path, test_no_permission, test_bad_
         build.ProjectBuilder(test_bad_syntax_path)
 
 
+def test_python_executable(test_flit_path):
```
gaborbernat, line 132: Do you want to test here passing in bytes too though, not? And explicit Unicode?

src/build/__init__.py
```diff
@@ -161,6 +164,21 @@ def __init__(self, srcdir='.', config_settings=None, python_executable=sys.execu
             python_executable=python_executable,
         )
 
+    @property
+    def python_executable(self):  # type: () -> Union[bytes, Text]
+        print(self.hook.python_executable, type(self.hook.python_executable))
+        if (
+            not isinstance(self.hook.python_executable, str)
+            and not isinstance(self.hook.python_executable, bytes)
+            and not isinstance(self.hook.python_executable, unicode)
```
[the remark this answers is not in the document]
layday: I don't understand what this check accomplishes even if it'd been defined in the setter? Note that the `python_executable` can be passed as an argument to `__init__` and will bypass the `@property` entirely.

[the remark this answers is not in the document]
FFY00: We do not control what comes out of `pep517`, and this makes mypy happy.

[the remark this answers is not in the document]
layday: I suggest adding a comment mentioning that this is done to placate mypy because pep517 is untyped and the type of `self.hook.python_executable` is therefore unknown because it was not obvious to me or Bernát (I think).

src/build/__init__.py
```diff
@@ -161,6 +163,20 @@ def __init__(self, srcdir='.', config_settings=None, python_executable=sys.execu
             python_executable=python_executable,
         )
 
+    @property
+    def python_executable(self):  # type: () -> Union[bytes, Text]
+        print(self.hook.python_executable, type(self.hook.python_executable))
```
layday, line 168: I think you must've inserted the `print` here for debugging?

FFY00: Yes, just removed.

src/build/__init__.py
```diff
@@ -161,6 +163,17 @@ def __init__(self, srcdir='.', config_settings=None, python_executable=sys.execu
             python_executable=python_executable,
         )
 
+    @property
+    def python_executable(self):  # type: () -> Union[bytes, Text]
+        # make mypy happy
+        if not isinstance(self.hook.python_executable, typing.Text) and not isinstance(self.hook.python_executable, bytes):
+            raise TypeError('python_executable is not a string: {}'.format(type(self.hook.python_executable)))
+        return self.hook.python_executable
```
layday, line 171: If it's just to make mypy happy, this would work:

```suggestion
        python_executable = self.hook.python_executable  # type: Union[bytes, Text]
        return python_executable
```

And then you can revert to returning `self.hook.python_executable` directly whenever pep517 adds type hints.

src/build/__init__.py
```diff
@@ -161,6 +163,17 @@ def __init__(self, srcdir='.', config_settings=None, python_executable=sys.execu
             python_executable=python_executable,
         )
 
+    @property
+    def python_executable(self):  # type: () -> Union[bytes, Text]
+        # make mypy happy
+        if not isinstance(self.hook.python_executable, typing.Text) and not isinstance(self.hook.python_executable, bytes):
```
gaborbernat, line 169: I think you can replace this with ``isinstance(self.hook.python_executable, (typing.Text, bytes))`` 👍 no need for the and here 🤔

src/build/__init__.py
```diff
@@ -161,6 +163,17 @@ def __init__(self, srcdir='.', config_settings=None, python_executable=sys.execu
...
+            raise TypeError('python_executable is not a string: {}'.format(type(self.hook.python_executable)))
+        return self.hook.python_executable
+
+    @python_executable.setter
+    def python_executable(self, value):  # type: (Union[bytes, Text]) -> None
+        self.hook.python_executable = value
```
gaborbernat, line 175: Shouldn't the type  validation be performed here, and not during read?

src/build/env.py
```diff
@@ -12,7 +12,7 @@
 from types import TracebackType
 from typing import Iterable, Optional, Tuple, Type
 
-from ._compat import abstractproperty, add_metaclass
+from ._compat import add_metaclass
```
gaborbernat, line 15: Eh, why revert this part. I prefer the way it was previously 🤔

gaborbernat (changes requested):
Move the type validation to the setter, and use the shorter isinstance check.

src/build/env.py
FFY00: It was breaking mypy.

src/build/__init__.py
FFY00: I missed that in the `isinstance` documentation :upside_down_face:

src/build/env.py
gaborbernat: Wait, do you have an old mypy. It worked in the CI and locally for me?

src/build/__init__.py
FFY00: Sure. But the current approach works and does not cost us very much.

src/build/env.py
gaborbernat: Check here https://github.com/pypa/build/runs/1293111599#step:6:7, it passes 👍

src/build/__init__.py
FFY00: It was to make mypy happy. We have type hints here.

src/build/_compat.py
```diff
@@ -25,18 +23,4 @@ def wrapper(cls):  # type: (_T) -> _T
...
-
-    def abstractproperty(func):  # type: (F) -> F
-        return cast(F, property(abc.abstractmethod(func)))
-
-
-__all__ = (
-    'abstractproperty',
-    'add_metaclass',
-)
+__all__ = ('add_metaclass',)
```
gaborbernat, line 26: Please restore this. mypy was happy on the CI and for me locally.

gaborbernat (changes requested):

src/build/__init__.py
gaborbernat: I think we should raise an error if the input is not of those types.

tests/test_projectbuilder.py
```diff
@@ -111,6 +130,15 @@ def test_init(mocker, test_flit_path, legacy_path, test_no_permission, test_bad_
         build.ProjectBuilder(test_bad_syntax_path)
 
 
+@pytest.mark.parametrize('value', [b'something', typing.Text('something_else')])
+def test_python_executable(test_flit_path, value):
```
gaborbernat, line 134: You test the happy path, but not the sad path? IMHO we should do both.

src/build/env.py
FFY00:
```
$ git checkout 6efa710
$ mypy src/build
src/build/__main__.py:45: error: Incompatible types in assignment (expression has type "Callable[[], str]", variable has type "Union[bytes, str]")
Found 1 error in 1 file (checked 5 source files)
```

src/build/_compat.py
FFY00: It wasn't with the changes, probably a mypy bug.

src/build/__init__.py
layday: The rest of the API doesn't perform any kind of input validation - why single out `python_executable`?

tests/test_projectbuilder.py
FFY00: I removed the checks.

src/build/env.py
gaborbernat: eh, ``mypy --version``?

FFY00:
```
$ mypy --version
mypy 0.790
```

The CI also failed.

gaborbernat: Smaller fix for that that keeps the compatibility layer for that separate:

```
--- a/src/build/_compat.py
+++ b/src/build/_compat.py
@@ -32,8 +32,8 @@ else:

     F = TypeVar('F', bound=Callable[..., Any])

-    def abstractproperty(func):  # type: (F) -> F
-        return cast(F, property(abc.abstractmethod(func)))
+    def abstractproperty(func):  # type: (F) -> Any
+        return cast(Any, property(abc.abstractmethod(func)))

```

layday: Returning `Any` will result in loss of type information.  The issue here is that mypy cannot resolve an `abstractproperty` function into a `property`, even if you were to return a simple `property`.  I suggest:

```diff
diff --git a/src/build/_compat.py b/src/build/_compat.py
index 3a5da79..ab95485 100644
--- a/src/build/_compat.py
+++ b/src/build/_compat.py
@@ -1,6 +1,6 @@
 import abc
 import sys
-from typing import Callable, TypeVar
+from typing import TYPE_CHECKING, Callable, TypeVar
 
 _T = TypeVar('_T')
 
@@ -28,12 +28,12 @@ def add_metaclass(metaclass):  # type: (type) -> Callable[[_T], _T]
 if sys.version_info[0] == 2:
     abstractproperty = abc.abstractproperty
 else:
-    from typing import Any, cast
+    if TYPE_CHECKING:
+        abstractproperty = abc.abstractproperty
+    else:
 
-    F = TypeVar('F', bound=Callable[..., Any])
-
-    def abstractproperty(func):  # type: (F) -> F
-        return cast(F, property(abc.abstractmethod(func)))
+        def abstractproperty(func):
+            return property(abc.abstractmethod(func))
 
 
 __all__ = (

```

FFY00:
@gaborbernat please do not merge yet.

gaborbernat:
https://github.com/FFY00/python-build/pull/1 adds the fix

src/build/_compat.py
```diff
@@ -28,12 +28,15 @@ def wrapper(cls):  # type: (_T) -> _T
...
-    F = TypeVar('F', bound=Callable[..., Any])
+    if TYPE_CHECKING:  # pragma: no cover
+        abstractproperty = property  # pragma: no cover
+    else:
+        F = TypeVar('F', bound=Callable[..., Any])
 
-    def abstractproperty(func):  # type: (F) -> F
-        return cast(F, property(abc.abstractmethod(func)))
+        def abstractproperty(func):  # type: (F) -> Any
+            return cast(Any, property(abc.abstractmethod(func)))
```
layday, line 39: The cast here is redundant if you are returning Any though this function is not type checked anyway.

gaborbernat (approved):

PR #150, opened 2020-10-20
build: make pypa/pep517 hook private

FFY00:
On top of #149

### DIFF
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -155,7 +155,7 @@ class ProjectBuilder(object):
 
         self._backend = self._build_system['build-backend']
 
-        self.hook = pep517.wrappers.Pep517HookCaller(
+        self._hook = pep517.wrappers.Pep517HookCaller(
             self.srcdir,
             self._backend,
             backend_path=self._build_system.get('backend-path'),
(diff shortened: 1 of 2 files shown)


gaborbernat (approved):

src/build/__init__.py
```diff
@@ -151,10 +154,23 @@ def __init__(self, srcdir='.', config_settings=None):  # type: (str, Optional[Co
...
-            self.srcdir, self._backend, backend_path=self._build_system.get('backend-path')
+        self._hook = pep517.wrappers.Pep517HookCaller(
+            self.srcdir,
+            self._backend,
+            backend_path=self._build_system.get('backend-path'),
+            python_executable=python_executable,
         )
 
+    @property
+    def python_executable(self):  # type: () -> unicode
```
gaborbernat, line 165: Any reason you're not using here ``typing.Text``, and also annotate as such the python executable in the constructor to not have this weird isinstance check?

src/build/__init__.py
```diff
@@ -151,10 +154,23 @@ def __init__(self, srcdir='.', config_settings=None):  # type: (str, Optional[Co
...
+        if not isinstance(self._hook.python_executable, unicode):
+            raise ValueError('python_executable is not a string')
+        return self._hook.python_executable
+
+    @python_executable.setter
+    def python_executable(self, value):  # type: (str) -> None
```
gaborbernat, line 171: value should be Text

gaborbernat (changes requested):

src/build/__init__.py
```diff
@@ -21,9 +21,12 @@
 if sys.version_info < (3,):
     FileNotFoundError = IOError
     PermissionError = OSError
+else:
+    unicode = str
 
 
 ConfigSettings = Dict[str, Union[str, List[str]]]
+Path = Union[str, bytes, unicode]
```
gaborbernat, line 29: I don't think we should name this `Path` because clashes with `Path` from pathlib. IMHO no need to define this, you can just use `Union[bytes, Text]` though I don't think we should allow bytes on python 3.

gaborbernat (changes requested):

src/build/__init__.py
FFY00: Please review on #149.

gaborbernat:
@FFY00 this did not have approval, you merged it regardless?

PR #153, opened 2020-10-21
tox: don't run integration tests by default

FFY00:
People can opt in via posargs.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -75,7 +75,7 @@ jobs:
         run: tox -vv --notest -e ${{env.BASE}},${{env.BASE}}-path,${{env.BASE}}-sdist,${{env.BASE}}-wheel
 
       - name: Run test suite via tox
-        run: tox -e ${{env.BASE}} --skip-pkg-install -- --run-integration
+        run: tox -e ${{env.BASE}} --skip-pkg-install
 
       - name: Run test suite via path
         run: tox -e ${{env.BASE}}-path --skip-pkg-install -- --run-integration
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -75,16 +75,16 @@ jobs:
         run: tox -vv --notest -e ${{env.BASE}},${{env.BASE}}-path,${{env.BASE}}-sdist,${{env.BASE}}-wheel
 
       - name: Run test suite via tox
-        run: tox -e ${{env.BASE}} --skip-pkg-install
+        run: tox -e ${{env.BASE}} --skip-pkg-install -- --run-integration
 
       - name: Run test suite via path
-        run: tox -e ${{env.BASE}}-path --skip-pkg-install
+        run: tox -e ${{env.BASE}}-path --skip-pkg-install -- --run-integration
 
       - name: Run test suite via sdist
-        run: tox -e ${{env.BASE}}-sdist --skip-pkg-install
+        run: tox -e ${{env.BASE}}-sdist --skip-pkg-install -- --run-integration
 
       - name: Run test suite via wheel
-        run: tox -e ${{env.BASE}}-wheel --skip-pkg-install
+        run: tox -e ${{env.BASE}}-wheel --skip-pkg-install -- --run-integration
 
       - name: Rename coverage report file
         run: |
--- a/tox.ini
+++ b/tox.ini
@@ -35,7 +35,7 @@ commands =
     pytest -rsx --cov --cov-config setup.cfg \
       --cov-report=html:{envdir}/htmlcov --cov-context=test \
       --cov-report=xml:{toxworkdir}/coverage.{envname}.xml \
-      tests {posargs:--run-integration}
+      tests {posargs}
 
 [testenv:fix]
 description = run static analysis and style checks


gaborbernat (changes requested):
I prefer to run them by default, and people can opt-out via the posargs.

FFY00:
They can't, we specifically did not add that to keep things simple. We have two options, `--run-integration`, that enables the integration tests, and `--only-integration`, that enables the integration tests and disables the rest.

When we discussed moving the integration tests to pytest we agreed that they would be disabled by default, most of the time you don't want to run them. They will still be run on the CI.

pganssle:
> I prefer to run them by default, and people can opt-out via the posargs.

Same here. I have not found them to be particularly slow. When I run `tox` unqualified I want it to run all the tests so I don't have to wait for CI to provision a million environments and whatnot. When I'm testing in a tight loop I use posargs to skip certain tests.

If the tests took an absurdly long amount of time like multiple minutes I might feel differently, but my understanding was that the integration tests were skipped to avoid requiring hitting the internet, which is not a concern 95% of the time.

Perhaps we can compromise and allow flagging this with an environment variable, so users can easily set it for an entire shell session (or perhaps flip the default by including it in `~/.bashrc`)?

gaborbernat:
> When we discussed moving the integration tests to pytest we agreed that they would be disabled by default, most of the time you don't want to run them. They will still be run on the CI.

I think we should reconsider. I'd keep them on always. Remove ``--only-integration`` because is useless. And only run them when ``--run-integration`` is passed, and this is passed by default via posarg inside tox. Users can easily then skip it by doing: ``tox -e py39 -- -v`` (this replaces the posarg of ``--run-integration`` with just ``-v``).

FFY00:
I don't know how long it takes there, but here it takes ~1m15, which compared to the 8s of the test run without them, I think it's an absurd amount of time, but that's just my opinion. To me, 1min is not a reasonable delay in the development loop, at least when it can be avoided.

The integration tests were not designed to be run in the normal suite, that's why they build several different projects in different ways. I think they are something definitely good to keep around as a sanity check, but they are generally redundant with the rest of the tests, and with themselves. There are some exceptions, but they are *very* minor. These tests are not something I want to run all the time. I do not want to add over 1min to the cycle to run redundant tests.

Running the integration tests by default is also, IMO, a bad UX for contributors. They will likely not be affected by them, it will just be something they have to wait on. Sure they can disable them, but this is not clear to new contributors. I find the current approach much better, the tests will be disabled and pytest will report the reason. Users will be notified that the tests were skipped and if want to run them they can easily find out how.

```
====================================================================================================================================== short test summary info =======================================================================================================================================
SKIPPED [2] tests/test_integration.py: integration tests not run (no --run-integration flag)
SKIPPED [60] tests/test_integration.py:70: integration tests not run (no --run-integration flag)
```

Because of this, I am -1 on running them by default.

> Perhaps we can compromise and allow flagging this with an environment variable, so users can easily set it for an entire shell session (or perhaps flip the default by including it in `~/.bashrc`)?

I can accept this as a compromise. I will not be the target user for it, but if it makes your life easier, sure.

> Remove `--only-integration` because is useless.

It isn't.

> Users can easily then skip it by doing: `tox -e py39 -- -v` (this replaces the posarg of `--run-integration` with just `-v`). This way less to type.

This is definitely not intuitive.

gaborbernat:
Intuitiveness is relative. For me what we have is not intuitive either 🤷‍♂️

FFY00:
> Intuitiveness is relative.

It is, but here I am optimizing for new/casual contributors.

> For me what we have is not intuitive either man_shrugging

I am not saying it is, but I think this way is more intuitive for people out of the loop. They get a sane default and get properly notified about the tests that ran and how to adjust that.

> when would you want to run only integrations?

When working on the integrations tests for eg. I added it because I found need for it, but this is not really the point here :smile:

gaborbernat:
Why do we need a flag for it if pytest has built-in support for selective test running (think markers and -k flag)?

FFY00:
Because I want to skip them by default.

And I am talking about the whole UX, not just the flag discoverability, if I was a new contributor and wanted propose a simple patch, I would be a little bit turned off by the over 1min test suite. This is something unnecessary and that we can skip by default. As I said, users are notified that the tests were skipped and the `--run-integration` is pointed out. IMO it is more user friendly than making people wait a bunch of time on unnecessary tests.

This was already discussed previously, you just changed the behavior when you moved to tox.

Unless you want to put your foot down, I will merge this. We can make this configurable with an environment variable as Paul suggested if you find yourself annoyed when using the CLI, this would essentially achieve the same end experience both for me and for you.

gaborbernat:
Do our coverage decrease without the integration test? If not then I'm ok to make integration test optional.

pganssle:
> It is, but here I am optimizing for new/casual contributors.

Whenever I am a casual or new contributor to something, what really bothers me is when I run `tox`, push to CI, and then find out that half the tests I'm expected to make pass don't get run in the `tox` suite. Then I have to dig around looking for ways to make them pass.

Yes, the test suite takes 1 minute to run and the mocked-up code takes 8 seconds, but it's faster than the mock-heavy code partially because the mocked-up code doesn't test much of anything at all! I won't even know if I've broken anything until after I've run the tests!

Also, is there no way to speed up the test suite using `xdist` and/or additional caching? In my tests it takes just as long to run the tests the second time around as it does the first — is that because we're downloading all the projects every time, or because caching the project downloads wouldn't speed anything up?

gaborbernat:
Currently, the coverage goes from 94 to 88 percent without the integration tests. So there's a 6 percent of the code that as @pganssle pointed out might be totally broken, but users who don't run integration tests never find it out.

> In my tests it takes just as long to run the tests the second time around as it does the first

```
10M     .integration-sources/
```

If you have decent internet running the test suite dominates downloading 10MBs.

FFY00:
> Whenever I am a casual or new contributor to something, what really bothers me is when I run `tox`, push to CI, and then find out that half the tests I'm expected to make pass don't get run in the `tox` suite. Then I have to dig around looking for ways to make them pass.

I will reiterate again, the integration tests are just meant to be a sanity check. They should not fail if the normal tests aren't.

> If you have decent internet running the test suite dominates downloading 10MBs.

Yes, and the times I mentioned were with everything cached. I have been ignoring the cache buildup delay.

> because the mocked-up code doesn't test much of anything at all!

That is not true. I agree that the current tests are suboptimal, and that is something that I want to fix. But that is parallel to the integration tests, because they were simply not meant for this. I want to write a really minimal PEP 517 backend just for use in the tests. We should not be relying on the integration tests, the normal test suite should be enough. And the integration tests do not guarantee that anything works other than the process exits successfully, that is just what they were designed to do.

> Currently, the coverage goes from 94 to 88 percent without the integration tests. So there's a 6 percent of the code that as @pganssle pointed out might be totally broken, but users who don't run integration tests never find it out.

The integration tests are not necessarily testing that code. They are not meant to test that code, just because they run it, it doesn't mean that they test it. The code can be totally and utterly broken, that doesn't guarantee that test integration test will fail, they were not meant to do that! This just cooks up the coverage stats and gives you a false sense of security because you think the code is being tested, when it really isn't.

---

Anyway, do whatever. I have better things to do than to argue that we should not be relying on integration tests and that they should be parallel to the normal tests.

layday:
Apropos of nothing but these two lines in the integration checks will always succeed in Python 3 where `filter` returns an iterator:

https://github.com/pypa/build/blob/82bc16c6d43d0881f4e0ad9b98c8ef33e615588a/tests/test_integration.py#L116-L117

FFY00:
> Apropos of nothing but these two lines in the integration checks will always succeed in Python 3 where `filter` returns an iterator:

You are right, please open a new issue.

gaborbernat (approved):

PR #154, opened 2020-10-22
Run test suite in parallel

gaborbernat:
Using the pytest-xdist module to run up to number of user CPUs in
parallel. Fixed some bugs discovered by doing so.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,6 +7,7 @@ import shutil
 import stat
 import subprocess
 import sys
+import sysconfig
 import tempfile
 
 import pytest
@@ -40,6 +41,13 @@ def pytest_addoption(parser):
     parser.addoption('--run-integration', action='store_true', help='run the integration tests')
     parser.addoption('--only-integration', action='store_true', help='only run the integration tests')
 
+    # pytest-xdist does not work on Windows PyPy2 - we need to disable it
+    if platform.python_implementation() == 'PyPy' and sys.version_info[0] == 2 and os.name == 'nt':
+        # the patch here is compatible with pytest 4 + pytest-xdist 1.31 (last release on python2)
+        num_process_option_xdist = next(o for o in parser.getgroup('xdist').options if o.dest == 'numprocesses')
+        # change the type to return None -> no parallel runs
+        num_process_option_xdist._attrs['type'] = lambda s: None  # noqa
+
 
 PYPY3_WIN_VENV_BAD = platform.python_implementation() == 'PyPy' and sys.version_info[0] == 3 and os.name == 'nt'
 PYPY3_WIN_M = 'https://foss.heptapod.net/pypy/pypy/-/issues/3323 and https://foss.heptapod.net/pypy/pypy/-/issues/3321'
@@ -72,6 +80,12 @@ def pytest_collection_modifyitems(config, items):
 
 def is_integration(item):
     return os.path.basename(item.location[0]) == 'test_integration.py'
+
+
+@pytest.fixture(scope='session', autouse=True)
+def ensure_syconfig_vars_created():
+    # the config vars are globally cached and may use get_path, make sure they are created
+    sysconfig.get_config_vars()
 
 
 @pytest.fixture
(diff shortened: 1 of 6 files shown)


gaborbernat:
@FFY00 this is now ready and cuts the CI turnaround time in half.

tests/test_env.py
```diff
@@ -75,6 +76,7 @@ def test_create_isolated_build_has_with_pip(tmp_path, capfd, mocker):
 
 @pytest.mark.skipif(sys.version_info[0] == 2, reason='venv module used on Python 3 only')
 def test_fail_to_get_script_path(mocker):
+    sysconfig.get_config_vars()  # the config vars are globally cached and may use get_path, make sure they are created
```
pganssle, line 79: I don't really like that we need to remember to put this any time we do the `get_path` mocking.

A few options to handle it:

1. Create an autouse session- or module-scoped fixture that does `sysconfig.get_config_vars()` so that the cache is always warmed.
2. Create a fixture for a `get_path_mocker` that warms the cache, like this:

```python
@pytest.fixture
def get_path_mocker(mocker):
    import sysconfig
    # This is a global cache, and may use get_path, so make sure to call it before mocking
    sysconfig.get_config_vars()
    def patch(side_effect=None):
        mocker.patch('sysconfig.get_path', side_effect=side_effect)
    yield patch
```

There are other ways to do this as well. Dealer's choice.

tests/conftest.py
```diff
@@ -39,6 +39,9 @@ def pytest_addoption(parser):
     os.environ['PIP_DISABLE_PIP_VERSION_CHECK'] = '1'  # do not pollute stderr with upgrade advisory
     parser.addoption('--run-integration', action='store_true', help='run the integration tests')
     parser.addoption('--only-integration', action='store_true', help='only run the integration tests')
+    if platform.python_implementation() == 'PyPy' and sys.version_info[0] == 2 and os.name == 'nt':
+        # pytest-xdist does not work on Windows PyPy2, disable it => last compatible: pytest 4 + pytest-xdist 1.31
+        next(o for o in parser.getgroup('xdist').options if o.dest == 'numprocesses')._attrs['type'] = lambda s: None  # noqa
```
pganssle, line 44: There's a lot going on in this one-liner. I recommend breaking it up into smaller chunks.

numprocesses = next(o for o in parser.getgroup('xdist').options if o.dest == 'numprocesses')
numprocesses._attrs['type'] = lambda s: None

tests/conftest.py
```diff
@@ -39,6 +39,9 @@ def pytest_addoption(parser):
     os.environ['PIP_DISABLE_PIP_VERSION_CHECK'] = '1'  # do not pollute stderr with upgrade advisory
     parser.addoption('--run-integration', action='store_true', help='run the integration tests')
     parser.addoption('--only-integration', action='store_true', help='only run the integration tests')
+    if platform.python_implementation() == 'PyPy' and sys.version_info[0] == 2 and os.name == 'nt':
+        # pytest-xdist does not work on Windows PyPy2, disable it => last compatible: pytest 4 + pytest-xdist 1.31
```
pganssle, line 43: Do we require a more modern version of pytest, or can we express this in the test requirements with environment specifiers and/or constraints files?

gaborbernat: Not that I'm aware of because not installing this package will cause pytest to complain about-n not being a valid argument. IMHO this is the cleanest.

tests/test_env.py
gaborbernat: Yeah can make it an auto use session fixture.

tests/conftest.py
pganssle: I meant actually install `pytest-xdist`, just constrain the versions. The comment makes it sound like if we installed `pytest` version 4 and `pytest-xdist` version 1.31, it would work.

gaborbernat: The comment is that the patch works for that version combo, pytest-xdist is just not supported for pypy2 windows 🤷‍♂️

gaborbernat:
@FFY00 ping for review.

pganssle (approved):

pganssle:
@gaborbernat Any idea what the deal is with the coverage numbers?

gaborbernat:
codecov being flaky 🤦

PR #159, opened 2020-10-23
main: add --version/-V

FFY00:

### DIFF
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -97,6 +97,7 @@ def main_parser():  # type: () -> argparse.ArgumentParser
     """
     cwd = os.getcwd()
     out = os.path.join(cwd, 'dist')
+    # mypy does not recognize module.__path__
     # https://github.com/python/mypy/issues/1422
     paths = build.__path__  # type: Iterable[Optional[str]]  # type: ignore
     parser = argparse.ArgumentParser()
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -59,7 +59,7 @@ def _build_in_current_env(builder, outdir, distributions, skip_dependencies=Fals
         builder.build(dist, outdir)
 
 
-def build(srcdir, outdir, distributions, config_settings=None, isolation=True, skip_dependencies=False):
+def build_package(srcdir, outdir, distributions, config_settings=None, isolation=True, skip_dependencies=False):
     # type: (str, str, List[str], Optional[ConfigSettings], bool, bool) -> None
     """
     Runs the build process
@@ -183,7 +183,7 @@ def main(cli_args, prog=None):  # type: (List[str], Optional[str]) -> None
     if not distributions:
         distributions = ['sdist', 'wheel']
 
-    build(args.srcdir, args.outdir, distributions, config_settings, not args.no_isolation, args.skip_dependencies)
+    build_package(args.srcdir, args.outdir, distributions, config_settings, not args.no_isolation, args.skip_dependencies)
 
 
 def entrypoint():  # type: () -> None
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -6,8 +6,9 @@ import sys
 import traceback
 import warnings
 
-from typing import List, Optional, TextIO, Type, Union
+from typing import List, Optional, TextIO, Type, Union, Iterable
 
+import build
 from build import BuildBackendException, BuildException, ConfigSettings, ProjectBuilder
 from build.env import IsolatedEnvBuilder
 
@@ -96,6 +97,8 @@ def main_parser():  # type: () -> argparse.ArgumentParser
     """
     cwd = os.getcwd()
     out = os.path.join(cwd, 'dist')
+    # https://github.com/python/mypy/issues/1422
+    paths = build.__path__  # type: Iterable[Optional[str]]  # type: ignore
     parser = argparse.ArgumentParser()
     parser.add_argument(
         'srcdir',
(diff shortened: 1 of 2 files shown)


src/build/__main__.py
```diff
@@ -8,6 +8,7 @@
 
 from typing import List, Optional, TextIO, Type, Union
 
+import build as _build
```
gaborbernat, line 11: Can you explain why we need this weird alias?

src/build/__main__.py
```diff
@@ -139,7 +146,7 @@ def main_parser():  # type: () -> argparse.ArgumentParser
     return parser
 
 
-def main(cli_args, prog=None):  # type: (List[str], Optional[str]) -> None
+def main(cli_args, prog=None):  # type: (List[str], Optional[str]) -> None  # noqa: C901
```
gaborbernat, line 149: can you put here in brackets the human-readable equivalent of the code/

src/build/__main__.py
```diff
@@ -150,6 +157,10 @@ def main(cli_args, prog=None):  # type: (List[str], Optional[str]) -> None
         parser.prog = prog
     args = parser.parse_args(cli_args)
 
+    if args.version:
```
gaborbernat, line 160: Oh, no need of this cryptic solution, just use the `version` action type from https://docs.python.org/3/library/argparse.html#action

gaborbernat (changes requested):

tests/test_main.py
```diff
@@ -86,6 +86,20 @@ def test_prog():
             build.__main__.main(['--help'], prog='something')
 
 
+def test_version():
+    out = io.StringIO()
```
gaborbernat, line 90: No need for this magic. Just let the version go through and use the capsys fixture from pytest.

src/build/__main__.py
```diff
@@ -150,6 +157,10 @@ def main(cli_args, prog=None):  # type: (List[str], Optional[str]) -> None
...
+        print('build {} ({})'.format(_build.__version__, ', '.join(path for path in _build.__path__ if path)))
```
gaborbernat, line 161: Would it no suffice to just use ``build.__file__`` you can throw an os.path.basename if you must.

src/build/__main__.py
FFY00: We already define `build` in the file.

src/build/__main__.py
FFY00: This makes it more portable and future proof. It can also help debug some weird issues due to namespace colisions.

src/build/__main__.py
```diff
@@ -96,6 +97,8 @@ def main_parser():  # type: () -> argparse.ArgumentParser
     """
     cwd = os.getcwd()
     out = os.path.join(cwd, 'dist')
+    # https://github.com/python/mypy/issues/1422
+    paths = _build.__path__  # type: Iterable[Optional[str]]  # type: ignore
```
gaborbernat, line 101: You'll need to add some details here. What makes ``__path__`` any better compared to ``__file__``? Also, why is this variable declared here if you're using ten lines below, do we really need a variable for it?

src/build/__main__.py
gaborbernat: Perhaps then we shouldn't if we clash with our package name.

src/build/__main__.py
FFY00: `__path__` shows the path of the whole package, not just where `__init__.py` is. We are not a file package, we are a module package.

The variable is there to set the type. mypy does not recognize `module.__path__`, see the issue linked above.

gaborbernat: Isn't the path of the whole package ``os.path.dirname(build.__file__)``? I mean we did not yet split this package over multiple folders, so this feels to me like very premature optimization. One that requires a Github issue link, a type annotation, and slight confusion of how can one package has multiple paths. IMHO accommodating for a potential future where this can be in multiple places is a case of YAGNI.

src/build/__main__.py
```diff
@@ -96,6 +97,8 @@ def main_parser():  # type: () -> argparse.ArgumentParser
     """
     cwd = os.getcwd()
     out = os.path.join(cwd, 'dist')
+    # https://github.com/python/mypy/issues/1422
```
pganssle, line 100:
```suggestion
    # mypy does not recognize module.__path__
    # https://github.com/python/mypy/issues/1422
```

It'd be nice to add a comment explaining what this is. Saves people a click.

src/build/__main__.py
```diff
@@ -105,6 +108,12 @@ def main_parser():  # type: () -> argparse.ArgumentParser
         default=cwd,
         help='source directory (defaults to current directory)',
     )
+    parser.add_argument(
+        '--version',
+        '-V',
+        action='version',
+        version='build {} ({})'.format(build.__version__, ', '.join(path for path in paths if path)),
```
pganssle, line 115: With regards to the `__path__` vs `__file__` thing, I think either is essentially correct, but Filipe's version is I guess slightly more correct?

I'm not sure I'd be able to understand the information I'm being shown if there were multiple paths, but I think I'd probably rather see that something funky is going on than not?

tests/test_main.py
```diff
@@ -86,13 +86,24 @@ def test_prog():
             build.__main__.main(['--help'], prog='something')
 
 
+@pytest.mark.skipif(
+    sys.version_info[0] == 2,
+    reason='capsys not working on Python 2. See https://github.com/pytest-dev/pytest/issues/2327'
+)
+def test_version(capsys):
+    with pytest.raises(SystemExit):
```
gaborbernat, line 94: have you tried with capfd?

gaborbernat (approved):

PR #160, opened 2020-10-23
Remove `srcdir` default value

layday:
As discussed in https://github.com/pypa/build/pull/157 this is not a good default for the programmatic API.


gaborbernat (approved):

FFY00 (approved):
:tada:

PR #161, opened 2020-10-23
Default to `{srcdir}/dist` for `outdir` if option is omitted

layday:
As discussed in #69 and #157, a predictable default is (marginally :) preferable.

### DIFF
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -95,8 +95,6 @@ def main_parser():  # type: () -> argparse.ArgumentParser
     """
     Constructs the main parser
     """
-    cwd = os.getcwd()
-    out = os.path.join(cwd, 'dist')
     # mypy does not recognize module.__path__
     # https://github.com/python/mypy/issues/1422
     paths = build.__path__  # type: Iterable[Optional[str]]  # type: ignore
@@ -105,8 +103,7 @@ def main_parser():  # type: () -> argparse.ArgumentParser
         'srcdir',
         type=str,
         nargs='?',
-        metavar='sourcedir',
-        default=cwd,
+        default=os.getcwd(),
         help='source directory (defaults to current directory)',
     )
     parser.add_argument(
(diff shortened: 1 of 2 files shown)


src/build/__main__.py
```diff
@@ -117,7 +114,13 @@ def main_parser():  # type: () -> argparse.ArgumentParser
         action='store_true',
         help='build a wheel',
     )
-    parser.add_argument('--outdir', '-o', metavar='dist', type=str, default=out, help='output directory')
+    parser.add_argument(
+        '--outdir',
+        '-o',
+        metavar='dir',
```
FFY00, line 120: Can we make this `srcdir/dist`?

FFY00 (changes requested):

src/build/__main__.py
FFY00: Or rather, `os.pathsep.join('srcdir', 'dist')`.

layday: I don't mind either way but it's a little unusual for the [metavar](https://docs.python.org/3/library/argparse.html#metavar) to display the default value for the argument.  This is what the help text looks like right now:

```
 --outdir dir, -o dir  output directory (defaults to {srcdir}/dist)
```

And without `metavar`:

```
  --outdir OUTDIR, -o OUTDIR
                        output directory (defaults to {srcdir}/dist)
```

FFY00: But that information is kind of redundant, isn't it? I usually prefer to put the default value there. And we are doing that in other arguments.

gaborbernat: I like this as is, aka the metavar not containing the default.

gaborbernat (approved):

src/build/__main__.py
layday: > And we are doing that in other arguments.

Not that I can see - there are only two labels that appear in the help text - one for `--outdir` and the other for `--config-setting`: `CONFIG_SETTING` - and I feel like it comes at the cost of clarity, that the default should appear in the label's position.

layday: (Arguably `CONFIG_SETTING` should be something like `opt[=val]` instead if we're pursuing descriptive labels.)

FFY00:
Please don't use merge commits, rebase on top of master instead.

```
git reset --hard HEAD~1  # to remove the merge commit
git rebase master
```

layday:
Rebased.

src/build/__main__.py
layday: How do you want to proceed here?

FFY00: I've stated my preference. Since both Bernát and you have the same opinion, I guess go with that. But please fix all other arguments to match this.

layday: Does fixing mean lowercasing?  And that'd only be for `--config-setting`, right?

FFY00 (approved):

PR #162, opened 2020-10-24
setup.cfg: add package description

FFY00:

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,7 @@
 [metadata]
 name = build
 version = 0.0.4
+description = A simple, correct PEP517 package builder
 long_description = file: README.md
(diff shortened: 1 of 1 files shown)


gaborbernat (approved):

setup.cfg
```diff
@@ -1,6 +1,7 @@
 [metadata]
 name = build
 version = 0.0.4
+description = A simple, correct PEP517 package builder
```
gaborbernat, line 4: I'm not sure either the simple or correct has any merit here, though. I mean, we obviously will try to make it work right and keep it as simple as possible. I feel these words could/should be universally true for any tool and, as such, don't add much value.

FFY00: We can open an issue to discuss this.

PR #167, opened 2020-10-26
Dep versions

FFY00:


setup.cfg
```diff
@@ -43,15 +43,15 @@ console_scripts =
...
+    sphinx-autodoc-typehints~=1.10
+    sphinxcontrib-autoprogram~=0.1.0
 test =
-    filelock
-    pytest
-    pytest-cov
-    pytest-mock
-    pytest-xdist>=1.34
+    filelock~=3.0
+    pytest>=5,<=6
```
gaborbernat, line 51: For python 2 the latest version is 2, for python 3 6. Any reason we're not pulling in the latest for all above 4?

setup.cfg
```diff
@@ -43,15 +43,15 @@ console_scripts =
...
+    pytest-cov~=2.0
+    pytest-mock>=2,<=3
```
gaborbernat, line 53: Any reason you need to upper pin here? I'd rather we upper pin when the new version breaks us. More often tends to not, see https://www.bernat.tech/version-numbers/ and https://snarky.ca/why-i-dont-like-semver/

gaborbernat (changes requested):

setup.cfg
FFY00: Hum, I have my reservations but sure.

setup.cfg
```diff
@@ -43,14 +43,14 @@ console_scripts =
 [options.extras_require]
 doc =
     furo
-    sphinx
-    sphinx-autodoc-typehints
-    sphinxcontrib-autoprogram
+    sphinx~=3.0
```
gaborbernat, line 46: I guess the same here too and the following, any reason we need compatibility and not just take whatever is latest?

FFY00: I find that is harder to detect issues in sphinx as it sometimes just ends up wrongly formatting the docs, without raising an error or warning. I'd prefer updating to new major sphinx versions manually, but if you have a strong opinion on this I can change.

setup.cfg
```diff
@@ -43,14 +43,14 @@ console_scripts =
...
+    sphinx-autodoc-typehints~=1.10
+    sphinxcontrib-autoprogram~=0.1.0
 test =
-    filelock
-    pytest
-    pytest-cov
-    pytest-mock
+    filelock~=3.0
```
gaborbernat, line 50: Ok then, let's make this to ``>= 3``. We can keep the docs part to ``~=``.

gaborbernat (changes requested):

setup.cfg
FFY00: Okay.

FFY00:
How do you think we should do `install_requires`?

gaborbernat:
> How do you think we should do `install_requires`?

Can you detail what you mean?

FFY00:
Should we add requirements to any of the dependencies there? The usage of the unversioned dependencies should not depend on any newly introced API, but there have been quite a few bugfixes in code that we use. I don't have time to go find out the very first version that *should* work for us, or the bugs that might have affected us. Should we pick a recent enough version and add a requirement on it or keep unversioned?

gaborbernat:
I think everything should be versioned to be at least with the version we tried (are running with). Now if we want to relax it, to basically not require the latest of everything we can do that in a follow-up PR. We can there introduce a constraint file, where we can use some binary search to find out their minimum version... And add for targets to our test matrix using this (one windows, one linux, only python 2, one python 3).

FFY00:
Sure, then we'll defer for later.

setup.cfg
gaborbernat: This change hasn't been addressed yet.

FFY00: Sorry, forgot to push :upside_down_face:

gaborbernat (approved):

gaborbernat (approved):

PR #168, opened 2020-10-26
release 0.1.0

FFY00:
- Moved the upstream to PyPA
- Fixed building with isolation in a virtual environment
- Renamed the entrypoint script to pyproject-build (!)
- Removed default arguments from all paths in ProjectBuilder (!)
- Removed ProjectBuilder.hook (!)
- Renamed __main__.build to __main__.build_package (!)
- Changed the default outdir value to {srcdir}/dest (!)
- Removed env.IsolatedEnvironment (!)
- Added env.IsolatedEnv abstract class
- Added env.IsolatedEnvBuilder (replaces env.IsolatedEnvironment usages)
- Added python_executable argument to the ProjectBuilder constructor
- Added --version/-V option to the CLI
- Added support for Python 3.9
- Added py.typed marker
- Various miscelaneous fixes in the virtual environment creation
- Many general improvements in the documentation
- Documentation moved to the furo theme
- Updated the CoC to the PSF CoC, which PyPA has adopted

(!) - breaking change

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,6 @@
 [metadata]
 name = build
-version = 0.0.4
+version = 0.1.0
 description = A simple, correct PEP517 package builder
 long_description = file: README.md
 long_description_content_type = text/markdown
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -3,7 +3,7 @@
 """
 build - A simple, correct PEP517 package builder
 """
-__version__ = '0.0.4'
+__version__ = '0.1.0'
 
 import contextlib
 import difflib


gaborbernat (commented):
Do we have some kind of chanelog we need to update or none yet?

gaborbernat (approved):

FFY00:
I will write it up and add it to the commit description. Preferably this should be done in the git tag, but Github shows the commit description instead of the tag description in the releases tab.

pganssle:
@FFY00 Can we add a `NEWS` or `CHANGELOG` or `NEWS.rst` or something of that nature instead of keeping it in the commit message?

The commit messages are very inaccessible to most end users. The changelog should also be included in the docs.

We really should have `towncrier` or something similar set up, because otherwise it becomes a real pain to track down all the changes in the release and tie them to relevant PRs and issues.  Much easier to do it as you go along. Of course, I don't plan on preparing or making releases with the current workflow, so I suppose it doesn't make much difference to me how the changelog is managed.

FFY00:
Sure, if someone sets it up.

gaborbernat:
Let's not burden the release with this now, I've created https://github.com/pypa/build/issues/169 to address the underlying issue. From my POV is fine we do it for the next release.

pganssle:
@gaborbernat You mean we shouldn't block on a changelog for 0.1.0? Or we shouldn't block on setting up towncrier?

I agree we shouldn't block on towncrier, since none of its value is retrospective. I disagree that we should just make the 0.1.0 release with no changelog. From what I can tell, a good fraction of our users are "power users" who are likely to find bugs and then work hard to track them down to their source. Changelogs are an invaluable aid in tracking down regressions or pointing to likely sources of new issues. We'd be doing our users a huge disservice by asking them to comb through the repo to find the changes associated with a given release.

gaborbernat:
> You mean we shouldn't block on a changelog for 0.1.0

Yes, on both, IMHO. But this in particular.

FFY00:
Added a changelog, please let me know if you have any issues with it.

pganssle:
> Added a changelog, please let me know if you have any issues with it.

It should be in a file in the repository called something like `(NEWS|CHANGELOG)(.(rst|md))`.

Ideally it would also link to the PRs and issues involved (which is why I suggested adding `towncrier` back in July, since it's way easier to track these things when you do it at the time).

Ideally the changes file would also be included in the docs somewhere (which argues for using `.rst`), but I'd say that's not critical for the release.

FFY00:
Let's defer that to when we setup a news system. This changelog will be included in the release commit as well as the tag. When we setup a news system we can get the changelogs from the previous versions and put them on the news file.

pganssle:
> Let's defer that to when we setup a news system. This changelog will be included in the release commit as well as the tag. When we setup a news system we can get the changelogs from the previous versions and put them on the news file.

Ugh. That's not how any of the existing news systems work. You still need to do all the same work, just now it automatically appends to the changelog file.

I realize that you want to get the release out the door, but the only reason I didn't insist that we set up town crier months ago was because you said that you "know what the changes are" and would write up the changelog. I don't think we should make our users jump through hoops because the technical debt that you deliberately accrued makes the release process difficult for you.

FFY00:
The current development process is not set up in that way and it is partly my fault. If it was my call, all commits would be properly written with descriptive descriptions (they don't need to be long, ) where it makes sense and have references to the relevant issues. 

https://lists.freedesktop.org/archives/input-tools/2020-September/001542.html

layday:
Can we not pull the commit messages from previous releases into a simple `CHANGELOG.rst` for 0.1.0?  We can reformat it when we do adopt towncrier/equivalent.

FFY00:
The current development process is not set up in that way and it is partly my fault. If it was my call, all commits would be properly written with descriptive descriptions (they don't need to be long, just give the general background). I don't understand why we are coming up with this increasingly more complicated way to save metadata where we have a perfectly capable one that we are using anyway... Why can't we just have a readable git history? We should be able to just write a few highlights for the relevant changes and point at the git history for the full changelog. Instead of depending on issue and pull request links, why don't we just put the end result of the discussion in the commit? The discussions we have are full of endless bikeshedding and not useful for users to read, I actually think they are a disservice to most. Unless they are looking for the exact discussion, to the full extent, that led to the decisions made, they will have a *very* hard time getting something out of them. Why do I must add news entry saying "fixed X (issue #999)" instead of having a commit say "fixed X" and inclusing a summary of the problem and the chosed fix with a link to the github issue.

I have been doing this in a few projects with an active userbase. And so have many other projects.

https://lists.freedesktop.org/archives/input-tools/2020-September/001542.html
https://lists.freedesktop.org/archives/input-tools/2020-June/001540.html
https://github.com/pwr-Solaar/Solaar/releases

Here are a few examples of commits that required such background to be laid out.

https://github.com/libratbag/libratbag/commit/9fe511fcc5de36e11d5b2843606d331f829b15cd
https://github.com/libratbag/libratbag/commit/20b6e3deeb43cfe0838dee8db1c313b7856373b0
https://github.com/libratbag/libratbag/commit/8f88573a086ce8c20f150025d053441ee0351868
https://github.com/libratbag/libratbag/commit/b31a8c5355b6cb5b821c3343d4e7d312c4887b5f
https://github.com/libratbag/libratbag/commit/c1265bc98af90fb1cbb0ad02153b888c009d3c43

If you look at the git history for this release, my commits at least are pretty readable and easy to read and follow, even though I admitedly have not been doing an amazing job.

https://github.com/pypa/build/compare/0.0.4...master

From my experience, this metodology has been pretty sucessfull both for users and contributors. There is no need for a detailed news file as users can simply read the git history, which includes the required background, if needed, in the commits' description. A general overview of the changes, just like I did here, completes the git history pretty well.

I think your proposed approch is much worse in

It also has a very very big plus for me, which is that the repo does not become mostly useless if we lose access to the issue tracker for some reason.

So, I don't appreciate being solely blamed on deliberately introducing technical debt. I understand that I may be difficult to work with due to our different approaches, but having a different approach from what you consider optimal does not mean that I just do not care and want to make users' lives worse. This project was in fact using the current scheme from the beggining, I relaxed the git history a bit when that became an issue. Aditionally to that I have compromised in a lot of places, like the pressured adoption of black, with some still being actively discussed, like my hard position on providing locally signed release artifacts, to the point I do not feel good contributing to my own project anymore. Don't get me wrong, I am open to compromise and try to acomodate everyone's workflows, but I do not appreciate this kind of suggestions that I do not care about the users. I also want to note that I never blocked the addition of a news system, I just said that I wouldn't do it myself as I had (and still have) many other things to do, even though I do not think it is the best approach.

FFY00:
I will maybe come back to this this week if I am in a mindset for it.

pganssle:
Feel free to run the project however you like.

I will find it incredibly annoying that I need to dig around the git history for a changelog. You will get requests from users for a changelog. You will not be able to get meaningful commit messages from contributors, and the changelogs you linked are nearly useless to me.

I hope I am wrong, but I doubt I will be.

gaborbernat (approved):
IMHO we can move ahead and do the release now, with the changelog as proposed. People are hurt more by not having a release than having a release without a changelog.

Now as far as the changelog maintenance goes, I'm really not against the raw source being in git. However, if we go down that path we should add some checks to ensure the structure is upheld. We can do this via some Github Action check. For contributors that don't adhere to it, we can probably work around by squashing their commits and reformatting the squash message accordingly at merge time. One thing we need to accommodate for is commits/PRS that is not user-facing and as such, they don't need a changelog entry... so some commits do need to excluded from the changelog.

PR #172, opened 2020-10-27
Use setuptools-scm to generate version from SCM (fallback to static)

gaborbernat:
👍 Compared to what we have now, we're a bit ahead in the sense of post this we'll only have one version number within our repo, and the version number will be automatically generated from the SCM (if present). We'll need to make sure to update the version file before we do a release, but we're already here today.

Closes https://github.com/pypa/build/issues/170


FFY00 (changes requested):
I have said I do not want to pull any other bootstraping dependencies.

gaborbernat:
> I have said I do not want to pull any other bootstraping dependencies.

* ``bootstrapping``

Look man, this is not a hard dependency, it's a soft one. The package works just fine (as is today) without it. But if installed provide additional functionality today not present.

pganssle:
Yeah, it has an additional dependency in the `pyproject.toml`, but anyone who cares about bootstrapping isn't blindly pulling in the `pyproject.toml` dependencies anyway.

For building from `sdist`, I think `setuptools_scm` isn't even really used, since there's no git for it to even interact with (though there's no way to indicate this in `pyproject.toml` without modifying the file in place as part of the sdist build).

layday:
I assume the tests are failing because we need a `MANIFEST.in` for `version.fallback` for when the `setuptools_scm` file finder is not in use (i.e. without isolation). `setuptools_scm` will pull everything that's tracked by the SCM into the sdist automatically.  We might also want to customise `MANIFEST.in` to exclude files which wouldn't be included by vanilla `setuptools` if sdist parity is important.

FFY00 (approved):
It is still annoying for packagers, but sure.

Please properly document that `setuptools-scm` is not actually required for the build, and the project can be bootstrapped without it.

Other than that, do whatever.

gaborbernat:
> It is still annoying for packagers, but sure.

Why? Can you expand on this a bit? 😊

FFY00:
> Why? Can you expand on this a bit?

Not all packagers will know this, it takes a bit of digging to find out about this.

We should definitely document this.

src/build/__init__.py
```diff
@@ -3,20 +3,20 @@
...
 import sys
 import warnings
-
 from typing import Iterator, Mapping, Optional, Sequence, Set, Text, Union
 
 import pep517.wrappers
 import toml
 import toml.decoder
 
+from .version import __version__
```
FFY00, line 18: Did you forget to commit `version.py`?

FFY00: nvm

webknjaz (approved):

gaborbernat:
Will create new PR.

PR #175, opened 2020-10-31
docs: build on RTD using Python 3.8 

### LINKED ISSUE #118: Missing types in sphinx docs
AFAIK sphinx-autodoc-typehints is being used correctly, not sure why it isn't showing.

@pradyunsg any ideas?

layday:
autodoc-typehints is not able to display type hints under
Python 3.7 and earlier without the `typed_comments` extra.

Closes #118.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,6 +1,7 @@
 version: 2
 
 python:
+  version: 3.8
   install:
     - requirements: docs/requirements.txt
     - method: pip


FFY00 (approved):
LGTM :+1: 

Could you prefix the commit with `docs: ` so that it is easier to identify this is related to the docs?

layday:
Yep, done.

gaborbernat (changes requested):
> autodoc supports Python 2-compatible `type:` annotations natively
> which sphinx-autodoc-typehints appears not to.

That's not true I have projects that use it.

Also looking at the preview render I see no type information for elements, see https://pypa-build--175.org.readthedocs.build/en/175/api.html#build.ProjectBuilder.build_dependencies

layday:
Yes, sphinx-autodoc-typehints is not able to display the types. Have you tried running make html locally with this change? The type annotations should appear in the docs.

layday:
Something’s not quite right with the readthedocs build. It says the build version is 0.0.4? Why would that be?

gaborbernat:
> Yes, sphinx-autodoc-typehints is not able to display the types.

I'm saying it is. I've used it for it and was doing a stellar job.

> Have you tried running make html locally with this change?

Not sure what make HTML does, we really should remove the make file. Use the tox invocation via ``tox -e docs`` to replicate more to what Readthedocs does. The Readthedocs link I've linked in my previous post contains no type information, and that's what matters.

gaborbernat:
> Something’s not quite right with the readthedocs build. It says the build version is 0.0.4? Why would that be?

Because the version is taken from the source code, and you did not rebase against the master to pull in the new version release.

layday:
ah, I was sure I did.  I've run `tox -e docs` locally with and without this change and the type hints appear in the documentation (although in a different fashion).  I can probably drop `sphinx-autodoc-typehints` anyway once we've figured out why they're not showing on Read the Docs.

gaborbernat:
> We can probably drop `sphinx-autodoc-typehints` anyway once we've figured out why they're not showing on Read the Docs.

I'm personally not sold on this yet. I don't like how the type annotations are injected in-line on the signature... rather than attached to the arguments (with `sphinx-autodoc-typehints`).

layday:
There is an option to change that in autodoc:

```
autodoc_typehints = 'description'
```

layday:
I was able to repro this issue by building the docs on Python 3.7 instead of 3.8.  RTD uses 3.7.

layday:
I assume this has something to do with how forward references are evaluated in Python 3.7 (and earlier) by autodoc.  Bumping the RTD Python to 3.8 appears to have fixed it.

Rebased.

FFY00:
Then we can drop the first commit, as we will keep using sphinx-autodoc-typehints.

gaborbernat (approved):

Issue #132, opened 2020-10-09
We are relying on `get_requires_for_build_{sdist,wheel}` to be present


FFY00:
In light of #130 I noticed we are wrongly relying on `get_requires_for_build_{sdist,wheel}`, optional hooks, to be present.

/me blames bad assumptions due to him being super distracted

layday:
pep517 does not raise if these two hooks are undefined, it simply returns an empty list, so we shouldn't have to do anything special here.

FFY00:
Ah, missed that when I revisited this. This behavior is controlled by `_allow_fallback`, which defaults to `True`, so we should be good.

layday:
I believe `_allow_fallback` is only available for `prepare_metadata_for_build_wheel` - pep517 always falls back on `[]` for `get_requires_for_build_{sdist,wheel}`

gaborbernat:
Ideally, we'd close this by adding a test to validate this assumption and make sure new versions of pep517 don't break it.

PR #176, opened 2020-10-31
Verify `get_dependencies` returns empty set when its hooks are undefined

layday:
We're testing this against a local backend which does not expose
any hooks.

See #132.


FFY00 (approved):
LGTM :+1: 

Can you prefix the commit with `tests: `?

gaborbernat (approved):

PR #178, opened 2020-11-01
docs: beef up the `ProjectBuilder`'s method docstrings

layday:


gaborbernat (approved):

src/build/__init__.py
```diff
@@ -169,6 +169,9 @@ def __init__(self, srcdir, config_settings=None, python_executable=sys.executabl
 
     @property
     def python_executable(self):  # type: () -> Union[bytes, Text]
+        """
+        The Python executable used by :mod:`pep517` to invoke the backend.
```
FFY00, line 173: Using `pep517` is an implementation detail, please remove the reference.

FFY00 (changes requested):

src/build/__init__.py
layday: Removed :+1:

gaborbernat: I considered he made the reference here as pep 517 the API, not the package. And in that sense the sentence is right and does not exposes implementation details.

layday: It was meant for the package.  I'm not sure what the 'executable used by the PEP 517 API' would mean in this context.

gaborbernat: Well the python executable hosting the build backend (as opposed the build frontend - current process), https://www.python.org/dev/peps/pep-0517/#build-backend-interface, what else could it?

FFY00: Yeah, `:mod:` references a module.

FFY00 (approved):

src/build/__init__.py
layday: @gaborbernat I mean that it'd have been confusing to introduce another concept (the 'API') in addition to the frontend and backend - it'd have been somewhat of a tautology.  If we wanted to throw PEP 517 in there, I'd have rephrased it as:

> The Python executable used to invoke the PEP 517 backend.

But I think it goes without saying that the backend conforms to PEP 517 - build is a PEP 517 frontend.

PR #179, opened 2020-11-01
Reorganise docs

layday:


setup.cfg
```diff
@@ -41,9 +41,9 @@ console_scripts =
     pyproject-build = build.__main__:entrypoint
 
 [options.extras_require]
-doc =
+docs =
```
FFY00, line 44: Why?

FFY00 (approved):
It's fine, although I'd prefer to have the changes in separate commits. If you don't want to do that, don't let it be a blocker.

gaborbernat (approved):

setup.cfg
layday: I suppose because we speak of "docs" in the commit messages and the folder itself is called "docs".

layday: (I won't split the commits up if that's ok but I'll keep it in mind for future PRs.)

FFY00: Yeah, that's fine.

PR #183, opened 2020-11-08
Enhance `check_version`

layday:


src/build/__init__.py
```diff
@@ -55,43 +55,47 @@ class IncompleteCheckWarning(Warning):
...
     else:
         import importlib_metadata
 
-    req = packaging.requirements.Requirement(requirement_string)
-    env = {'extra': extra}
+    req = Requirement(req_string)
 
-    if req.marker and not req.marker.evaluate(env):
-        return True
+    if req.marker:
```
gaborbernat, line 75: Could we perhaps use somehow ``pip check`` instead?

layday: Well, that would place a runtime dependency on pip (we could use `ensurepip` I suppose but distros might then be inclined to patch `build` to 'unvendor' pip), and unless we want to be calling its internal API we're looking at a ~400ms overhead per check purely in import time.  IUC pip does not keep track of extras so `pip check` would not error for e.g. `setuptools_scm[toml]` if the toml extra dependencies weren't available.

gaborbernat: We already have pip as a runtime dependency - https://github.com/pypa/build/blob/master/src/build/env.py#L173-L188; and this would allow us to drop packaging as install dependency.

layday: Even then, we'd have to combine `check` with `list` or `show` to verify that the build dependencies are installed and `list` is the only one of those commands which produces machine-readable data.  I don't know how we'd verify that extras are satisfied with pip.

FFY00: The current runtime dependency on pip is for the environments, and it's not even required if ensurepip works. This adds a dependency on pip for `--no-isolation` which we most definitely do not want, that was the whole point of this routine.

layday:
I've made a few changes to return the dependency chain similar to how `pip check` does it.

src/build/__init__.py
```diff
@@ -49,49 +49,47 @@ class TypoWarning(Warning):
...
-def check_version(requirement_string, extra=''):  # type: (str, str) -> bool
+    :param req_string: Requirement string
+    :param parent_extras: Extras (eg. "test" in myproject[test])
+    :yields: Unmet dependencies
     """
-    :param requirement_string: Requirement string
-    :param extra: Extra (eg. test in myproject[test])
-    """
-    import packaging.requirements
+    from packaging.requirements import Requirement
```
FFY00, line 61: Please keep the original import, instead of using a partial import.

src/build/__main__.py
```diff
@@ -55,7 +55,7 @@ def _build_in_current_env(builder, outdir, distributions, skip_dependencies=Fals
         if not skip_dependencies:
             missing = builder.check_dependencies(dist)
             if missing:
-                _error('Missing dependencies:' + ''.join(['\n\t' + dep for dep in missing]))
+                _error('Unmet dependencies:' + ''.join(['\n\t' + ' -> '.join(unmet_deps) for unmet_deps in missing]))
```
FFY00, line 58: Why change the message string here ("missing" -> "unmet")?

I am not really a fan of the way this is displayed in the CLI:
```
ERROR Unmet dependencies:
	apispec[validation] -> prance[osv] (>=0.11) ; extra == 'validation' -> openapi-spec-validator (>0.2,>=0.2.1) ; extra == 'osv'
```

I think the following would be better:
```
ERROR Unmet dependencies:
	apispec[validation]
        prance[osv] (>=0.11) -> openapi-spec-validator (>0.2,>=0.2.1)
```

src/build/__init__.py
```diff
@@ -49,49 +49,47 @@ class TypoWarning(Warning):
     """
 
 
-class IncompleteCheckWarning(Warning):
+def check_dependency(req_string, ancestral_req_strings=(), parent_extras=frozenset()):
```
FFY00, line 52: I think it would be better to put effort in upstreaming this into pypa/packaging first, as proposed in pypa/packaging#317.

tests/test_projectbuilder.py
```diff
@@ -60,35 +59,35 @@ def from_name(cls, name):
...
+        ('something[some_extra] == 2.0.0', ('something[some_extra] == 2.0.0',)),
     ],
 )
-def test_check_version(requirement_string, expected, extra_warning):
-    if extra_warning:
-        with pytest.warns(build.IncompleteCheckWarning):
-            assert build.check_version(requirement_string) == expected
-    else:
-        assert build.check_version(requirement_string) == expected
+def test_check_dependency(requirement_string, expected):
```
FFY00, line 86: If we are gonna merge this here, we need to also test recursive extra dependencies.

tests/test_projectbuilder.py
```diff
@@ -165,30 +164,27 @@ def test_build_missing_backend(packages_path, distribution, tmpdir):
...
     mocker.patch('pep517.wrappers.Pep517HookCaller.get_requires_for_build_sdist')
     mocker.patch('pep517.wrappers.Pep517HookCaller.get_requires_for_build_wheel')
-    mocker.patch('build.check_version', autospec=True)
 
     builder = build.ProjectBuilder(test_flit_path)
 
     side_effects = [
         [],
-        ['something'],
+        ['foo'],
```
FFY00, line 174: Why? This is unrelated and just dirties git blame.

FFY00 (changes requested):

tests/test_projectbuilder.py
layday: Because `importlib_metadata.Distribution` is mocked further up and `something` is would return a `Distribution` instance, i.e. the dependency would be satisfied.

src/build/__init__.py
layday: The impression I got was that they were reluctant to add anything which manipulates distributions.  `distlib` would probably be a good fit but everybody seems to be moving away from it.

src/build/__main__.py
layday: > Why change the message string here ("missing" -> "unmet")?

'Missing' might lead one to think that the dependency can't be located when its version is incompatible.

> I think the following would be better:

It's not clear that the unmet dependency here is `openapi-spec-validator`.  Maybe something like this would be better?

```
ERROR Unmet dependencies:
	openapi-spec-validator (>0.2,>=0.2.1) ; extra == 'osv'
		from prance[osv] (>=0.11) ; extra == 'validation'
				from apispec[validation]
```

src/build/__init__.py
gaborbernat: We talked about this earlier and for what it's worth, I'm still strongly -1 on original imports and +1 on partial imports. However 🤷

src/build/__init__.py
gaborbernat: In my experience there's a long cycle with adding stuff to distlib... so if we want it soon we better add it here. We can always remove it if we do manage to upstream it.

src/build/__init__.py
gaborbernat: Seemed a worthwhile topic to bring up. Considering pip already has a checking functionality, perhaps we don't want to reinvent the wheel.

src/build/__init__.py
layday: Restored.

tests/test_projectbuilder.py
layday: Added.

src/build/__main__.py
FFY00: > 'Missing' might lead one to think that the dependency can't be located when its version is incompatible.

The requirement string would include the version requirement, so I find that very unlikely. I do not think this is reason enough to change.

> It's not clear that the unmet dependency here is `openapi-spec-validator`. Maybe something like this would be better?

Well, it is certainly not optimal but I think it's clear enough. Your proposed way does not work great from long chains, I see two issues with it: 1) it puts all dependencies on a new line, which works very badly on >5 chains, and 2) it puts the actual requirement in the end, which is hard to find. It also has the `; extra = ...` which is completely avoidable.

As a user, what I am most interested in knowing is the 1st level missing dependency, in this case `apispec[validation]`, so that I can `pip install apispec[validation]`. As a distro packager, I want to know the actual dependency in the tree, and that is clear enough to me in the format I proposed. We should optimize for the general user.

So, for the interface here I have the following requirements:
  - make it clear, and easily accessible, what is 1st level missing dependency (in this case `apispec[validation]`)
  - provide information about the missing dependency chain (but do not optimize for this)
  - make sure the output does not become unreadable on long chains

src/build/__init__.py
FFY00: I am in no hurry to merge this as it only affects people using `--no-isolation` when trying to verify recursive extras (extras that depend on extras), which is mostly distro packagers, for whom this is only a nice to have. I would rather put the effort into merging somewhere this could live long term rather than here. Before merging it here, we should figure out what would it take to put it somewhere else and then make the call.

gaborbernat: > I am in no hurry to merge this as it only affects people using `--no-isolation`

As a (new) contributor to this project, this sentence on its own would be enough for me to abandon spending any more effort on this PR.

src/build/__main__.py
layday: I've changed the formatting as suggested.

tests/test_projectbuilder.py
layday: I've removed the assignment to `importlib_metadata` from the global scope and reverted this change.

src/build/__init__.py
layday: > as it only affects people ... trying to verify recursive extras

`check_version` would not verify transitive dependencies at all, whether or not they came from extras.

FFY00: Yes sorry, my bad.

FFY00 (approved):
LGTM, let's wait a few days to see if we can hear something from pypa/packaging#317.

layday:
While we are waiting, would anybody know why the pypy3 tests are failing?

gaborbernat:
> While we are waiting, would anybody know why the pypy3 tests are failing?

No, cold you investigate?

layday:
It looks like it's flaky - it passed on Ubuntu this time.  Appears to be coverage-related, possibly https://github.com/nedbat/coveragepy/issues/1010, judging from this line:

```
E                   _sqlite3.InterfaceError: Error binding parameter 0 - probably unsupported type.
```

I don't know why it'd appear now, there might be some interaction with the `monkeypatch` fixture which I replaced the global overwrite of `importlib_metadata.Distribution` with?

layday:
I dropped the commit which enabled branch coverage and that seems to have fixed it (or perhaps not).  We can add branch cov in a separate commit.

layday:
Should we merge this now?

gaborbernat (approved):

FFY00:
I specifically asked to wait, please do merge things like this without getting the ack from everybody that raised concerns.

layday:
Well, I merely asked if we should merge it - that is, if it would've been time to merge it.  It wasn't an invitation to merge it - sorry if I'd not made myself understood.

FFY00:
The message was directed to Bernát, not you.

gaborbernat:
Sorry, thought we moved past that.

PR #185, opened 2020-11-08
Fix isort config

layday:


layday:
I'm actually not sure if `/build` and `/dist` are interpreted relative to the project folder.  Might be better to just take them out of the list.

FFY00 (approved):

PR #189, opened 2020-11-14
build: add exports to __init__.py for strict type checking support

layday:
With `mypy --strict` (or `pyright: strict`, etc.) you cannot
import symbols which are not explicitly (re-)exported.
This can be accomplished in one of two ways: using `__all__`
or with `import x as x` if a symbol is imported from another module.

### DIFF
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -241,3 +241,14 @@ class ProjectBuilder(object):
             raise BuildException("Backend '{}' is not available.".format(self._backend))
         except Exception as e:  # noqa: E722
             raise BuildBackendException('Backend operation failed: {!r}'.format(e))
+
+
+__all__ = (
+    '__version__',
+    'ConfigSettings',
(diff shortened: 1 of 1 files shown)


FFY00 (changes requested):
Sorry for the late review!

You can add `check_dependency` and `ConfigSettings`.

layday:
I wasn't sure if `check_dependency` should be public, especially if we're looking to eventually import it or its equivalent from `packaging`. What's your opinion?

FFY00:
Well, it is already public. And we will need it for the wheel installer. I'd really like to see this going to someplace else, but that doesn't seem like it's gonna happen anytime soon.

From a bootstraping perspective, I would much rather have it here than in another custom package, as I would need `build` anyway. So, I say we temporarily provide it here. When it gets adopted somewhere else, we can import it from there and add a deprecation warning.

FFY00 (approved):

Issue #188, opened 2020-11-13
Verbosity flag


uranusjr:
`build` currently passes all output from the backend, but some of them are very cluttered by default. It would be nice to have a quiet mode that only outputs what backend is being used and what hooks are being called, and supress outputs from the backend if the hook call succeeds.

FFY00:
Related to #142.

> only outputs what backend is being used

Can you elaborate?

uranusjr:
Show the value in `build-system` so the user knows what is actually being run (and can more easily debug typos), something like

```
Using backend: setuptools.build_meta
```

PR #192, opened 2020-11-26
Curate output

### LINKED ISSUE #142: Curate output
Right now we do not care much about the output. I wanted to pay a bit more attention to that.

I propose to keep the output limited to build backend output, this means swallowing the ensurepip output, and to curated messages from us (`build`). Our own messages should be clear that come from us, either by having an identifiable prefix or a specific style. Our messages should also be pretty limited, keeping the output clean. If we see need for debug messages we can offer that under a `--debug` flag.

For our messages, I propose adding a `build.print_info` method that could be overwritten, in `warnings.showwarning` fashion. The default implementation would use the `logging` module, but we would overwrite it in `__main__` to use a colorized raw print.

FFY00:
Fixes #142

This only implements the changes suggested in #142 to curate the output. It does not implement debug messages, as I said there, if we see the need, we can do it in the future. Things like the output of the environment creation, like pip, ensurepip, venv or virtualenv that are swallowed could be outputed in the future with a `-v` flag or something similar.

### DIFF
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -42,6 +42,19 @@ def _error(msg, code=1):  # type: (str, int) -> None  # pragma: no cover
     exit(code)
 
 
+def _print_info(msg):  # type: (str) -> None
+    """
+    Custom build.print_info implementation to print directly to stdout, and with
+    colors.
+    """
+    prefix = '*'
+    end = ''
+    if sys.stdout.isatty():
+        prefix = '\33[1m\33[36m' + prefix + '\33[39m'
+        end = '\33[0m'
+    print('{} {}{}'.format(prefix, msg, end))
+
+
 def _format_dep_chain(dep_chain):  # type: (Sequence[str]) -> str
     return ' -> '.join(dep.partition(';')[0].strip() for dep in dep_chain)
 
@@ -167,6 +180,8 @@ def main(cli_args, prog=None):  # type: (List[str], Optional[str]) -> None
 
     :param cli_args: CLI arguments
     """
+    build.print_info = _print_info
+
     parser = main_parser()
     if prog:
         parser.prog = prog
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -11,7 +11,7 @@ import sysconfig
 import tempfile
 
 from types import TracebackType
-from typing import Iterable, Optional, Tuple, Type
+from typing import IO, Iterable, Optional, TextIO, Tuple, Type, Union
 
 from ._compat import abstractproperty, add_metaclass
 
@@ -42,9 +42,20 @@ class IsolatedEnv(object):
 
 
 class IsolatedEnvBuilder(object):
-    def __init__(self):  # type: () -> None
-        """Builder object for isolated environment."""
+    def __init__(self, silent=True):  # type: (bool) -> None
+        """
+        Builder object for isolated environment
+
+        :param silent: Silence the output
+        """
         self._path = None  # type: Optional[str]
+        self._silent = silent
+        self._outfd = sys.stdout  # type: Union[int, TextIO, IO[str]]
+        if silent:
+            if sys.version_info[0] == 2:
+                self._outfd = open(os.devnull, 'w')
+            else:
+                self._outfd = subprocess.DEVNULL
 
     def __enter__(self):  # type: () -> IsolatedEnv
         """
@@ -54,8 +65,17 @@ class IsolatedEnvBuilder(object):
         """
         self._path = tempfile.mkdtemp(prefix='build-env-')
         try:
-            executable, pip_executable = _create_isolated_env(self._path)
-            return _IsolatedEnvVenvPip(path=self._path, python_executable=executable, pip_executable=pip_executable)
+            executable, pip_executable = _create_isolated_env(
+                self._path,
+                self._silent,
+                self._outfd,
+            )
+            return _IsolatedEnvVenvPip(
+                path=self._path,
+                python_executable=executable,
+                pip_executable=pip_executable,
+                outfd=self._outfd,
+            )
         except Exception:  # cleanup folder if creation fails
             self.__exit__(*sys.exc_info())
             raise
@@ -80,18 +100,20 @@ class _IsolatedEnvVenvPip(IsolatedEnv):
     Non-standard paths injected directly to sys.path still be passed to the environment.
     """
 
-    def __init__(self, path, python_executable, pip_executable):
-        # type: (str, str, str) -> None
+    def __init__(self, path, python_executable, pip_executable, outfd):
+        # type: (str, str, str, Union[int, TextIO, IO[str]]) -> None
         """
         Define an isolated build environment.
 
         :param path: the path where the environment exists
         :param python_executable: the python executable within the environment
         :param pip_executable: an executable that allows installing packages within the environment
+        :param outfd: Output file descriptor
         """
         self._path = path
         self._pip_executable = pip_executable
         self._python_executable = python_executable
+        self._outfd = outfd
 
     @property
     def path(self):  # type: () -> str
@@ -132,14 +154,15 @@ class _IsolatedEnvVenvPip(IsolatedEnv):
                 '-r',
                 os.path.abspath(req_file.name),
             ]
-            subprocess.check_call(cmd)
+            subprocess.check_call(cmd, stdout=self._outfd)
         finally:
             os.unlink(req_file.name)
 
 
 if sys.version_info[0] == 2:  # noqa: C901 # disable if too complex
 
-    def _create_isolated_env(path):  # type: (str) -> Tuple[str, str]
+    def _create_isolated_env(path, quiet=True, outfd=sys.stdout):
+        # type: (str, bool, Union[int, TextIO, IO[str]]) -> Tuple[str, str]
         """
         On Python 2 we use the virtualenv package to provision a virtual environment.
 
@@ -151,6 +174,8 @@ if sys.version_info[0] == 2:  # noqa: C901 # disable if too complex
         cmd = [str(path), '--no-setuptools', '--no-wheel', '--activators', '']
         if pip is not None:
             cmd.append('--no-pip')
+        if quiet:
+            cmd.append('--quiet')
         result = cli_run(cmd, setup_logging=False)
         executable = str(result.creator.exe)
         pip_executable = executable if pip is None else sys.executable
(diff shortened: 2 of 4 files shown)


src/build/__main__.py
```diff
@@ -42,6 +42,19 @@ def _error(msg, code=1):  # type: (str, int) -> None  # pragma: no cover
...
 
+def _print_info(msg):  # type: (str) -> None
+    """
+    Custom build.print_info implementation to print directly to stdout, and with
+    colors.
+    """
+    prefix = '*'
+    end = ''
+    if sys.stdout.isatty():
+        prefix = '\33[1m\33[36m' + prefix + '\33[39m'
```
gaborbernat, line 53: If you're introducing colors we should respect ``NO_COLOR`` env var, and add a flag to disable it. It can be on by default.

src/build/__main__.py
```diff
@@ -42,6 +42,19 @@ def _error(msg, code=1):  # type: (str, int) -> None  # pragma: no cover
     exit(code)
 
 
+def _print_info(msg):  # type: (str) -> None
```
gaborbernat, line 45: Any reason this is private, and we could not allow the caller of the API (when called programmatically) to choose stdout/logging)?

src/build/env.py
```diff
@@ -42,9 +42,20 @@ def install(self, requirements):  # type: (Iterable[str]) -> None
...
+        Builder object for isolated environment
+
+        :param silent: Silence the output
+        """
         self._path = None  # type: Optional[str]
+        self._silent = silent
+        self._outfd = sys.stdout  # type: Union[int, TextIO, IO[str]]
+        if silent:
+            if sys.version_info[0] == 2:
+                self._outfd = open(os.devnull, 'w')
```
gaborbernat, line 56: How does this work on Windows?

src/build/env.py
```diff
@@ -132,14 +154,15 @@ def install(self, requirements):  # type: (Iterable[str]) -> None
                 '-r',
                 os.path.abspath(req_file.name),
             ]
-            subprocess.check_call(cmd)
+            subprocess.check_call(cmd, stdout=self._outfd)
```
gaborbernat, line 157: What about stderr?

src/build/env.py
```diff
@@ -182,11 +208,14 @@ def _create_isolated_env(path):  # type: (str) -> Tuple[str, str]
             # Scenario 4: no pip can be found, we might be able to provision one into the build env via ensurepip
             cmd = [executable, '-Im', 'ensurepip', '--upgrade', '--default-pip']
             try:
-                subprocess.check_call(cmd, cwd=path)
+                subprocess.check_call(cmd, cwd=path, stdout=outfd)
```
gaborbernat, line 211: stderr?

src/build/env.py
```diff
@@ -182,11 +208,14 @@ def _create_isolated_env(path):  # type: (str) -> Tuple[str, str]
...
             except subprocess.CalledProcessError:  # pragma: no cover
                 pass  # pragma: no cover
             # avoid the setuptools from ensurepip to break the isolation
-            subprocess.check_call([executable, '-Im', 'pip', 'uninstall', 'setuptools', '-y'])
+            subprocess.check_call(
+                [executable, '-Im', 'pip', 'uninstall', 'setuptools', '-y'],
+                stdout=outfd,
+            )
```
gaborbernat, line 218: stderr?

gaborbernat (changes requested):
Needs addressed:

- no_color, --no-color flag, env/var
- stderr handling too alongside stdout
- tests for logging printer

layday:
Some thoughts:

* `get_dependencies` should also print a message before calling `get_requires_for_x`.  We could think about adding more diagnostic messages, e.g. with the name of the backend as suggested in #188 or with the reason a dependency wasn't met in `check_dependencies`.
* I don't like that customising the output requires overwriting a module variable.  I would rather a callback or a context manager was used.
* It's hard telling `build`'s and the backend's output apart.  If we're to take a leaf out of `pip`'s book, we could indent the backend's output.

src/build/__main__.py
FFY00: This color is only available in the main and is only active in a TTY. But I can add an env var if that really matters. We currently already have colors in main and nobody has asked for that. I believe checking for a TTY should handle 99.9% of the cases, a lot of other software does the same.

src/build/__main__.py
FFY00: This is the private main implementation, the default one uses the logging module.

src/build/env.py
FFY00: It should.

https://docs.python.org/3/library/os.html#os.devnull

src/build/env.py
FFY00: Well, if there is an error we want to report that. All the things we are using seem to follow the standard of only printing to stderr when things go wrong.

FFY00:
> `get_dependencies` should also print a message before calling `get_requires_for_build_x`. We could think about adding more diagnostic messages, e.g. with the name of the backend as suggested in #188 or with the reason a dependency wasn't met in `check_dependencies`.

I think that would make sense for a debug message, not a standard info message. Users do not need to know when/if we are fetching the dependencies.
As I said in #142 I want to make the output from us pretty limited, this would mean refraining to the high level action, which in this case is build a distribution.
I am totally fine adding logs for `get_requires_*` and other diagnostic stuff, but I think that should not be printed by default.

> I don't like that customising the output requires overwriting a module variable. I would rather a callback or a context manager was used.

You are right. I wanted to specifically allow to overwrite it module-wise, but people might not necessarily want to do that.
I propose making `print_info` a `@staticmethod` of `ProjectBuilder`. This allows people to either overwrite the class method or the instance method.

> It's hard telling `build`'s and the backend's output apart. If we're to take a leaf out of `pip`'s book, we could indent the backend's output.

Hum, I wanted something simple and a bit minimalistic. I thought making the whole message bold would be reasonably distinctive. Indenting the backend output would be reasonable but makes the implementation more complex, especially if we want the output in realtime. If the issue is the uncolored output, we can totally change the prefix there to something like `[build]`.

layday:
> > `get_dependencies` should also print a message before calling `get_requires_for_build_x`. 

> I think that would make sense for a debug message, not a standard info message. Users do not need to know when/if we are fetching the dependencies.

The backend (e.g. `setuptools`) might print to stdout when building without isolation, when we call `get_dependencies` as part of checking if dependencies are satisfied, so there might be a block of text which will precede `* Building ...`.  This confused me when I tried it out, hence my suggestion.

> This allows people to either overwrite the class method or the instance method.

I'm generally averse to mutating things in Python unless it's done for performance reasons.  My preference would be to pass a callback to `__init__` which would default to `print_info` from `build`.  Consumers would be able to use `functools.partial` or a wrapper to override the default for all instances of the class if that's what they want.

> we can totally change the prefix there to something like `[build]`

I think that'd help with making our output stand out.

FFY00:
> The backend (e.g. `setuptools`) might print to stdout when building without isolation, when we call `get_dependencies` as part of checking if dependencies are satisfied, so there might be a block of text which will precede `* Building ...`. This confused me when I tried it out, hence my suggestion.

I did notice that in my testing, it makes sense.

> I'm generally averse to mutating things in Python unless it's done for performance reasons. My preference would be to pass a callback to `__init__` which would default to `print_info` from `build`. Consumers would be able to use `functools.partial` or a wrapper to override the default for all instances of the class if that's what they want.

How would people override for all instances? Overriding a public method is essentially the same as overriding a public callback property.

> I think that'd help with making our output stand out.

Just to make it clear, I was suggesting that for the uncolored output.

layday:
> How would people override for all instances? Overriding a public method is essentially the same as overriding a public callback property.

I meant that `print_info` should be passed as an argument to the `ProjectBuilder`'s constructor, so that people wouldn't be enticed to overwrite the method on the class instead of the instance: `ProjectBuilder(print_info=...)`.

FFY00:
> > How would people override for all instances? Overriding a public method is essentially the same as overriding a public callback property.
> 
> I meant that `print_info` should be passed as an argument to the `ProjectBuilder`'s constructor, so that people wouldn't be enticed to overwrite the method on the class instead of the instance: `ProjectBuilder(print_info=...)`.

Overriding the class method is exactly what I want to support. In most cases you want to change the implementation for all instances, not just one.

layday:
Printing is not something I’d fully expect the ProjectBuilder class to implement, but overriding the method by subclassing ProjectBuilder is perfectly valid of course.

src/build/__main__.py
gaborbernat: Yeah, let's please follow https://no-color.org as does pytest, tox, etc.

src/build/env.py
gaborbernat: > All the things we are using seem to follow the standard of only printing to stderr when things go wrong.

Can you detail what things are you referring to?

gaborbernat:
> I think that would make sense for a debug message, not a standard info message. Users do not need to know when/if we are fetching the dependencies.

I don't agree with this. They might not want to know by default about those on the success case, however when the build fails they'll want to see it to track down the issues. And even on the success case often it's preferable to have the option to display the output of those tools so you can see where/what happens (this is important if for example a given file is not included in the package, you'll need the output to understand why).

src/build/env.py
FFY00: Okay, so, all these changes about the stdout handler are to private API. I introduced a `silent` bool argument to `IsolatedEnvBuilder`, but for easier implementation I opted to pass a file descriptor for stdout to most of the private API.
In the private API, remember this is the environment building, we only call pip, ensurepip and virtualenv, and all of those will only print to stderr when something goes wrong. This means if everything goes fine, there will be no output, but if there is an error in the subprocess they can still print to stderr and the users will see the error.

TLDR: I don't want to swallow stderr because I want users to see the error if for eg. pip fails install a bad dependency. Only stdout, which irrelevant if everything goes right.

Does that make sense?

FFY00:
> I don't agree with this. They might not want to know by default about those on the success case, however when the build fails they'll want to see it to track down the issues. And even on the success case often it's preferable to have the option to display the output of those tools so you can see where/what happens (this is important if for example a given file is not included in the package, you'll need the output to understand why).

Yeah, I agreed to this when @layday presented their point. I will add the messages.

src/build/env.py
FFY00: @gaborbernat does this address your concerns? If so, I can go ahead and push a new version fixing everything.

gaborbernat: My preferred approach to this problem is to always swallow everything in a silent mode, but save in memory. If the build fails show all. This is what tox does too. Sometimes the stdout also contains useful information to track down issues.

gaborbernat:
@FFY00 you planning to address these? Do you want to open up a new PR? (or will this block #220 for a while?)

FFY00:
I am planning to address this, just ETIME :stuck_out_tongue: 
With so many moving parts and limited time, it's hard to manage things I am supposed to work on.

This won't block #220, I wasn't considering this an "active" PR.

FFY00:
Closing as we'll rename master. I'll open a new PR soon once I have address the comments here.

Issue #194, opened 2020-12-02
Dependencies from `get_requires_for_build_sdist` are not installed in isolated environment


Carreau:
It seem the with the latest version of build; setuptools-scm decides that my project version is always 0.0.0; is that expected ? I found other issues mentioning setutools-scm, but it seem more targeted toward adopting setuptools-scm for build itself. Anly pointer on how to fix this beyond not using 0.1 ? 

See https://github.com/zarr-developers/zarr-python/pull/666

layday:
This isn't an issue with `build`; you haven't added `setuptools-scm` to your `pyproject.toml` requires.  I presume because setuptools is invoked through PEP 517 it does not honour `setup_requires`.

gaborbernat:
Yeah, you actually need to specify your build dependencies in pyproject.toml.

What you have https://github.com/zarr-developers/zarr-python/blob/master/pyproject.toml#L2, and what you should https://github.com/pypa/virtualenv/blob/main/pyproject.toml#L5

`setup_requires` is deprecated and has been for a long time now.

layday:
Upon closer inspection `setuptools` should return its `setup_requires` from `get_requires_for_build_x`.  Are we not installing the dependencies from `get_requires` before builing a distribution in an isolated environment?

layday:
I'm reopening this because it seems like a legitimate issue with isolated builds.  Acorrding to PEP 517:

> The `build_sdist` hook is executed in an environment which contains the bootstrap requirements from `pyproject.toml` and those specified by the `get_requires_for_build_sdist` hook.

We'll probably want to install these after the `build_dependencies` in `_build_in_isolated_env`:

https://github.com/pypa/build/blob/b39dcddb6c7e822398c49b81d73fcbcdedea607b/src/build/__main__.py#L53

Thanks for reporting, @Carreau.

FFY00:
Fixed in 260c51800dc4df0fff7e522bff7bba9d975a0bcb.

PR #195, opened 2020-12-02
build: install backend specified dependencies when building with isolation

FFY00:
PEP 517 says that we should use the get_requires_for_* hooks to fetch
extra build dependencies the backend might specify. Currently, we are
not installing these dependencies into the isolted environment, this
patch addresses that.

It also slightly changes the isolation build logic to create a isolated
environment for each distribution we want to build. Since we are now
calling the get_requires_for_{sdist,wheel} hooks to install extra
dependencies, the might change between a sdist and wheel. Since we do
not want sdist dependencies leaking into the environment where the wheel
is built, or vice-versa, we need to separate the environments.

Perhaps in the future we could try to optimize this to see if the sdist
and wheel dependencies are the same and avoid creating a new isolated
environment if they aren't?

Addresses #194

### DIFF
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -48,10 +48,13 @@ def _format_dep_chain(dep_chain):  # type: (Sequence[str]) -> str
 
 def _build_in_isolated_env(builder, outdir, distributions):
     # type: (ProjectBuilder, str, List[str]) -> None
-    with IsolatedEnvBuilder() as env:
-        builder.python_executable = env.executable
-        env.install(builder.build_dependencies)
-        for distribution in distributions:
+    for distribution in distributions:
+        with IsolatedEnvBuilder() as env:
+            builder.python_executable = env.executable
+            # first install the build dependencies
+            env.install(builder.build_dependencies)
+            # then get the extra required dependencies from the backend (which was installed in the call above :P)
+            env.install(builder.get_dependencies(distribution))
             builder.build(distribution, outdir)
 
 
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -102,13 +102,23 @@ def test_version(capsys):
 @pytest.mark.isolated
 def test_build_isolated(mocker, test_flit_path):
     build_cmd = mocker.patch('build.ProjectBuilder.build')
+    required_cmd = mocker.patch(
+        'build.ProjectBuilder.get_dependencies',
+        side_effect=[
+            ['dep1', 'dep2'],
+        ],
+    )
     mocker.patch('build.__main__._error')
     install = mocker.patch('build.env._IsolatedEnvVenvPip.install')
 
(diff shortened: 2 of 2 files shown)


gaborbernat (changes requested):
Add tests, please.

gaborbernat (approved):

src/build/__main__.py
```diff
@@ -48,10 +48,13 @@ def _format_dep_chain(dep_chain):  # type: (Sequence[str]) -> str
...
-        builder.python_executable = env.executable
-        env.install(builder.build_dependencies)
-        for distribution in distributions:
+    for distribution in distributions:
+        with IsolatedEnvBuilder() as env:
+            builder.python_executable = env.executable
+            # first install the build dependencies
+            env.install(builder.build_dependencies)
+            # then get the extra required dependencies from the backend
+            env.install(builder.get_dependencies(distribution))
```
gaborbernat, line 57: What will happen here if the backend returns some non-PEP-440 compliant specs?

layday, line 57: Should we combined these so we don't have to invoke `pip` twice, which will slow down isolated builds?

FFY00: We first need to install the build dependencies, say setuptools, only after that we can ask the backend, setuptools in this case, which extra dependencies it needs.

FFY00: PEP 517 says

> This hook MUST return an additional list of strings containing PEP 508 dependency specifications

PEP 508 says

> Versions may be specified according to the PEP-440 [4] rules

FFY00: https://www.python.org/dev/peps/pep-0508/#grammar

layday: Ah yes, I didn't think that through.

FFY00:
Adjusted the comment to clarify layday's question.

src/build/__main__.py
gaborbernat: I know what they must, but if they mess it up will we handle it graciously, and by graciously I mean error out and show the adequate error?

gaborbernat: Yeah, PEP-517 spec mandates this to be two separate calls.

FFY00: Right now we pass it down to pip and let it handle it. So even with #192 the user would see something like.

```
ERROR: Could not find a version that satisfies the requirement ...
ERROR: No matching distribution found for ...
```

I am not opposed to use pypa/packaging (only in the isolation build!) to should a custom error, but I do not think it should block this.

gaborbernat: As long as pip displays an adequate error I'm fine, but we can work on making clearer the source of the error. Because in this case, the error will be in the backend configuration files... Though doesn't have to be in this PR.

PR #197, opened 2020-12-07
build: return the filename of the built dist from `build`

layday:


gaborbernat (approved):

src/build/__init__.py
```diff
@@ -218,12 +218,13 @@ def check_dependencies(self, distribution):  # type: (str) -> Set[Tuple[str, ...
...
         return {u for d in dependencies for u in check_dependency(d)}
 
-    def build(self, distribution, outdir):  # type: (str, str) -> None
+    def build(self, distribution, outdir):  # type: (str, str) -> Text
         """
         Build a distribution.
 
         :param distribution: Distribution to build (``sdist`` or ``wheel``)
         :param outdir: Output directory
+        :returns: The filename of the built distribution
```
gaborbernat, line 227: This part is called the basename (as per PEP-517), so we should use the same terminology here:

> build_wheel
> Must build a .whl file, and place it in the specified wheel_directory. It must return the basename (not the full path) of the .whl file it creates, as a unicode string.

See https://www.python.org/dev/peps/pep-0517/#build-wheel

gaborbernat: That being said I'd say we should probably return `outdir / filename` here though as it's more usable by the end user.

layday: Good call, changed it to `outdir / filename`.

FFY00 (approved):

PR #200, opened 2020-12-17
Added .editorconfig according to PEP 8

KOLANICH:
See https://editorconfig.org/ for more info.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+root = true
+
+[*.py]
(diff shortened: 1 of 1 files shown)


gaborbernat (commented):
I'm personally -0.5 on this.

layday:
I don’t think that this is really necessary when we’re using Black with pre-commit. Editorconfig almost feels like a vestige from a bygone era.

PR #203, opened 2021-01-11
Document CLI usage

EpicWink:
Document example CLI usage in the README and on the home-page of the docs. Move CLI help to a separate docs page "CLI Usage"

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,15 @@ See the [documentation](https://pypa-build.readthedocs.io/en/latest/) for more i
 $ pip install build
 ```
 
+### Usage
+
+```bash
+python -m build . --sdist --wheel
+```
+
+This will build the package in an isolated environment, generating a
+source-distribution and wheel in the directory `dist/`.
+
 ### Code of Conduct
 
 Everyone interacting in the build's codebase, issue trackers, chat rooms, and mailing lists is expected to follow
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -1,0 +1,16 @@
+=========
+CLI Usage
+=========
+
+.. autoprogram:: build.__main__:main_parser()
+   :prog: python -m build
+
+.. note::
+
+   A ``pyproject-build`` CLI script is also available, so that tools such as pipx_
+   can use it.
+
+By default build will build the package in an isolated
+environment, but this behavior can be disabled with `--no-isolation`.
+
+.. _pipx: https://github.com/pipxproject/pipx
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -9,16 +9,12 @@ A simple, correct :pep:`517` package builder.
 build will invoke the :pep:`517` hooks to build a distribution package.
 It is a simple build tool and does not perform any dependency management.
 
-.. autoprogram:: build.__main__:main_parser()
-   :prog: python -m build
+.. code-block::
 
-.. note::
+   python -m build . --sdist --wheel
 
-   A ``pyproject-build`` CLI script is also available, so that tools such as pipx_
-   can use it.
-
-By default build will build the package in an isolated
-environment, but this behavior can be disabled with `--no-isolation`.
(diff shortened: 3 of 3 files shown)


gaborbernat (approved):

FFY00:
I do not agree with these changes. While I think it is fine to add the usage to the README, I think replacing CLI description in the main page of the documentation with a very minimal one is, arguably, worse. This is something that would make sense if the main page was too big, but it isn't, the full CLI usage fits in one page. Putting it on a separate page makes this information yet another click away from the user, and this is a CLI app, it would make sense to have the usage in the main page. I am reverting this part.

gaborbernat:
Hmmm, I have the opposite POV but 🤷🏻

EpicWink:
> Putting it on a separate page makes this information yet another click away from the user, and this is a CLI app, it would make sense to have the usage in the main page.

Two arguments against this:
* The main page is the first thing a potential or existing user sees. For potential users, they need to know what `build` is, how it can solve their problems, and maybe how they can get started quickly. For existing users, they may want to know recent changes, complete CLI/API reference. Both camps would probably like the full feature-set of `build`.
* As a CLI app, the CLI usage can be obtained by running `python3 -m build --help`, which it prompts the user to run when they mess up the arguments. If you're using the CLI usage for anything other than a description of the command-line parameters, then a section specialised to that purpose should be made; for example, to describe features, have a features section

> This is something that would make sense if the main page was too big, but it isn't

With the inclusion of an enumeration of `build`'s features, it could be longer. I agree that after my PR the main page was a stub, but I felt like the CLI usage was in part used to fill the space.

To fill space, I'd recommend adding a features list, and a statement on the scope of `build`. I know, the features list would be like "make sdist, make wheel, uses PEP 517, fast", but having that list will give newcomers more information to make better decisions. Come to think of it, maybe even a sentence explaining the difference between a build front-end vs back-end. I'm happy to make a PR if you'd like, as I've just spent a few minutes thinking about it.

> the full CLI usage fits in one page.

PS: still need to scroll to see the no-isolation note at the bottom on a 1080p monitor. Maybe move it above?

FFY00:
> The main page is the first thing a potential or existing user sees. For potential users, they need to know what `build` is, how it can solve their problems, and maybe how they can get started quickly. For existing users, they may want to know recent changes, complete CLI/API reference. Both camps would probably like the full feature-set of `build`.

Sure, but I see no reason why the CLI usage needs to be in another page, the opposite, it's annoying to me. We can have description of the project and maybe an example usage and still have CLI usage below. If the page starts becoming too big we could maybe move the CLI usage to a different page, but that is not the case here.

For existing users that want to see the changelog we will add a changelog section available in the menu. I am trying to optimize here for the most combined time from people, which is getting started.

The main purpose of this project is to provide a CLI tool, but it does also provide a Python API. This API however, is not useful to many people and really is just there as an extra to enable people to do their own thing.

> With the inclusion of an enumeration of `build`'s features, it could be longer. I agree that after my PR the main page was a stub, but I felt like the CLI usage was in part used to fill the space.
>
> To fill space, I'd recommend adding a features list, and a statement on the scope of `build`. I know, the features list would be like "make sdist, make wheel, uses PEP 517, fast", but having that list will give newcomers more information to make better decisions. Come to think of it, maybe even a sentence explaining the difference between a build front-end vs back-end. I'm happy to make a PR if you'd like, as I've just spent a few minutes thinking about it.

The CLI is there, deliberately, to provide a quick intro to the tool, as well as the available features. build is fairly simple, you don't need a big explanation. I find it describing all features textually very cumbersome, as they are pretty simple. You need to look at the CLI if you want to use any of them, it will just be repeating what you say said previously.

I do think, however, that the introduction could be enhanced, specifically to be more friendly to beginners.

FFY00:
TLDR: The introduction to the tool could be worked on, specifically to be more beginner friendly, but I find it very unlikely you will convince me we need to have the CLI usage in a different page.

PR #204, opened 2021-01-11
Allow pre-releases in `check_dependency`

layday:


layday:
The Windows 3.8 failure appears to be `xdist`-related.  I'm not sure why a test suite which takes 6s to run on my machine is failing after 10m in CI but these issues are recurring and frustrating to debug.

FFY00 (approved):

PR #212, opened 2021-01-26
Work around https://github.com/python/mypy/issues/1153

musicinmybrain:
Work around https://github.com/python/mypy/issues/1153; see also https://github.com/pypa/twine/pull/551.

This fixes #221, “Mypy error: Module 'importlib' has no attribute 'metadata'”.

### DIFF
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -64,7 +64,7 @@ def check_dependency(req_string, ancestral_req_strings=(), parent_extras=frozens
     if sys.version_info >= (3, 8):
         from importlib import metadata as importlib_metadata
     else:
-        import importlib_metadata
+        import importlib_metadata  # type: ignore
 
     req = packaging.requirements.Requirement(req_string)
 


src/build/__init__.py
```diff
@@ -64,7 +64,7 @@ def check_dependency(req_string, ancestral_req_strings=(), parent_extras=frozens
     if sys.version_info >= (3, 8):
         from importlib import metadata as importlib_metadata
     else:
-        import importlib_metadata
+        import importlib_metadata  # type: ignore
```
FFY00, line 67: Please add a quick comment here saying that we need the `# type: ignore` and linking to the mypy issue.

Otherwise, LGTM. Thanks :blush:

FFY00 (changes requested):

src/build/__init__.py
FFY00: Probably best to put the comment on the line above.

musicinmybrain:
This is not a correct approach; `build` already implements the workaround `twine` implemented. Plus, it does not fix the problem.

src/build/__init__.py
musicinmybrain: Please see https://github.com/pypa/build/pull/213 instead. Thanks!

Issue #211, opened 2021-01-26
Mypy error: Module 'importlib' has no attribute 'metadata'


musicinmybrain:
In a fresh checkout:

`python3 --version; mypy --version; mypy src/build`

results in:

```
Python 3.9.1
mypy 0.782
src/build/__init__.py:65: error: Module 'importlib' has no attribute 'metadata'
Found 1 error in 1 file (checked 4 source files)
```

(The versions are those current in Fedora 33, but the particular versions are not terribly important.)

This looks like https://github.com/python/mypy/issues/1153. I will prepare a PR with the same workaround used in https://github.com/pypa/twine/pull/551.

musicinmybrain:
https://github.com/pypa/build/pull/212

musicinmybrain:
Well, reading farther into the twine patch, it seems they ended up doing a sys.version_info test similar to what `build` already does: https://github.com/pypa/twine/pull/551/files

So I am not sure why I see this…

layday:
I cannot reproduce this:

```
$ python3 --version; mypy --version; mypy src/build
Python 3.9.1
mypy 0.782
Success: no issues found in 4 source files
```

Can you try deleting your mypy cache and trying again?  This looks like a typeshed-related issue to me.

PR #213, opened 2021-01-26
mypy: fix importlib.metadata import error

musicinmybrain:
This fixes https://github.com/pypa/build/issues/211. It replaces incorrect PR https://github.com/pypa/build/pull/212.


layday:
Hmm, that doesn't seem to be the same issue as the one you linked from the mypy tracker.

FFY00 (approved):

PR #217, opened 2021-01-29
build: implement metadata preparation hook

### LINKED ISSUE #130: Support for metadata hook
Am I missing something, or is [`prepare_metadata_for_build_wheel()`](https://www.python.org/dev/peps/pep-0517/#id13) not implemented?

It seems like it's something that isn't specified in PEP-517, but I was of the impression, that metadata hook is a mandatory call for the frontend, which _will_ call that hook before `build_wheel()`, if available. After all, it's not up to the build frontend whether metadata is needed, the responsibility for that decision lies on the code that actually builds the distribution.

Am I correct in my interpretation of the spec?

uranusjr:
Fix #130.

I make the function signature require the user to pass in `"wheel"` explicitly. This is of course redundant, but I have a feeling we’re going to have a `prepare_metadata_for_sdist` at some point and it’s better to leave room now than having to change things later. And `prepare("wheel")` is arguably more consistent with `build("wheel")` anyway.

Due to the issues mentioned in https://github.com/pypa/build/issues/130#issuecomment-769849702, I’m *not* using the `build_wheel` fallback provided by `pep517`, and instead return `None` when the backend does not implement `prepare_metadata_for_wheel`. The returned result is stored in a member attribute and popped out when `build_wheel` is called.

### DIFF
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -168,6 +168,7 @@ class ProjectBuilder(object):
             python_executable=python_executable,
             runner=self._runner,
         )
+        self._prepared_metadata_directories = {}  # type: Dict[str, str]
 
     def _runner(self, cmd, cwd=None, extra_environ=None):
         # type: (Sequence[str], Optional[Union[bytes, Text]], Optional[Dict[str, str]]) -> None
@@ -247,6 +248,36 @@ class ProjectBuilder(object):
         dependencies = self.get_dependencies(distribution, config_settings).union(self.build_dependencies)
         return {u for d in dependencies for u in check_dependency(d)}
 
+    def prepare(self, distribution, outdir, config_settings=None):
+        # type: (str, str, Optional[ConfigSettings]) -> Optional[str]
+        """
+        Prepare metadata for a distribution.
+
+        :param distribution: Distribution to build (must be ``wheel``)
+        :param outdir: Output directory
+        :returns: The full path to the prepared metadata directory
+        """
+        prepare = getattr(self._hook, 'prepare_metadata_for_build_{}'.format(distribution))
+        outdir = os.path.abspath(outdir)
+
+        if os.path.exists(outdir):
+            if not os.path.isdir(outdir):
+                raise BuildException("Build path '{}' exists and is not a directory".format(outdir))
+        else:
+            os.mkdir(outdir)
+
+        try:
+            with _working_directory(self.srcdir):
+                basename = prepare(outdir, config_settings, _allow_fallback=False)  # type: str
+            self._prepared_metadata_directories[distribution] = path = os.path.join(outdir, basename)
+        except pep517.wrappers.BackendUnavailable:
+            raise BuildException("Backend '{}' is not available.".format(self._backend))
+        except pep517.wrappers.HookMissing:
+            return None
+        except Exception as e:  # noqa: E722
+            raise BuildBackendException('Backend operation failed: {!r}'.format(e))
+        return path
+
     def build(self, distribution, outdir, config_settings=None):  # type: (str, str, Optional[ConfigSettings]) -> str
         """
         Build a distribution.
@@ -266,8 +297,15 @@ class ProjectBuilder(object):
             os.mkdir(outdir)
 
         try:
+            metadata_directory = self._prepared_metadata_directories.pop(distribution)
+        except KeyError:
+            kwargs = {}  # type: Dict[str, str]
+        else:
+            kwargs = {'metadata_directory': metadata_directory}
+
+        try:
             with _working_directory(self.srcdir):
-                basename = build(outdir, config_settings)  # type: str
+                basename = build(outdir, config_settings, **kwargs)  # type: str
                 return os.path.join(outdir, basename)
         except pep517.wrappers.BackendUnavailable:
             raise BuildException("Backend '{}' is not available.".format(self._backend))
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -48,7 +48,7 @@ class BuildBackendException(Exception):
         super(BuildBackendException, self).__init__()
         self.exception = exception  # type: Exception
 
-    def __repr__(self):  # type: () -> str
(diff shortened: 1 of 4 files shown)


layday:
I like that `prepare` follows the same pattern as `build`. Why is the metadata path discarded during the build?

uranusjr:
PEP 517 does not say whether the prepared metadata directory may be reused across multiple `build_wheel` call, only that it should be reused, so I took the safe route and use it only exactly once.

It also just occured to me that the PEP says the frontend should only pass the metadata directory into `build_whell` *[if the frontend] depends on the wheel resulting from this call to have metadata matching this earlier call*, so an option should be provided for the frontend to have a choice to not do it. So maybe it’s not that good an idea to store the prepare result as a state after all, and we should just let the user choose what to they want.

uranusjr:
> nobody seems to have sought to use the `metadata_directory` argument of `build_wheel`, that I can tell.

pip does. And arguably all major backends are implemented incorrectly. According to PEP 517, if a backend implements `prepare_metadata_for_wheel`, it MUST ensure the resulting wheel contains identical metadata to the prepared metadata directory. setuptools actually fails to meet the promise in at least one situation, leading to numerous pip bug reports.

layday:
> setuptools actually fails to meet the promise in at least one situation, leading to numerous pip bug reports.

Is that because the prepared metadata becomes stale (e.g. when editing a package locally) or because the prepared metadata is simply incorrect?

uranusjr:
The prepared metadata is incorrect in the particular case (pypa/pip#9446).

layday:
We could check with backend devs to see why this wasn't implemented, if the PEP needs to be modified and if pip wouldn't be better off comparing prepared and built wheel metadata on their end.

uranusjr:
There are issue filed to those backends now (I filed one for Flit just a moment ago): pypa/setuptools#1825 python-poetry/poetry#1078 takluyver/flit#389

uranusjr:
So I raised the issue on discussion.python.org, and from the response, I think we should still implement the `metadata_directory` argument because it is the right thing to do. Backends don’t use it, but that’s an issue they should fix.

FFY00:
I agree. I think a backend simply ignoring the argument is *fine*, though I think they should all verify if it's the same metadata if they're gonna rebuild it anyway.

src/build/__init__.py
```diff
@@ -219,7 +219,38 @@ def check_dependencies(self, distribution, config_settings=None):
...
+            os.mkdir(outdir)
+
+        try:
+            with _working_directory(self.srcdir):
+                basename = prepare(outdir, config_settings, _allow_fallback=False)  # type: str
+            path = os.path.join(outdir, basename)
+        except pep517.wrappers.BackendUnavailable:
+            raise BuildException("Backend '{}' is not available.".format(self._backend))
+        except pep517.wrappers.HookMissing:
+            return None
```
uranusjr, line 247: `prepare('wheel')` currently returns `None` if `prepare_metadata_for_build_wheel` is not implemented. Should an exception be raised instead?

src/build/__init__.py
```diff
@@ -219,7 +219,38 @@ def check_dependencies(self, distribution, config_settings=None):
...
+        except Exception as e:  # noqa: E722
+            raise BuildBackendException('Backend operation failed: {!r}'.format(e))
+        return path
+
+    def build(self, distribution, outdir, metadata_directory=None, config_settings=None):
```
layday, line 252: `metadata_directory` -> `metadatadir` or `outdir` -> `output_directory` for consistency.

src/build/__init__.py
layday: Possibly, would that mean exposing `pep517`'s exceptions or creating our own equivalents?

src/build/__init__.py
uranusjr: PEP 517 names the arguments `wheel_directory` and `metadata_directory`. Would that work?

layday: Well, `build` works both with wheels and sdists, hence 'out' or 'output'.

src/build/__init__.py
layday: That is to say, I don't think a generic `BuildBackendException` is satisfacory.

src/build/__init__.py
uranusjr: I’m going with `output_directory`.

src/build/__init__.py
uranusjr: Agreed. The exception raised here need to be different, otherwise the caller won’t be able to distinguish an expected code path (`prepare_metadata_for_build_wheel` not implemented by the backend) from a legistimate error in the backend (`BuildBackendException`) or the build process (`BuildException`). So a new exception will be needed if we raise here.

gaborbernat (approved):
@uranusjr made some changes to reduce code duplication, but now looks 👍🏻

tests/test_projectbuilder.py
```diff
@@ -384,3 +384,48 @@ def test_build_with_dep_on_console_script(tmp_path, demo_pkg_inline, capfd, mock
...
+def test_prepare_not_dir_outdir(mocker, tmp_dir, test_flit_path):
+    mocker.patch('pep517.wrappers.Pep517HookCaller', autospec=True)
+
+    builder = build.ProjectBuilder(test_flit_path)
+
+    out = os.path.join(tmp_dir, 'out')
+    with open(out, 'w') as f:
+        f.write('Not a directory')
+
+    with pytest.raises(build.BuildException):
```
FFY00, line 430: Could we use `match` here to also cover the `__repr__`?

https://docs.pytest.org/en/stable/reference.html#pytest-raises

FFY00 (changes requested):

tests/test_projectbuilder.py
gaborbernat: Did not want to touch that, as left it as @uranusjr did it, but now updated.

FFY00 (approved):

PR #219, opened 2021-02-01
docs: add changelog

### LINKED ISSUE #169: Add a changelog system
pypa uses https://pypi.org/project/towncrier mostly, though we can also check https://github.com/nedbat/scriv as an option.

FFY00:
Fixes #169

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,0 +1,112 @@
++++++++++
+Changelog
++++++++++
+
+
+Unreleased
+==========
+
+- Check dependencies recursively (`PR #183`_, Fixes `#25`_)
+- Build wheel and sdist distributions in separate environments, as they may have different dependencies (`PR #195`_, Fixes `#194`_)
+- Add support for pre-releases in ``check_dependency`` (`PR #204`_, Fixes `#191`_)
+- Do not add the default backend requirements to ``requires`` when no backend is specified (`PR #177`_, Fixes `#107`_)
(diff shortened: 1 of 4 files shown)


gaborbernat (approved):
How do we generate this? Is it manually updated? 🤔

docs/changelog.rst
```diff
@@ -0,0 +1 @@
+../CHANGELOG.rst
```
uranusjr, line 1: Is this correct? I thought a `.. include:` derivative is needed.

CHANGELOG.rst
```diff
@@ -0,0 +1,112 @@
++++++++++
+Changelog
++++++++++
+
+
+Unreleased
+==========
+
+- Check dependencies recursively (`PR #183`_, Fixes `#25`_)
+- Build wheel and sdist distributions in separate envrionments, as they may have different dependencies (`PR #195`_, Fixes `#194`_)
```
layday, line 10: envrionments -> environments

docs/changelog.rst
FFY00: Github doesn't show file flags, but that is a symlink.

FFY00:
> How do we generate this? Is it manually updated? thinking
> 

Yes, manually updated.

gaborbernat:
> Yes, manually updated.

Would prefer something automatic, but we can start with this and add stuff on top of it.

docs/changelog.rst
uranusjr: Oh, makes sense, thanks.

PR #221, opened 2021-02-03
build: ensure console scripts are available on PATH during build

### LINKED ISSUE #214: Console scripts not available during build
When building, console scripts paths are not injected into the path hence console scripts are not available during build.

Example:
When cmake is installed, `cmake.exe` is not on path, leading to error during when when `cmake` is invoked with subprocess.

gaborbernat:
Resolves #214.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -55,6 +55,8 @@ test =
 typing =
     mypy==0.790
     typing-extensions>=3.7.4.3
+virtualenv =
+    virtualenv>=20.0.35
 
 [options.package_data]
 build =
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -158,7 +158,7 @@ class ProjectBuilder(object):
 
         self._build_system = build_system
         self._backend = self._build_system['build-backend']
-        self.script_dir = os.path.dirname(python_executable) if scripts_dir is None else scripts_dir
+        self._scripts_dir = scripts_dir
         self._hook = pep517.wrappers.Pep517HookCaller(
             self.srcdir,
             self._backend,
@@ -170,11 +170,13 @@ class ProjectBuilder(object):
     def _runner(self, cmd, cwd=None, extra_environ=None):
         # type: (Sequence[str], Optional[Union[bytes, Text]], Optional[Dict[str, str]]) -> None
         # if script dir is specified must be inserted at the start of PATH (avoid duplicate path while doing so)
-        paths = OrderedDict()  # type:  Dict[str, None]
-        paths[str(self.script_dir)] = None
-        paths.update((i, None) for i in os.environ.get('PATH', '').split(os.pathsep))
-        extra_environ = {} if extra_environ is None else extra_environ
-        extra_environ['PATH'] = os.pathsep.join(paths)
+        if self.scripts_dir is not None:
+            paths = OrderedDict()  # type:  Dict[str, None]
+            paths[str(self.scripts_dir)] = None
+            if 'PATH' in os.environ:
+                paths.update((i, None) for i in os.environ['PATH'].split(os.pathsep))
+            extra_environ = {} if extra_environ is None else extra_environ
+            extra_environ['PATH'] = os.pathsep.join(paths)
         pep517.default_subprocess_runner(cmd, cwd, extra_environ)
 
     @property
@@ -189,6 +191,17 @@ class ProjectBuilder(object):
     @python_executable.setter
     def python_executable(self, value):  # type: (Union[bytes, Text]) -> None
         self._hook.python_executable = value
+
+    @property
+    def scripts_dir(self):  # type: () -> Union[None, bytes, Text]
+        """
+        The folder where the scripts are stored for the python executable.
+        """
+        return self._scripts_dir
+
+    @scripts_dir.setter
+    def scripts_dir(self, value):  # type: (Union[None, bytes, Text]) -> None
+        self._scripts_dir = value
 
     @property
     def build_dependencies(self):  # type: () -> Set[str]
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -11,7 +11,8 @@ import os
 import sys
 import warnings
 
-from typing import AbstractSet, Iterator, Mapping, Optional, Sequence, Set, Text, Tuple, Union
+from collections import OrderedDict
+from typing import AbstractSet, Dict, Iterator, Mapping, Optional, Sequence, Set, Text, Tuple, Union
 
 import pep517.wrappers
 import toml
@@ -116,12 +117,13 @@ def _working_directory(path):  # type: (str) -> Iterator[None]
 
 
 class ProjectBuilder(object):
-    def __init__(self, srcdir, python_executable=sys.executable):
-        # type: (str, Union[bytes, Text]) -> None
+    def __init__(self, srcdir, python_executable=sys.executable, scripts_dir=None):
+        # type: (str, Union[bytes, Text], Optional[Union[bytes, Text]]) -> None
         """
         Create a project builder.
 
         :param srcdir: the source directory
+        :param scripts_dir: the location of the scripts dir (defaults to the folder where the python executable lives)
         :param python_executable: the python executable where the backend lives
         """
         self.srcdir = os.path.abspath(srcdir)  # type: str
@@ -156,13 +158,24 @@ class ProjectBuilder(object):
 
         self._build_system = build_system
         self._backend = self._build_system['build-backend']
-
+        self.script_dir = os.path.dirname(python_executable) if scripts_dir is None else scripts_dir
         self._hook = pep517.wrappers.Pep517HookCaller(
             self.srcdir,
             self._backend,
             backend_path=self._build_system.get('backend-path'),
             python_executable=python_executable,
+            runner=self._runner,
         )
+
+    def _runner(self, cmd, cwd=None, extra_environ=None):
+        # type: (Sequence[str], Optional[Union[bytes, Text]], Optional[Dict[str, str]]) -> None
+        # if script dir is specified must be inserted at the start of PATH (avoid duplicate path while doing so)
+        paths = OrderedDict()  # type:  Dict[str, None]
+        paths[str(self.script_dir)] = None
+        paths.update((i, None) for i in os.environ.get('PATH', '').split(os.pathsep))
+        extra_environ = {} if extra_environ is None else extra_environ
+        extra_environ['PATH'] = os.pathsep.join(paths)
+        pep517.default_subprocess_runner(cmd, cwd, extra_environ)
 
     @property
     def python_executable(self):  # type: () -> Union[bytes, Text]
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -132,7 +132,7 @@ class _IsolatedEnvVenvPip(IsolatedEnv):
         try:
             cmd = [
                 self.executable,
-                '-{}m'.format('E' if sys.version_info[0] == 2 else ''),
+                '-{}m'.format('E' if sys.version_info[0] == 2 else 'I'),
                 'pip',
                 'install',
                 '--no-warn-script-location',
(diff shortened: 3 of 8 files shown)


setup.cfg
```diff
@@ -55,6 +55,8 @@ test =
 typing =
     mypy==0.790
     typing-extensions>=3.7.4.3
+virtualenv =
```
gaborbernat, line 58: This extra allows users to opt-in into a faster isolated env creator.

src/build/__init__.py
```diff
@@ -156,14 +158,25 @@ def __init__(self, srcdir, python_executable=sys.executable):
 
         self._build_system = build_system
         self._backend = self._build_system['build-backend']
-
+        self.script_dir = os.path.dirname(python_executable) if scripts_dir is None else scripts_dir
```
FFY00, line 161: Defaulting to `os.path.dirname(python_executable)` is implementation specific, please remove it.

Call it `self.scripts_dir` to be more consistent. Here and in the other places.

tests/test_projectbuilder.py
```diff
@@ -7,6 +7,8 @@
 import sys
 import typing
 
+from textwrap import dedent
```
FFY00, line 10: Use top-level imports.

tests/test_projectbuilder.py
```diff
@@ -336,3 +333,55 @@ def test_not_dir_outdir(mocker, tmp_dir, test_flit_path):
...
+        if sys.version_info[0] == 3:
+            import shutil
+            exe_at = shutil.which("demo-pkg-inline")
+        else:
+            paths = os.environ["PATH"].split(os.pathsep)
+            exe_at = os.path.join(paths[0], "demo-pkg-inline{}".format(".exe" if sys.platform == "win32" else ""))
+            if not os.path.exists(exe_at):
+                exe_at = None
+        print("BB " + exe_at)
+    """
```
FFY00, line 371: Tab here. Is black running here?

FFY00 (changes requested):
Thank you! I just have some minor comments.

src/build/__init__.py
gaborbernat: > Defaulting to `os.path.dirname(python_executable)` is implementation specific

Felt like the less bad of just leaving it empty, it's only used for non-isolated builds. Not?

tests/test_projectbuilder.py
gaborbernat: Where do you see tab? Triple strings are not changed by pre-commit. Yes it's running, the CI passed.

src/build/__init__.py
FFY00: No, it's public API...

You could always
```
python -c 'import sysconfig; print(sysconfig.get_path("scripts"))'
```
which would be the correct solution.

tests/test_projectbuilder.py
FFY00: Sorry, I was asking to add a tab. The opening quotes are at a different level. Does black format it like this? Seems weird.

```python
    code = dedent(
        """
        import os
        import sys
        print("BB " + os.environ["PATH"])
        if sys.version_info[0] == 3:
            import shutil
            exe_at = shutil.which("demo-pkg-inline")
        else:
            paths = os.environ["PATH"].split(os.pathsep)
            exe_at = os.path.join(paths[0], "demo-pkg-inline{}".format(".exe" if sys.platform == "win32" else ""))
            if not os.path.exists(exe_at):
                exe_at = None
        print("BB " + exe_at)
    """
    )
```

src/build/__init__.py
```diff
@@ -156,14 +158,25 @@ def __init__(self, srcdir, python_executable=sys.executable):
...
             python_executable=python_executable,
+            runner=self._runner,
         )
 
+    def _runner(self, cmd, cwd=None, extra_environ=None):
+        # type: (Sequence[str], Optional[Union[bytes, Text]], Optional[Dict[str, str]]) -> None
+        # if script dir is specified must be inserted at the start of PATH (avoid duplicate path while doing so)
+        paths = OrderedDict()  # type:  Dict[str, None]
+        paths[str(self.script_dir)] = None
+        paths.update((i, None) for i in os.environ.get('PATH', '').split(os.pathsep))
```
layday, line 175: This will add `''`, the current working directory, to the path if the path is undefined or the empty string, is that intentional?

tests/test_projectbuilder.py
uranusjr: Black does not move triple quotes (that would change the string’s content and is not safe to do in general).

src/build/env.py
```diff
@@ -31,6 +31,11 @@ def executable(self):  # type: () -> str
         """Return the executable of the isolated build environment."""
         raise NotImplementedError
 
+    @abstractproperty
+    def script_dir(self):  # type: () -> str
+        """Return the executable of the isolated build environment."""
```
layday, line 36: Docstring refers to executable.

src/build/env.py
```diff
@@ -137,78 +144,58 @@ def install(self, requirements):  # type: (Iterable[str]) -> None
...
-        executable = os.path.join(path, env_scripts, exe)
-        if not os.path.exists(executable):
-            raise RuntimeError('Virtual environment creation failed, executable {} missing'.format(executable))
-        return executable
+    :param path: the location of the virtual environment
+    :return: the python executable
+    """
+    config_vars = sysconfig.get_config_vars().copy()  # globally cached, copy before altering it
+    config_vars['base'] = path
+    env_scripts = sysconfig.get_path('scripts', vars=config_vars)
```
layday, line 189: `sysconfig` will merge the variables from `get_config_vars`, copying and modifying it is not necessary:

```suggestion
    env_scripts = sysconfig.get_path('scripts', vars={'base': path})
```

src/build/env.py
```diff
@@ -137,78 +144,58 @@ def install(self, requirements):  # type: (Iterable[str]) -> None
...
+    if not env_scripts:
+        raise RuntimeError("Couldn't get environment scripts path")
+    exe = 'pypy3' if platform.python_implementation() == 'PyPy' else 'python'
+    if os.name == 'nt':
+        exe = '{}.exe'.format(exe)
+    executable = os.path.join(path, env_scripts, exe)
```
layday, line 195: `env_scripts` is relative to `path`:

```suggestion
    executable = os.path.join(env_scripts, exe)
```

src/build/env.py
```diff
@@ -137,78 +144,58 @@ def install(self, requirements):  # type: (Iterable[str]) -> None
...
+    :return: the isolated build environment executable, and the pip to use to install packages into it
+    """
+    import venv
 
-else:
+    venv.EnvBuilder(with_pip=True).create(path)
+    executable, script_dir = _find_executable_and_scripts(path)
+    # avoid the setuptools from ensurepip to break the isolation
+    if sys.version_info[0:2] == (3, 5):  # python3.5 by default comes with pip 9 that's too old, for new standards
+        subprocess.check_call([executable, '-m', 'pip', 'install', '-U', 'pip'])
```
layday, line 175: These commands used to be run in isolated mode (`-I`), why is that not done anymore?

src/build/env.py
```diff
@@ -119,15 +131,10 @@ def install(self, requirements):  # type: (Iterable[str]) -> None
             req_file.write(os.linesep.join(requirements))
```
layday, line 131: A comment should explain why a requirements file is used and why `delete` is set to `False` and the file deleted manually on line 144 both of which are non-obvious.

setup.cfg
layday: Why is `venv` slower?

src/build/__init__.py
layday: Also, why does the `PATH` need to be deduplicated?

src/build/__init__.py
layday, line 161: Public attributes should be annotated:

```suggestion
        self.script_dir = os.path.dirname(python_executable) if scripts_dir is None else scripts_dir  # type: Optional[Union[bytes, Text]]
```

layday: Are we adding the `scripts_dir` to the path when build is run without isolation, or why should it default to `scripts_dir`?

src/build/env.py
gaborbernat: It is. Try to remove and see how the CI fails without it.

src/build/env.py
gaborbernat: Because beforehand the pip could have been global, now it's guaranteed to be within a virtual environment so it's no longer needed.

src/build/env.py
gaborbernat: out of scope for this PR

src/build/__init__.py
gaborbernat: But it's deduced automatically by mypy.

setup.cfg
gaborbernat: see https://virtualenv.pypa.io/en/latest/user_guide.html#seeders, but boils down to virtualenv using a cached seed technique, venv does not

src/build/__init__.py
layday: The public interface should not be inferred and some type checkers like Pyright do not perform inference on `py.typed` packages at all.

gaborbernat: I'm tempted to call this a pyright bug. Isn't Python typing meant to be gradual as per https://www.python.org/dev/peps/pep-0484/#abstract (and https://www.python.org/dev/peps/pep-0483/#summary-of-gradual-typing)?

gaborbernat: If this would hold I'd expect mypy to warn if strict mode is enabled, but it does not 🤔

layday: Gradual typing does not refer to inference - it refers to the fact that typing is opt-in.  If you run mypy in strict mode, gradual typing ceases to be an option.  It's not a bug in Pyright that public attributes must be annotated - it's a conscious design decision on their part.  That mypy doesn't error for untyped public attributes is very likely an omission.

src/build/env.py
layday: `-I` still provides isolation from (a) the cwd and (b) host Python env vars.

src/build/__init__.py
layday: I see that you've changed this to a read-only property which is annotated, collapsing.

gaborbernat:
@layday @FFY00 I think now I've addressed all raised concerns 👍🏻 and CI is passing. @FFY00 do you want to add the changelog now?

layday (approved):
LGTM, thanks for working on this.

gaborbernat:
@FFY00 can we get this out the door?

FFY00 (approved):
Thanks :blush:

gaborbernat:
Considering this is a critical bugfix can we do a release with what we have (0.1.1?), thanks!

FFY00:
Just a FYI, since we now have a changelog you can add an entry to `CHANGELOG.rst` in the PR. The order is supposed to be by relevance. I did it manually now.

FFY00:
> Considering this is a critical bugfix can we do a release with what we have (0.1.1?), thanks!

We can have a new release sure, but because there were a few behavior changes I think it should be `0.2.0`, let's save `0.1.X` for bugfixes.

gaborbernat:
Sure, I generally reserve minor versions for feature additions, not bug fixes, but don't mind.

> The order is supposed to be by relevance.

How do you determine relevance? This feels very subjective and open for debate. I'd suggest something more objective.

FFY00:
> Sure, I generally reserve minor versions for feature additions, not bug fixes, but don't mind.

I am generally inclined to also increment minor versions when there are behavior changes that might break stuff, which is the case with us now resolving dependencies recursively.

> How do you determine relevance? This feels very subjective and open for debate. I'd suggest something more objective.

Yes, it's subjective. I am not very picky here. The goal is to try to make things easier for users, if that requires two more sentences exchanged between maintainers, I am fine with that, if this becomes more we should revisit.

FFY00:
Also, there was a breaking change https://pypa-build.readthedocs.io/en/stable/changelog.html#breaking-changes.

PR #222, opened 2021-02-03
pre-commit: upgrade packages

gaborbernat:


layday (approved):

layday:
We should probably add mypy in pre-commit as well at some point.

FFY00:
No, that was discussed and decided against because it takes some time, so it's not suitable for a git pre-commit hook.

PR #223, opened 2021-02-03
setup.cfg: use mypy 0.800

gaborbernat:


layday:
Looks like #223 has the same change, so we close this one?

gaborbernat:
A mistake adding it there. Now it's only here 👍🏻

layday (approved):

PR #225, opened 2021-02-08
docs: improve the formatting, style and accuracy of docstrings

layday:
Sphinx does not extract docstrings from `__init__` - these
must be placed under the class.


gaborbernat (changes requested):
use instead https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#configuration autoclass both

gaborbernat (approved):

FFY00 (approved):

PR #229, opened 2021-02-10
env: upgrade pip based on pi version (fixes Python 3.6.5-3.8 issues, including in GitHub Actions via pipx)

### LINKED ISSUE #228: pipx run --spec build still broken
I'm still getting the same error as in #226:

```
Run pipx run --spec build==0.2.1 pyproject-build --sdist
Uninstalling setuptools-39.0.1:
  Successfully uninstalled setuptools-39.0.1
25l25h
Usage:   
  /tmp/build-env-5zipwlv4/bin/python -m pip install [options] <requirement specifier> [package-index-options] ...
  /tmp/build-env-5zipwlv4/bin/python -m pip install [options] -r <requirements file> [package-index-options] ...
  /tmp/build-env-5zipwlv4/bin/python -m pip install [options] [-e] <vcs project url> ...
  /tmp/build-env-5zipwlv4/bin/python -m pip install [options] [-e] <local project path> ...
  /tmp/build-env-5zipwlv4/bin/python -m pip install [options] <archive url/path> ...

no such option: --no-warn-script-location
  File "/opt/pipx/.cache/14c4e297c8a7e45/bin/pyproject-build", line 8, in <module>
    sys.exit(entrypoint())
  File "/opt/pipx/.cache/14c4e297c8a7e45/lib/python3.6/site-packages/build/__main__.py", line 210, in entrypoint
    main(sys.argv[1:])
  File "/opt/pipx/.cache/14c4e297c8a7e45/lib/python3.6/site-packages/build/__main__.py", line 206, in main
    build_package(args.srcdir, outdir, distributions, config_settings, not args.no_isolation, args.skip_dependencies)
  File "/opt/pipx/.cache/14c4e297c8a7e45/lib/python3.6/site-packages/build/__main__.py", line 94, in build_package
    _build_in_isolated_env(builder, outdir, distributions, config_settings)
  File "/opt/pipx/.cache/14c4e297c8a7e45/lib/python3.6/site-packages/build/__main__.py", line 56, in _build_in_isolated_env
    env.install(builder.build_dependencies)
  File "/opt/pipx/.cache/14c4e297c8a7e45/lib/python3.6/site-packages/build/env.py", line 143, in install
    subprocess.check_call(cmd)
  File "/usr/lib/python3.6/subprocess.py", line 311, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['/tmp/build-env-5zipwlv4/bin/python', '-Im', 'pip', 'install', '--no-warn-script-location', '-r', '/tmp/build-reqs-pgst_mml.txt']' returned non-zero exit status 2.
```

See here:
https://github.com/scikit-hep/boost-histogram/runs/1868826515?check_suite_focus=true

I can reproduce it locally with these four lines:

```bash
docker run --rm -it ubuntu:18.04
apt-get update && apt-get install -y python3-pip python3-venv
python3 -m pip install pipx
pipx run --spec build pyproject-build
```

henryiii:
Fixes #228

- fix: Upgrade pip for all Python < 3.7
- fix: avoid breaking so badly if pip is too old

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -13,8 +13,15 @@ import tempfile
 from types import TracebackType
 from typing import Iterable, Optional, Tuple, Type
 
+import packaging.version
+
 from ._compat import abstractproperty, add_metaclass
 
+
+if sys.version_info < (3, 8):
+    import importlib_metadata as metadata
+else:
+    from importlib import metadata
 
 try:
     import virtualenv
@@ -169,20 +176,28 @@ def _create_isolated_env_venv(path):  # type: (str) -> Tuple[str, str]
     import venv
 
     venv.EnvBuilder(with_pip=True).create(path)
-    executable, script_dir = _find_executable_and_scripts(path)
-    # avoid the setuptools from ensurepip to break the isolation
-    if sys.version_info < (3, 6, 6):  # python 3.5 up to 3.6.5 come with pip 9 that's too old, for new standards
+    executable, script_dir, purelib = _find_executable_and_scripts(path)
+
+    # Get the version of pip in the environment
+    pip_distribution = next(iter(metadata.distributions(name='pip', path=[purelib])))
+    pip_version = packaging.version.Version(pip_distribution.version)
+
+    # Currently upgrade if Pip 19.1+ not available, since Pip 19 is the first
+    # one to officially support PEP 517, and 19.1 supports manylinux1.
+    if pip_version < packaging.version.Version('19.1'):
         subprocess.check_call([executable, '-m', 'pip', 'install', '-U', 'pip'])
+
+    # Avoid the setuptools from ensurepip to break the isolation
     subprocess.check_call([executable, '-m', 'pip', 'uninstall', 'setuptools', '-y'])
     return executable, script_dir
 
 
-def _find_executable_and_scripts(path):  # type: (str) -> Tuple[str, str]
+def _find_executable_and_scripts(path):  # type: (str) -> Tuple[str, str, str]
     """
     Detect the Python executable and script folder of a virtual environment.
 
     :param path: The location of the virtual environment
-    :return: The Python executable and script folder
+    :return: The Python executable, script folder, and purelib folder
     """
     config_vars = sysconfig.get_config_vars().copy()  # globally cached, copy before altering it
     config_vars['base'] = path
@@ -195,7 +210,11 @@ def _find_executable_and_scripts(path):  # type: (str) -> Tuple[str, str]
     executable = os.path.join(env_scripts, exe)
     if not os.path.exists(executable):
         raise RuntimeError('Virtual environment creation failed, executable {} missing'.format(executable))
-    return executable, env_scripts
+
+    purelib = sysconfig.get_path('purelib', vars=config_vars)
+    if not purelib:
+        raise RuntimeError("Couldn't get environment purelib folder")
+    return executable, env_scripts, purelib
 
 
 __all__ = (
--- a/tests/test_env.py
+++ b/tests/test_env.py
@@ -8,6 +8,8 @@ import sysconfig
 
 import pytest
 
+from packaging.version import Version
+
 import build.env
 
 
@@ -60,6 +62,25 @@ def test_fail_to_get_script_path(mocker):
 
 @pytest.mark.skipif(sys.version_info[0] == 2, reason='venv module used on Python 3 only')
 @pytest.mark.skipif(IS_PYPY3, reason='PyPy3 uses get path to create and provision venv')
+def test_fail_to_get_purepath(mocker):
+    sysconfig_get_path = sysconfig.get_path
+
+    def mock_sysconfig_get_path(path, *args, **kwargs):
+        if path == 'purelib':
+            return ''
+        else:
+            return sysconfig_get_path(path, *args, **kwargs)
+
+    mocker.patch('sysconfig.get_path', side_effect=mock_sysconfig_get_path)
+
+    with pytest.raises(RuntimeError, match="Couldn't get environment purelib folder"):
+        env = build.env.IsolatedEnvBuilder()
+        with env:
+            pass
+
+
+@pytest.mark.skipif(sys.version_info[0] == 2, reason='venv module used on Python 3 only')
+@pytest.mark.skipif(IS_PYPY3, reason='PyPy3 uses get path to create and provision venv')
 def test_executable_missing_post_creation(mocker):
     mocker.patch.object(build.env, 'virtualenv', None)
     original_get_path = sysconfig.get_path
@@ -100,7 +121,9 @@ def test_isolated_env_has_install_still_abstract():
 
 
 @pytest.mark.isolated
-@pytest.mark.skipif(sys.version_info > (3, 6, 5), reason='inapplicable')
-def test_default_pip_is_upgraded_on_python_3_6_5_and_below():
+def test_default_pip_is_never_too_old():
     with build.env.IsolatedEnvBuilder() as env:
-        assert not subprocess.check_output([env.executable, '-m', 'pip', '-V']).startswith(b'pip 9.0')
+        version = subprocess.check_output(
+            [env.executable, '-c', 'import pip; print(pip.__version__)'], universal_newlines=True
+        ).strip()
+        assert Version(version) >= Version('19.1')
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,16 @@ Changelog
 +++++++++
 
 
+Unreleased
+==========
+
+- Upgrade pip based on venv pip version, avoids error from unrecognised pip flag on Debian Python 3.6.5-3.8 (`PR #229`_, Fixes `#228`_)
+
+.. _PR #229: https://github.com/pypa/build/pull/229
+.. _#228: https://github.com/pypa/build/issues/228
+
+
+
 0.2.1 (09-02-2021)
 ==================
 


gaborbernat (approved):

gaborbernat (changes requested):
You'll need to fix the breaking CI

layday:
I suggest passing `env=...` to `subprocess.check_call` with the pip flag instead of mutating the environment globally.

henryiii:
There are three directions to go here:

1. Fix the flag being passed to hide a warning causing a crash. Happing in the second commit here and likely a good idea!
2. Upgrade Pip based on Python version. Looks like, from https://github.com/pypa/packaging-problems/issues/444, it's at _least_ Python 3.8+ in some cases!
3. Upgrade pip based on pip version. I was going to say that might be hard - but we do have access to importlib-metadata in our small number of requirements! How about grabbing the pip version via importlib-metadata?

So 1. is a good idea, but eventually, especially when manylinux1 is dropped this summer, it might be important to be able to detect old pip's and upgrade them. 3. seems more reasonable than 2.?

layday:
>     2\. I really want to know how you get 3.8 + pip 9...

3.8 is apparently available on Ubuntu 18.04 as a separate package.

henryiii:
Wow, that's weird. There's a python3.8 package, and a python3.8-venv package, but it doesn't work unless python3-pip is installed, which of course is the 3.6, 9.0 pip. How is this even possible?

```bash
docker run --rm -it ubuntu:18.04
apt-get update
apt-get install python3.8-venv
python3.8 -m venv venv
# Error, with a rather broken message telling you to install python3-venv (python3-pip is enough)
apt-get install python3-pip
python3.8 -m pip --version
# pip 9.0.1 from /usr/lib/python3/dist-packages (python 3.8)
```

henryiii:
I've fixed the second commit, but I think the first commit should probably be adapted from strategy 2. above to 3. @FFY00?

Looks like import lib.metadata is not installed into the environment, so getting at the pip version might be a bit more manual. I've checked pip 8 and 10, both seem to support `pip.__version__` (10.0 dropped a bunch of public API, but this is still there). How about checking against this using packaging.Version?

henryiii:
Bringing back bad memories of the pip 10+ upgrade on my 16.04 docker image:

```
root@2ea07e6c6c06:/# pip install pip==10.0
Collecting pip==10.0
  Downloading https://files.pythonhosted.org/packages/62/a1/0d452b6901b0157a0134fd27ba89bf95a857fbda64ba52e1ca2cf61d8412/pip-10.0.0-py2.py3-none-any.whl (1.3MB)
    100% |################################| 1.3MB 1.3MB/s
Installing collected packages: pip
  Found existing installation: pip 8.1.1
    Not uninstalling pip at /usr/lib/python2.7/dist-packages, outside environment /usr
Successfully installed pip-10.0.0
You are using pip version 10.0.0, however version 21.0.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
root@2ea07e6c6c06:/# python -c "import pip; print(pip.__version__)"
10.0.0
root@2ea07e6c6c06:/# pip install pip==19.0
  File "/usr/bin/pip", line 9, in <module>
    from pip import main
ImportError: cannot import name main
```

layday:
If we're going to spawn another process I wonder if we shouldn't just upgrade pip indiscriminately.

henryiii:
Commit 3 changes the check to one for pip version, rather than Python version. If you don't want to do this, I can remove it. Or, of course, we could just always upgrade pip, but that would slow it down quite a bit, I think, and would force an extra internet connection even on an up-to-date host pip.

gaborbernat:
> If we're going to spawn another process I wonder if we shouldn't just upgrade pip indiscriminately.

Why spawn an expensive process to get this? Just use os.listdir on sitepackages and look for pip.dist-info. pip update is expensive and may fail behind a network firewall, so better avoid unless must.

henryiii:
Do we have sitepackages for the venv, I guess constructed from `self._path`?? Also, will the venv's pip every be older than the host pip? If not, we could just use `importlib.metadata.version("pip")` in the host process.

gaborbernat:
> Do we have sitepackages for the venv, I guess constructed from `self._path`? Also, will the venv's pip ever be older than the host pip? If not, we could just use `importlib.metadata.version("pip")` in the host process.

The venv pip and the host pip are totally different things and managed separately. So any combination of differences might occure. To get the sitepackage you want to do the same thing we do for scripts but instead use purelib.

henryiii:
PS. I think the answer to my question may be yes, because I've had issues with venvs being made with 20.2 on Big Sur when the host was 20.3.

henryiii:
Solution heavily inspired by https://github.com/pipxproject/pipx/blob/master/src/pipx/venv_inspect.py

src/build/env.py
```diff
@@ -126,6 +134,9 @@ def install(self, requirements):  # type: (Iterable[str]) -> None
         if not requirements:
             return
 
+        env = dict(os.environ)
+        env['PIP_NO_WARN_SCRIP_LOCATION'] = '1'
```
layday, line 138:
```suggestion
        env['PIP_NO_WARN_SCRIPT_LOCATION'] = '1'
```

layday: I suppose we can revert to using the flag now, actually.

henryiii: 😳

henryiii: When this passes, I'll update the changelog and test name.

henryiii: We could, but this is safer, just in case something goes wrong. But I can drop this commit if desired.

henryiii: Enough is passing, will repush

src/build/env.py
```diff
@@ -126,6 +134,9 @@ def install(self, requirements):  # type: (Iterable[str]) -> None
         if not requirements:
             return
 
+        env = dict(os.environ)
```
gaborbernat, line 137: it's better to do os.environ.copy

src/build/env.py
```diff
@@ -169,20 +179,28 @@ def _create_isolated_env_venv(path):  # type: (str) -> Tuple[str, str]
     import venv
 
     venv.EnvBuilder(with_pip=True).create(path)
-    executable, script_dir = _find_executable_and_scripts(path)
-    # avoid the setuptools from ensurepip to break the isolation
-    if sys.version_info < (3, 6, 6):  # python 3.5 up to 3.6.5 come with pip 9 that's too old, for new standards
+    executable, script_dir, site_packages = _find_executable_and_scripts(path)
```
gaborbernat, line 182: this is less site_packages more purelib

src/build/env.py
layday: This might also need to be set to `'0'` for it to have the desired effect, see https://github.com/pypa/pip/issues/5200.

src/build/env.py
henryiii: Sure, will change.

src/build/env.py
```diff
@@ -169,20 +179,28 @@ def _create_isolated_env_venv(path):  # type: (str) -> Tuple[str, str]
...
 
 
-def _find_executable_and_scripts(path):  # type: (str) -> Tuple[str, str]
+def _find_executable_and_scripts(path):  # type: (str) -> Tuple[str, str, str]
     """
     Detect the Python executable and script folder of a virtual environment.
 
     :param path: The location of the virtual environment
-    :return: The Python executable and script folder
+    :return: The Python executable, script folder, and site-packages folder
```
gaborbernat, line 203: purelib is the correct term, not site-packages

src/build/env.py
```diff
@@ -195,7 +213,11 @@ def _find_executable_and_scripts(path):  # type: (str) -> Tuple[str, str]
     executable = os.path.join(env_scripts, exe)
     if not os.path.exists(executable):
         raise RuntimeError('Virtual environment creation failed, executable {} missing'.format(executable))
-    return executable, env_scripts
+
+    lib = sysconfig.get_path('purelib', vars=config_vars)
+    if not lib:
+        raise RuntimeError("Couldn't get environment site-packages path")
```
gaborbernat, line 219: purelib, not site-packages

tests/test_env.py
```diff
@@ -100,7 +99,9 @@ def install(self, requirements):
 
 
 @pytest.mark.isolated
-@pytest.mark.skipif(sys.version_info > (3, 6, 5), reason='inapplicable')
-def test_default_pip_is_upgraded_on_python_3_6_5_and_below():
+def test_default_pip_is_never_too_old():
     with build.env.IsolatedEnvBuilder() as env:
-        assert not subprocess.check_output([env.executable, '-m', 'pip', '-V']).startswith(b'pip 9.0')
+        version = subprocess.check_output([env.executable, '-c', 'import pip; print(pip.__version__)']).strip()
```
gaborbernat, line 104: you don't have to do this dance if you pass universal_newlines=True

src/build/env.py
henryiii: I've been bit by the oddities of NO variables in pip before; would be nice if there was a good way to test this. Maybe by looking for the warning?

tests/test_env.py
```diff
@@ -100,7 +99,9 @@ def install(self, requirements):
...
+        if hasattr(version, 'decode'):
+            version = version.decode()
+        assert int(version.split('.')[0]) >= 10
```
gaborbernat, line 107: Any reason for not using Version as we do in the code for this check?

tests/test_env.py
henryiii: Good, I was going to do encoding=, but that was 3.6+.

tests/test_env.py
henryiii: Not really.

src/build/env.py
gaborbernat: IMHO we can use the flag here 👍🏻

src/build/env.py
henryiii: Oops, missed this one

src/build/env.py
layday: > I've been bit by the oddities of NO variables in pip before; would be nice if there was a good way to test this. Maybe by looking for the warning?

That would require figuring out how to trigger the warning.  If pip uses the same Python executable as the script we're installing, pip does not generate a warning; see https://github.com/pypa/pip/blob/c7419b2aac16c1190b274b9cd7c273d879816ddb/src/pip/_internal/operations/install/wheel.py#L178.

henryiii: I've made this `'0'`. I think, since this is just hiding a warning, the environment variable is a bit better. I would not want to risk breaking just to try to hide a warning. If we require pip 10+ (personally I like 18.1+ due to manylinux2010 support :) ) somewhere else, that's fine, but it probably would be better to break there.

I've still kept this a distinct commit so it's trivial to drop.

layday (approved):
LGTM!

src/build/env.py
```diff
@@ -118,14 +126,17 @@ def install(self, requirements):  # type: (Iterable[str]) -> None
         """
         Install packages from PEP 508 requirements in the isolated build environment.
 
-        :param requirements: PEP 508 requirement specification to install
+        :param requirements: PuP 508 requirement specification to install
```
layday, line 129: Find and replace gone awry? ;)

src/build/env.py
```diff
@@ -118,14 +126,17 @@ def install(self, requirements):  # type: (Iterable[str]) -> None
...
 
         :note: Passing non-PEP 508 strings will result in undefined behavior, you *should not* rely on it. It is
                merely an implementation detail, it may change any time without warning.
         """
         if not requirements:
             return
 
+        env = os.environ.copy()
+        env['PIP_NO_WARN_SCRIPT_LOCATION'] = '0'
```
gaborbernat, line 138: I'd actually like to make this a flag, instead of env-var 🤔 int the spirit of fail early

src/build/env.py
```diff
@@ -169,20 +179,26 @@ def _create_isolated_env_venv(path):  # type: (str) -> Tuple[str, str]
     import venv
 
     venv.EnvBuilder(with_pip=True).create(path)
-    executable, script_dir = _find_executable_and_scripts(path)
-    # avoid the setuptools from ensurepip to break the isolation
-    if sys.version_info < (3, 6, 6):  # python 3.5 up to 3.6.5 come with pip 9 that's too old, for new standards
+    executable, script_dir, purelib = _find_executable_and_scripts(path)
+
+    # Require a minimum version of pip (currently 10)
+    (pip_version,) = (
```
gaborbernat, line 185: I don't think you need the left side parenthesis

src/build/env.py
```diff
@@ -169,20 +179,26 @@ def _create_isolated_env_venv(path):  # type: (str) -> Tuple[str, str]
...
+        Version(d.version) for d in metadata.distributions(path=[purelib]) if canonicalize_name(d.metadata['name']) == 'pip'
```
gaborbernat, line 186: do we need canonicalize_name? I don't think so, pip is same in all formats

src/build/env.py
henryiii: Oops, thought I corrected that! VI user error. :)

src/build/env.py
gaborbernat: also instead of iterating through all can't we just lookup pip?

henryiii:
This might be useful in the future for people like me who push too much ;) : https://github.com/marketplace/actions/cancel-workflow-action (disclaimer: I haven't tried it yet, keep intending too)

tests/test_env.py
```diff
@@ -100,7 +101,9 @@ def install(self, requirements):
 
 
 @pytest.mark.isolated
-@pytest.mark.skipif(sys.version_info > (3, 6, 5), reason='inapplicable')
-def test_default_pip_is_upgraded_on_python_3_6_5_and_below():
+def test_default_pip_is_never_too_old():
     with build.env.IsolatedEnvBuilder() as env:
-        assert not subprocess.check_output([env.executable, '-m', 'pip', '-V']).startswith(b'pip 9.0')
+        version = subprocess.check_output(
+            [env.executable, '-c', 'import pip; print(pip.__version__)'], universal_newlines=True
```
gaborbernat, line 107: I don't think pip supports any format other than CLI. Instead you should do ``python -m pip --version`` and then do a split(' ')[1] on the output.

gaborbernat (changes requested):

gaborbernat:
> This might be useful in the future for people like me who push too much ;) : [github.com/marketplace/actions/cancel-workflow-action](https://github.com/marketplace/actions/cancel-workflow-action) (disclaimer: I haven't tried it yet, keep intending to)

Sadly usage of that action requires exposing Github tokens we're not comfortable providing (it's a security risk).

src/build/env.py
henryiii: I believe Black requires this for readability, the comma is easy to miss. Though I've gotten used to writing it due to thinking Black requires it.

henryiii:
GitHub just added `[skip ci]` and such, so maybe there's hope for a first party skip rebuild eventually. :)

src/build/env.py
gaborbernat: Well also this is wasteful you exhaust the distributions irrespective of where pip is found (first or last). Use instead ``next`` to stop at the first, and then you don't need this tuple unrolling.

src/build/env.py
henryiii: Not sure if there's a simple API to do so, lookup by name doesn't have the ability to select a path. Also, this should be a very short list because it's a new venv?

tests/test_env.py
uranusjr: It might be worthwhile to ask ono pip’s issue tracker about this being an exception. It’s not like there are many obvious ways to read that value…

henryiii: They removed all public interface with the release of 10.0, including the interface the `pip` script used, yet this stayed.

src/build/env.py
uranusjr: Why not use the singular `metadata.distribution(d.metadata['name'], path=[purelib])` instead? There’s also a shortcut `metadata.version(d.metadata['name'], path=[purelib])` if you just want the version, but I didn’t check its availability in the stdlib module.

uranusjr: Oops, I took too long to reply and didn’t see the cot above. It seems like the shortcut functions indeed does not take context kwargs.

uranusjr: This may still be simpler though:

```python
next(metadata.distributions(name=d.metadata['name'], path=[purelib])).version
```

Also note that it is not safe to unpack the iterator directly into a one-element tuple. It is possible for importlib-metadata to return multiple pip copies if it finds e.g. one in system-site-packages and one in user-site.

henryiii:
The failures of the Windows 3.6 and 2.7 tests gives me a sneaking suspicion that more than Pip 10 is likely required on Windows.

src/build/env.py
henryiii: But we are only passing one path, so isn't multiple pip's an error we want to see?

tests/test_env.py
henryiii: I've asked here: https://discuss.python.org/t/is-is-safe-to-use-pip-version/7155

henryiii:
Current status:

* I've required Pip 19.1+, since Pip 19 is the first one to officially support PEP 517, and 19.1 supports manylinux1. There's also 19.3.1, which fixes installing with a missing setup.py, etc. Not sure what the best minimum is, but it is more than 10, since 10 breaks the tests on Windows.
* I've left the change to using an env var since it's (slightly) easier for me to drop the commit than put it back in if you do want it. It's definitely easier to discuss. Assuming @FFY00 would like to weigh in on that?
* Still have the tuple unpacking, so you get an error if there are multiple pip's in the purelib, can change, but it seems like "error early" (drop the first commit) would also go well with forcing just a single result. I also think it reads slightly better.

tests/test_env.py
henryiii: It's allowed. :)

src/build/env.py
gaborbernat: These three lines read to me quite bad 😿 I'd prefer to have a one-liner instead.

henryiii:
I'd like to make an immediate follow-on PR if this goes in. I've noticed `build` break on Big Sur because the venv bundles 20.2, and 20.3 is required for Big Sur (can't handle the name change to 11 before that), and on an M1 (I got mine about an hour ago), you need 21.0.1 to install universal2 wheels (packaging 20.9). So in the follow on PR, I could require a higher version of PIP if on macOS >= 11 and even higher if on an M1 and Python 3.6+.

(This breakage was in 0.1.0 as well)

src/build/env.py
henryiii: Was just hoping to get a little input from @FFY00 so I don't end up having to backtrack, but I guess I could backtrack if needed...

henryiii: The is one line now, though, isn't it?

```python
(pip_version,) = (Version(d.version) for d in metadata.distributions(name='pip', path=[purelib]))
```

gaborbernat: I guess I just don't like using tuple unrolling for one element. Looks odd 🤷‍♂️

gaborbernat: Much prefer next over a generator 👍

henryiii: Now that I have that version recorded above, I can change the one in the PR. ;)

henryiii: Do I really have to stick an `iter()` in here to make mypy happy? Technically, I guess that is correct, you should not iterate over something with `next()` without calling `iter()`.
```
src/build/env.py:184: error: No overload variant of "next" matches argument type "Iterable[Distribution]"
src/build/env.py:184: note: Possible overload variant:
src/build/env.py:184: note:     def [_T] next(Iterator[_T]) -> _T
src/build/env.py:184: note:     <1 more non-matching overload not shown>
```

henryiii: So it's:

```python
pip_version = Version(next(iter(metadata.distributions(name='pip', path=[purelib]))).version)
```

Saves about 4 characters, and has _way_ too many parentheses.

src/build/env.py
FFY00: I assume the move to env var here is because it is present in older versions of pip, where the CLI argument isn't, right? If not, I think the CLI would be preferred.

FFY00 (changes requested):
Just some minor comments and nitpicks. Thank you :blush:

We use the (main) affected component as the commit prefix, could you change the commit prefix to `env:`? Thanks.
Should be fairly straight forward, interactive rebase and reword the commits.

> I've required Pip 19.1+, since Pip 19 is the first one to officially support PEP 517, and 19.1 supports manylinux1. There's also 19.3.1, which fixes installing with a missing setup.py, etc. Not sure what the best minimum is, but it is more than 10, since 10 breaks the tests on Windows.

We should probably document this. Could you add this information as a comment next to the version check?

> I've left the change to using an env var since it's (slightly) easier for me to drop the commit than put it back in if you do want it. It's definitely easier to discuss. Assuming @FFY00 would like to weigh in on that?

I left a comment there, but essentially, if there is no difference, the CLI argument would be preferred.

You should be able to fix this easily with a [fixup commit](https://fle.github.io/git-tip-keep-your-branch-clean-with-fixup-and-autosquash.html) (in case you weren't aware). 

> Still have the tuple unpacking, so you get an error if there are multiple pip's in the purelib, can change, but it seems like "error early" (drop the first commit) would also go well with forcing just a single result. I also think it reads slightly better.

I am not exactly sure what is happening here, could you elaborate a bit?

henryiii:
> I am not exactly sure what is happening here, could you elaborate a bit?

We were discussing:

```python
(pip_version,) = (Version(d.version) for d in metadata.distributions(name='pip', path=[purelib]))
# VS:
pip_version = Version(next(iter(metadata.distributions(name='pip', path=[purelib]))).version)
```

Which do you prefer? If you have a large number of items and are worried about performance, or if there might be more than one, there is a bigger difference, but in this case, it's pretty close to stylistic choice.

FFY00:
Let's go for the iter one as it's faster but let's break it down into chunks to make it easier to read, I particularly have difficulty reading such nested expressions.

```python
pip_version = Version(
    next(iter(
        metadata.distributions(name='pip', path=[purelib]),
    )).version,
)
```

FFY00:
Note the trailing `,` to make black happy.

src/build/env.py
henryiii: Yes. This is what's breaking when trying to use build, for example using Ubuntu 18.04. If we fix the upgrade, this shouldn't break. I'll drop the commit,

henryiii:
I think you mean:

```python
    pip_version = Version(
        next(
            iter(
                metadata.distributions(name='pip', path=[purelib]),
            )
        ).version,
    )
```

Black doens't allow multiple parens on the same line when expanded, it can trick you eyes when reading. But I thought @gaborbernat was complaining about 3 lines, this is 7...

We could do one trailing comma:

```python
    pip_version = Version(
        next(iter(metadata.distributions(name='pip', path=[purelib]))).version,
    )
```

Or we could break this into two lines with an intermediate, which is often how you best appease Black (and future readers).

FFY00:
Ugh, seems black cannot be tricked into formatting as I wanted :upside_down_face: 

Let's put the distribution in a variable then.

```python
pip_distribution = next(iter(metadata.distributions(name="pip", path=[purelib])))
pip_version = Version(pip_distribution.version)
```

henryiii:
I have:

```python
    # Get the version of pip in the environment
    distributions = metadata.distributions(name='pip', path=[purelib])
    pip_version = Version(next(iter(distributions)).version)  # first match version
```

Look okay?

henryiii:
Nah, I like yours better.

henryiii:
Ahhh, blackwards quoted strings are driving me crazy. It's so weird to see blackish code but single quotes.

henryiii:
By default I use https://www.conventionalcommits.org/en/v1.0.0/ unless asked otherwise. :)

src/build/env.py
```diff
@@ -195,7 +210,11 @@ def _find_executable_and_scripts(path):  # type: (str) -> Tuple[str, str]
     executable = os.path.join(env_scripts, exe)
     if not os.path.exists(executable):
         raise RuntimeError('Virtual environment creation failed, executable {} missing'.format(executable))
-    return executable, env_scripts
+
+    purelib = sysconfig.get_path('purelib', vars=config_vars)
+    if not purelib:
+        raise RuntimeError("Couldn't get environment purelib folder")
```
gaborbernat, line 216: This line seems not covered by any test 🤔

gaborbernat (approved):
Otherwise looks ready to go.

henryiii:
How did just adding a Python 3 only test break `test_build[dateutil-isolated-entrypoint]` so that it can't get setuptools > 40 on Python 2? Guessing that's not related?

src/build/env.py
henryiii: Now it is. :)

henryiii:
I got "No matching distribution found for setuptools>=40.8.0" again, but for Windows 3.6, is this a known flake?

Why is this 40.8? Shouldn't it be 42? Setuptools 41 continued to fix a lot of serious bugs with PEP 517 support, and 42 is the [first version to officially support](https://setuptools.readthedocs.io/en/latest/history.html#v42-0-0) PEP 517/518. I've been told by the devs before to pin >=42 in a pyproject.toml, at least. (totally off topic, just noticed that printout)

henryiii:
Note to self: this is my planned followup PR:

```diff
diff --git a/src/build/env.py b/src/build/env.py
index 9d42461..affe6ec 100644
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -182,10 +182,22 @@ def _create_isolated_env_venv(path):  # type: (str) -> Tuple[str, str]
     pip_distribution = next(iter(metadata.distributions(name='pip', path=[purelib])))
     pip_version = Version(pip_distribution.version)

-    # Currently upgrade if Pip 19.1+ not available, since Pip 19 is the first
-    # one to officially support PEP 517, and 19.1 supports manylinux1.
     if pip_version < Version('19.1'):
+        # Currently upgrade if Pip 19.1+ not available, since Pip 19 is the first
+        # one to officially support PEP 517, and 19.1 supports manylinux1.
         subprocess.check_call([executable, '-m', 'pip', 'install', '-U', 'pip'])
+    elif platform.system() == 'Darwin' and platform.mac_ver() > (11, 0):
+        # macOS 11+ needs Pip 20.3+ due to the name scheme change.
+        # Note that macOS 11.0 can be told to report 10.16 for backwards compatibility;
+        # but that also fixes earlier versions of pip so this is only needed for 11+.
+        if pip_version < Version('20.3'):
+            subprocess.check_call([executable, '-m', 'pip', 'install', '-U', 'pip'])
+        elif pip_version < Version('21.0.1') and sys.version_info >= (3, 6) and platfrom.machine() != 'x86_64':
+            # The first version of Pip to fully support Apple Silicon macOS was 21.0.1.
+            subprocess.check_call([executable, '-m', 'pip', 'install', '-U', 'pip'])
+
+
+

     # Avoid the setuptools from ensurepip to break the isolation
     subprocess.check_call([executable, '-m', 'pip', 'uninstall', 'setuptools', '-y'])
```

Just in case I loose it. :)

src/build/env.py
```diff
@@ -169,20 +176,28 @@ def _create_isolated_env_venv(path):  # type: (str) -> Tuple[str, str]
     import venv
 
     venv.EnvBuilder(with_pip=True).create(path)
-    executable, script_dir = _find_executable_and_scripts(path)
-    # avoid the setuptools from ensurepip to break the isolation
-    if sys.version_info < (3, 6, 6):  # python 3.5 up to 3.6.5 come with pip 9 that's too old, for new standards
+    executable, script_dir, purelib = _find_executable_and_scripts(path)
+
+    # Get the version of pip in the environment
+    pip_distribution = next(iter(metadata.distributions(name='pip', path=[purelib])))
```
layday, line 182: Isn't the return value of `metadata.distributions` an iterator or why are we casting it to `iter`?

henryiii: Because it's an `Iteratable`, not an `Iterator`. You have to call `iter()` to get an `Iterator` that you call next with. Yes, it happens to work at runtime to call next on it, because the "Iterator" it returns happens to be itself, but it's technically wrong, and MyPy knows this.

In "real life", you never call `next()` without calling `iter()` first.

layday: Chain results are iterators.  If mypy thinks it's not an iterator, `distributions` is underspecified in the typeshed.

henryiii: It looks like these are correct here: https://github.com/python/typeshed/blob/master/stdlib/itertools.pyi

henryiii: However, this is returning an Iterable, for example, not an Iterator: https://github.com/python/typeshed/blob/27f38b8c23da0a2d0ff9ce486f84664c8b9a2d58/stdlib/importlib/metadata.pyi#L41

layday: Indeed, but it might be intentional, so that people don't rely on it being an iterator going forward.

layday: Let's keep it the way it is :)

henryiii: Where is `importlib<underscore>metadata` typed? Is that same file reused somehow?

henryiii: Okay. Up to you if you think it's worth asking about.

uranusjr:
> I got "No matching distribution found for setuptools>=40.8.0" again, but for Windows 3.6, is this a known flake? It was Windows 2.7 last time.
> 
> Why is this 40.8?

That probably comes from pipx’s own `pyproject.toml`:

https://github.com/pypa/build/blob/7d8f9150ae56c6bc43dc34bcefa30b0c6d03246a/pyproject.toml#L1-L3

If that’s correct, the error basically means it cannot install setuptools, for whatever reason (network issues?)

gaborbernat:
@henryiii note the CI currently is failing, and we can't merge until it passes.

layday:
The error from pip when attempting to install setuptools during one of the integration tests was:

> `WARNING: Skipping page https://pypi.org/simple/setuptools/ because the GET request got Content-Type: .The only supported Content-Type is text/html`
> `ERROR: Could not find a version that satisfies the requirement setuptools>=40.8.0 (from -r c:\users\runner~1\appdata\local\temp\build-reqs-c0aiyf.txt (line 3)) (from versions: none)`

I've trouble imagining how it might've been caused by any of the changes in this PR.

FFY00:
Yes. I retrigered.

henryiii:
I think it's passing?

src/build/env.py
```diff
@@ -13,9 +13,16 @@
 from types import TracebackType
 from typing import Iterable, Optional, Tuple, Type
 
+from packaging.version import Version
```
FFY00, line 16: I hadn't caught this. No partial import unless for type hints.
I will fix this and push to your branch to avoid bothering you even more :upside_down_face:

FFY00 (approved):

tests/test_env.py
```diff
@@ -58,6 +60,25 @@ def test_fail_to_get_script_path(mocker):
...
+@pytest.mark.skipif(sys.version_info[0] == 2, reason='venv module used on Python 3 only')
+@pytest.mark.skipif(IS_PYPY3, reason='PyPy3 uses get path to create and provision venv')
+def test_fail_to_get_purepath(mocker):
+    sysconfig_get_path = sysconfig.get_path
+
+    def mock_sysconfig_get_path(path, *args, **kwargs):
+        if path == 'purelib':
+            return ''
+        else:
+            return sysconfig_get_path(path, *args, **kwargs)
```
gaborbernat, line 72: You don't need the else branch here, and I think you can even make it a one-liner 🤔

layday: eh, I don't think it really matters and 'implicit' `else`s throw me off, personally.

henryiii:
Code coverage seems to be complaining about missing coverage in tests? Shouldn't the coverage be for the package, not the tests? Especially conftest.py, coverage metrics don't work well in tests. I've been a bit hap-hazard in my own coverage testing, so not being critical at all, just wondering if it would be better to limit the coverage measurement to just the package.

tests/test_env.py
henryiii: It's also bad Typing, you'd be implicitly returning `str | None` instead of `str`. I thought about doing `return '' if path == 'purelib' else sysconfig_get_path(path, *args, **kwargs)` but decided against it, this looked more readable I thought.

FFY00:
I made the same argument previously but we left it in to make sure we notice when something we think is running isn't.

henryiii:
@FFY00 any chance we could get a release? Several important bug fixes are in now, incl this one.

PR #230, opened 2021-02-12
env: fix issue installing wheels on Big Sur

henryiii:
Followup to #229. This fixes an issue I've seen trying to install `cmake` on Big Sur (Intel), and preemptively fixes an issue for Big Sur (Apple Silicon).

Edit: I'm now questioning if I got the version right for pip breaking on Big Sur Intel...
EditEdit: Ahh, I have a custom wheel cached as workaround. 🤦 Yes, 20.2 does break, as promised. 

To see the problem, `pyenv install 3.9.1`, activate it, install build, then build the following `pyproject.toml`:

```toml
[build-system]
requires = ["flit_core >=2,<4", "cmake"]
build-backend = "flit_core.buildapi"

[tool.flit.metadata]
module = "foobar"
author = "Sir Robin"
author-email = "robin@camelot.uk"
home-page = "https://github.com/sirrobin/foobar"
```

You will see `Building wheel for cmake (PEP 517)` because it's using the venv's pip 20.2.

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -11,7 +11,7 @@ import sysconfig
 import tempfile
 
 from types import TracebackType
-from typing import Any, Iterable, Optional, Tuple, Type, cast
+from typing import Iterable, Optional, Tuple, Type
 
 import packaging.version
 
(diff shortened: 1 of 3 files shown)


henryiii:
I can't test this locally anymore because brew's Python 3.9 now using pip 21.0.1. I can see what pyenv 3.9.1 comes with, I suppose. I know for a long time I was getting 20.2.x and nothing depending on `cmake` worked, because it couldn't get the wheel.

henryiii:
I'm getting a compilation from mypy 2.7, call to untyped function from a typed context for https://docs.python.org/2.7/library/platform.html#platform.mac_ver . (Which is very much there in 2.7's docs and at runtime). It's happy on 3.8 mode, though.

layday:
It's just not been annotated in the typeshed, see https://github.com/python/typeshed/blob/27f38b8c23da0a2d0ff9ce486f84664c8b9a2d58/stdlib/%40python2/platform.pyi#L23.

henryiii:
Works locally:

```bash
pip install git+https://github.com/henryiii/build@env/macOS
python -m build
```


```
Collecting pip
  Using cached pip-21.0.1-py3-none-any.whl (1.5 MB)
Installing collected packages: pip
  Attempting uninstall: pip
    Found existing installation: pip 20.2.3
    Uninstalling pip-20.2.3:
      Successfully uninstalled pip-20.2.3
Successfully installed pip-21.0.1
Found existing installation: setuptools 49.2.1
Uninstalling setuptools-49.2.1:
  Successfully uninstalled setuptools-49.2.1
Collecting flit_core<4,>=2
  Using cached flit_core-3.0.0-py3-none-any.whl (36 kB)
Collecting cmake
  Using cached cmake-3.18.4.post1-py3-none-macosx_10_6_x86_64.whl (42.8 MB)
Collecting pytoml
```

As noted in the description above, you need official CPython 3.9.1 or earlier, the homebrew one is patched to include the latest pip in venv (probably to solve the Apple Silicon issue, I'd guess).

henryiii:
Thoughts for a (nice) way to work around it for now? I can't add `type: ignore`, because that causes an error for 3.8 where the annotation is given.

hauntsaninja:
You could add a redundant sys.version check, one with type ignore, one without. Or maybe `cast(Any, platform.mac_ver)()` will work.

henryiii:
I already tried:

```python
        if sys.version_info < (3,):
            platform_mac_ver = platform.mac_ver()  # type: ignore
        else:
            platform_mac_ver = platform.mac_ver()
```

I thought that was the "not-nice" workaround, but turns out it doesn't even work. In 3.8 mode, it still complains. Before that, I tried a cast, which also failed (though I didn't try Any, I put the whole tuple thing in, but assume the result should be the same).


```bash
mypy --python-version 3.8  src/
src/build/env.py:198: error: unused 'type: ignore' comment
```

hauntsaninja:
No, you want to try `cast(Any, platform.mac_ver)()`, not `cast(Any / Tuple, platform.mac_ver())`

henryiii:
Yes, just found that out. :) Works now.

henryiii:
I didn't see the parenthesis in your example, you had it right. :)

setup.cfg
```diff
@@ -83,6 +83,7 @@ extend-ignore = E203
 [mypy]
 ignore_missing_imports = True
 strict = True
+files = src
```
henryiii, line 86: Oops, didn't mean to commit this. Though it is handy for quick running (I usually add mypy to pre-commit)

gaborbernat (changes requested):
@henryiii you still need to add tests for this 🤔

src/build/env.py
```diff
@@ -182,9 +182,20 @@ def _create_isolated_env_venv(path):  # type: (str) -> Tuple[str, str]
...
-    # Currently upgrade if Pip 19.1+ not available, since Pip 19 is the first
-    # one to officially support PEP 517, and 19.1 supports manylinux1.
+    needs_pip_upgrade = False
     if pip_version < packaging.version.Version('19.1'):
+        # PEP-517 and manylinux1 was first implemented in 19.1
+        needs_pip_upgrade = True
+    elif platform.system() == 'Darwin':
+        # macOS 11+ name scheme change requires 20.3. Intel macOS 11.0 can be told to report 10.16 for backwards
+        # compatibility; but that also fixes earlier versions of pip so this is only needed for 11+.
+        if int(cast(Any, platform.mac_ver)()[0].split('.')[0]) >= 11:
```
layday, line 192: Let's split this up:

```suggestion
        # ``platform.mac_ver`` is untyped for Python 2 in the typeshed
        #   TODO: remove when upgrading mypy from 0.800
        platform_mac_ver = platform.mac_ver  # type: Any
        mac_major_ver = int(platform_mac_ver()[0].split('.')[0])
        if mac_major_ver >= 11:
```

gaborbernat: I disagree, so @FFY00 will need to decide.

FFY00: I would actually prefer to put a Python version check and `# type: ignore` as that allows us to easily identify the code to be removed when we drop Python 2 and still does the proper checks on Python 3 without any impact.

henryiii: See my comment above, I tried that first and MyPy doesn't seem to handle it correctly.

henryiii: Though maybe a blend of the two? Python version check and cast? (Also, this will go away in the next version of MyPy)

FFY00: Hum, what if you put the comment in the else rather than the assign? Anyway, even if that doesn't work, we can still only do the custom handling there on Python 2.

henryiii: Doesn't like that ignore at all. Tried a couple of places. I've put the cast in only for Python 2, and have a blend of the two versions above. ;)

FFY00: Thanks.

src/build/env.py
```diff
@@ -182,9 +182,27 @@ def _create_isolated_env_venv(path):  # type: (str) -> Tuple[str, str]
...
+        # macOS 11+ name scheme change requires 20.3. Intel macOS 11.0 can be told to report 10.16 for backwards
+        # compatibility; but that also fixes earlier versions of pip so this is only needed for 11+.
+
+        #   TODO: remove when upgrading mypy from 0.800 or dropping Python 2
+        if sys.version_info >= (3,):
+            platform_mac_ver = platform.mac_ver
+        else:
+            platform_mac_ver = cast(Any, platform.mac_ver)
+
+        if int(platform_mac_ver()[0].split('.')[0]) >= 11:
```
FFY00, line 199: Could we save the string in `platform_mac_ver` instead of the callable? I fell that is a bit clearer.

src/build/env.py
```diff
@@ -182,9 +182,27 @@ def _create_isolated_env_venv(path):  # type: (str) -> Tuple[str, str]
...
+        needs_pip_upgrade = True
+    elif platform.system() == 'Darwin':
+        # macOS 11+ name scheme change requires 20.3. Intel macOS 11.0 can be told to report 10.16 for backwards
+        # compatibility; but that also fixes earlier versions of pip so this is only needed for 11+.
+
+        #   TODO: remove when upgrading mypy from 0.800 or dropping Python 2
+        if sys.version_info >= (3,):
+            platform_mac_ver = platform.mac_ver()
+        else:
+            platform_mac_ver = cast(Any, platform.mac_ver)()
```
henryiii, line 197: Ironically, since Python 2 does't have `venv`, this never runs on python 2, so misses coverage. :P

henryiii:
By the way, this only implements this for venv, not virtualenv, matching the old behavior. Is there a reason there wasn't done for virtualenv before? If it was because virtualenv almost always had 10+, that might not be good enough for macOS. If it's because a user can upgrade virtualenv, and it's not build's fault if virtualenv is too old, then that's okay to leave off?

src/build/env.py
layday: I would simply nest both `_create_isolated_env_venv` and `_find_executable_and_scripts` under an `if sys.version_info[0] > 2` and remove this kludge.

layday: The only reason this wasn't caught earlier is because `ignore_missing_imports` is set to true and `venv` is given a type of `Any` in Python 2.

henryiii: This won't keep us from writing a similar kludge if virtualenv needs to be supported too, but it would clean this up.

gaborbernat:
> By the way, this only implements this for venv, not virtualenv, matching the old behavior. Is there a reason there wasn't done for virtualenv before? If it was because virtualenv almost always had 10+, that might not be good enough for macOS. If it's because a user can upgrade virtualenv, and it's not build's fault if virtualenv is too old, then that's okay to leave off?

virtualenv has smart auto-upgrade functionality, https://virtualenv.pypa.io/en/latest/user_guide.html#wheels, so it is very hard for it to pull old pip versions. Note virtualenv is also a lot faster-creating build environments, so if you want some speed-up install `build[virtualenv]` and avoid a lot of these problems.

gaborbernat (approved):

FFY00:
@gaborbernat please no merge commits, rebase instead :blush:

I will fix it.

gaborbernat:
Just fixed it.

FFY00:
Thanks! While you're at it, could you prefix the `More compact implementation` commit? Should be fairly simple :grin:

gaborbernat:
Done.

henryiii:
@FFY00 We should probably get a release out soon, #229 especially is needed, causes an issue per day or so in packaging issues.

gaborbernat:
> @FFY00 We should probably get a release out soon, #229 especially is needed, causes an issue per day or so in packaging issues.

How many people have M1 chips 😮

henryiii:
> > @FFY00 We should probably get a release out soon, #229 especially is needed, causes an issue per day or so in packaging issues.
> 
> 
> 
> How many people have M1 chips 😮 

#229 was for Ubuntu 18.04. This one was for Big Sur (Intel or M1). And I have one. :)

PR #233, opened 2021-02-13
tox: update docs environment to Python 3.9

FFY00:

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -57,7 +57,7 @@ commands =
 
 [testenv:docs]
 description = build documentations
-basepython = python3.8
+basepython = python3.9
 extras =
     docs
 commands =


gaborbernat (approved):

webknjaz (commented):
How about also updating `.readthedocs.yml`?

webknjaz:
FYI I noticed that the docs job fails for an unrelated reason. This PR https://github.com/pypa/build/pull/235 should fix that.

layday:
Does RTD support 3.9? Otherwise we should probably keep them in sync?

gaborbernat:
https://docs.readthedocs.io/en/stable/config-file/v2.html#build-image says not yet

gaborbernat (changes requested):
Yeah, actually wait with this PR until readthedocs supports it.

webknjaz:
> [docs.readthedocs.io/en/stable/config-file/v2.html#build-image](https://docs.readthedocs.io/en/stable/config-file/v2.html#build-image) says not yet

Those docs are often out-of-date but it seems you're right: it was added to the testing image just two weeks ago — https://github.com/readthedocs/readthedocs-docker-images/commit/3c6e81fa67d8cfd833fa86faf2d138f6e992ea16.

PR #235, opened 2021-02-13
docs: fix sphinx syntax @ `--config-setting` doc

webknjaz:
RST treats single backticks as `:any:` role that tries
to automatically match something with that name. This
generates a warning because there's nothing to be found.
But running Sphinx in nitpicky mode turns the warning
into an error.

There's two ways of fixing this: either use double backticks
for inline code or drop them and maybe switch to double-
quotes. This change does the former.

### DIFF
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -160,7 +160,7 @@ def main_parser():  # type: () -> argparse.ArgumentParser
         '-C',
         action='append',
         help='pass options to the backend.  options which begin with a hyphen must be in the form of '
-        '`--config-setting=--opt(=value)` or `-C--opt(=value)`',
+        '``--config-setting=--opt(=value)`` or ``-C--opt(=value)``',
     )
     return parser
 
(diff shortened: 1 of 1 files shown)


src/build/__main__.py
```diff
@@ -160,7 +160,7 @@ def main_parser():  # type: () -> argparse.ArgumentParser
         '-C',
         action='append',
         help='pass options to the backend.  options which begin with a hyphen must be in the form of '
-        '`--config-setting=--opt(=value)` or `-C--opt(=value)`',
+        '``--config-setting=--opt(=value)`` or ``-C--opt(=value)``',
```
webknjaz, line 163: @FFY00 this is an alternative solution:
```suggestion
        '"--config-setting=--opt(=value)" or "-C--opt(=value)"',
```
(you can accept the change and merge it if you prefer it or just leave it as is, that's up to you)

gaborbernat (approved):

layday:
Thanks for the fix.

layday:
Ah, I just noticed this was two commits - should've squashed and updated the commit message before merging - apologies.

PR #236, opened 2021-02-13
docs: use sphinx-argparse-cli

gaborbernat:
This a CLI renderer I've created myself: https://github.com/gaborbernat/sphinx-argparse-cli#sphinx-argparse-cli

IMHO this renders a more compact and functional (each option generates a reference/anchor) documentation of the CLI.

New format: https://pypa-build--236.org.readthedocs.build/en/236/
Old format: https://pypa-build--235.org.readthedocs.build/en/235/


layday (approved):
LGTM, although I'm wondering if it would be possible to distinguish flags from options with arguments in the optional argument list?

gaborbernat:
Yes, we could, but this is in line with how the old renderer did not, and also how argparse keeps them together. That being said we definitely could identify flags and define a new group for them.

layday:
I was thinking about just showing a placeholder after the option, e.g. `--outdir, -o <outdir>` instead of `--outdir, -o`, to tell them apart.  This is how it's done in the CLI as well.

gaborbernat:
Great idea, will do so 👍

gaborbernat:
@layday done.

FFY00 (approved):

PR #237, opened 2021-02-13
misc: add @layday to the code owners

gaborbernat:
We should have done this when we invited @layday to be a maintainer 👏🏻

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-*       @FFY00 @gaborbernat
+*       @FFY00 @gaborbernat @layday


FFY00:
Eh, it's up to them. If they want to be always pinged for review, they can add themselves. If they just want to be pinged on a certain part of the code, they can do that too.

cc @henryiii

FFY00 (approved):

layday:
Happy to be added to owners, thanks.

henryiii:
> cc @henryiii

I'm not going to be much use for the next two weeks 👶, maybe after that?

henryiii:
The branch is sitting around still, would the auto-delete branches on merges option be a good idea, to keep things tidy?

PR #241, opened 2021-02-14
env: old virtualenv should not be used

henryiii:
Would have helped with https://github.com/pypa/packaging-problems/issues/450. I'm not going to have time to follow up on this, my daughter will be born in 24 or so hours. But I think the general idea is good? Hoping someone else might have a chance?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,9 +6,11 @@ Changelog
 Unreleased
 ==========
 
-- Upgrade pip based on venv pip version, avoids error from unrecognised pip flag on Debian Python 3.6.5-3.8 (`PR #229`_, Fixes `#228`_)
+- Upgrade pip based on venv pip version, avoids error on Debian Python 3.6.5-3.8 or issues installing wheels on Big Sur (`PR #229`_, `PR #230`_, Fixes `#228`_)
 - Build dependencies in isolation, instead of in the build environment (`PR #232`_, Fixes `#231`_)
+- Fallback on venv if virtualenv is too old (`PR 241`_)
 
+.. _PR #241: https://github.com/pypa/build/pull/241
 .. _PR #230: https://github.com/pypa/build/pull/230
 .. _PR #229: https://github.com/pypa/build/pull/229
 .. _#228: https://github.com/pypa/build/issues/228
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -13,7 +13,10 @@ import tempfile
 from types import TracebackType
 from typing import Iterable, Optional, Tuple, Type
 
+import packaging.requirements
 import packaging.version
+
+import build
 
 from ._compat import abstractproperty, add_metaclass
 
@@ -25,8 +28,8 @@ else:
 
 try:
     import virtualenv
-except ImportError:  # pragma: no cover
-    virtualenv = None  # pragma: no cover
+except ImportError:
+    virtualenv = None
 
 
 @add_metaclass(abc.ABCMeta)
@@ -56,8 +59,22 @@ class IsolatedEnv(object):
 class IsolatedEnvBuilder(object):
     """Builder object for isolated environments."""
 
+    _has_virtualenv = None  # type: Optional[bool]
+
     def __init__(self):  # type: () -> None
         self._path = None  # type: Optional[str]
+
+    def _should_use_virtualenv(self):  # type: () -> Optional[bool]
+        # virtualenv might be incompatible if it was installed separately
+        # from build. This verifies that virtualenv and all of its
+        # dependencies are installed as specified by build.
+        if self._has_virtualenv is None:
+            self.__class__._has_virtualenv = virtualenv is not None and not any(
+                packaging.requirements.Requirement(d[1]).name == 'virtualenv'
+                for d in build.check_dependency('build[virtualenv]')
+                if len(d) > 1
+            )
+        return self._has_virtualenv
 
     def __enter__(self):  # type: () -> IsolatedEnv
         """
@@ -67,8 +84,8 @@ class IsolatedEnvBuilder(object):
         """
         self._path = tempfile.mkdtemp(prefix='build-env-')
         try:
-            # use virtualenv on Python 2 (no stdlib venv) or when virtualenv is available (as it's faster than venv)
-            if sys.version_info[0] == 2 or virtualenv is not None:
+            # use virtualenv on Python 2 or when valid virtualenv is available (as it's faster than venv)
+            if sys.version_info < (3,) or self._should_use_virtualenv():
                 executable, scripts_dir = _create_isolated_env_virtualenv(self._path)
             else:
                 executable, scripts_dir = _create_isolated_env_venv(self._path)


src/build/env.py
```diff
@@ -23,8 +23,13 @@
...
 
+
 try:
-    import virtualenv
+    virtualenv_version = packaging.version.Version(metadata.version('virtualenv'))
+    if virtualenv_version < packaging.version.Version('20.0.35'):
+        virtualenv = None
+    else:
+        import virtualenv
 except ImportError:  # pragma: no cover
```
henryiii, line 33: This will be a `importlib.metadata.PackageNotFoundError`, but that has `ImportError` in its mro, and if anything really bad happened and `import virtualenv` failed via `ImportError`, this would still catch it, I suppose? it could be narrowed down if preferred. Not sure why there are pragma no covers here, seems like this could have been testable.

gaborbernat (changes requested):
I don't agree that we should add this unless you want to do so for all specifications within https://github.com/pypa/build/blob/master/setup.cfg#L30-L35

IMHO this is an installer issue, and if the users install with an installer that can't validate this, then the fix is on improving their installer.

henryiii:
No, this is special. If you have Python 3 and install build and have an old version of virtualenv, build will be broken because it tries to use the any present version automatically. Nobody cares if any other `extras` like typing is not correctly installed.

Technically, to match the `install_requires` spec, `virtualenv` should not be used if Python >= 3.

gaborbernat:
> Technically, to match the `install_requires` spec, `virtualenv` should not be used if Python >= 3.

Well only if the user did not install the ``virtualenv`` extra, but there's no way to tell that.

src/build/env.py
```diff
@@ -23,8 +23,13 @@
...
     from importlib import metadata
 
+
 try:
-    import virtualenv
+    virtualenv_version = packaging.version.Version(metadata.version('virtualenv'))
+    if virtualenv_version < packaging.version.Version('20.0.35'):
+        virtualenv = None
+    else:
+        import virtualenv
```
gaborbernat, line 32: Maybe we should make this instead of a cached function call, easier to test. Either add the pragma no coverage or additional tests. Will need a changelog too.

gaborbernat (changes requested):

gaborbernat:
> my daughter will be born in 24 or so hours.

Swift birth and healthy baby!

src/build/env.py
layday: This is only the case in importlib_metadata 3 which dropped support Python 2.7 or 3.5, and which we continue to support.

```
In [1]: !python -m pip install -U 'importlib_metadata==2.*'
Collecting importlib_metadata==2.*
  Downloading importlib_metadata-2.1.1-py2.py3-none-any.whl (10 kB)
Collecting zipp>=0.5
  Downloading zipp-3.4.0-py3-none-any.whl (5.2 kB)
Installing collected packages: zipp, importlib-metadata
Successfully installed importlib-metadata-2.1.1 zipp-3.4.0

In [2]: import importlib_metadata

In [3]: try:
   ...:     importlib_metadata.version('foo')
   ...: except Exception as e:
   ...:     e_ = e
   ...:

In [4]: e_
Out[4]: importlib_metadata.PackageNotFoundError('foo')

In [5]: type(e_).__mro__
Out[5]:
(importlib_metadata.PackageNotFoundError,
 FileNotFoundError,
 OSError,
 Exception,
 BaseException,
 object)
```

I suggest restructuring it as:

```py
try:
    import virtualenv
except ImportError:
    virtualenv = None
else:
    virtualenv_version = packaging.version.Version(metadata.version('virtualenv'))
    if virtualenv_version < packaging.version.Version('20.0.35'):
        virtualenv = None
```

This check should not be performed at import time.

henryiii:
I'm confused; the last few PRs seem to have been merged to master, but there's a main now _too_. Shouldn't master be renamed to main, not dual branches? GitHub even has special support if you delete master and add a new default branch (for example, by renaming): https://github.com/github/renaming

You can retarget a PR to a new branch.

CHANGELOG.rst
```diff
@@ -6,9 +6,11 @@ Changelog
 Unreleased
 ==========
 
-- Upgrade pip based on venv pip version, avoids error from unrecognised pip flag on Debian Python 3.6.5-3.8 (`PR #229`_, Fixes `#228`_)
+- Upgrade pip based on venv pip version, avoids error on Debian Python 3.6.5-3.8 or issues installing wheels on Big Sur (`PR #229`_, `PR #230`_, Fixes `#228`_)
 - Build dependencies in isolation, instead of in the build environment (`PR #232`_, Fixes `#231`_)
+- Fallback on venv if virtualenv is too old (`PR 241`_)
 
+.. _PR #241: https://github.com/pypa/build/pull/241
 .. _PR #230: https://github.com/pypa/build/pull/230
```
henryiii, line 14: The changelog entry seems to have been lost when rebasing #230, even though this ref is still here. I've added it back above (in a slightly different form).

henryiii:
The issue this fixes (https://github.com/pypa/packaging-problems/issues/450) is in a different repo, I could cross-link, or I could transfer the issue to this repo, if you want.

src/build/env.py
```diff
@@ -23,10 +23,15 @@
...
 try:
     import virtualenv
-except ImportError:  # pragma: no cover
-    virtualenv = None  # pragma: no cover
+except ImportError:
+    virtualenv = None
+else:
+    virtualenv_version = packaging.version.Version(metadata.version('virtualenv'))
+    if virtualenv_version < packaging.version.Version('20.0.35'):
+        virtualenv = None  # pragma: no cover
```
FFY00, line 34: I am not too happy with checking the virtualenv version directly. Could we please use `build.check_dependency` to validate `build[virtualenv]`?

FFY00 (changes requested):

src/build/env.py
layday: I would also prefer that this wasn't done on import.  Can we put it inside a function which is then cached when the isolated env builder is first called?

henryiii: If we bump the versions, say `build==0.3.0` suddenly requires `virtualenv>=21`, would `build[virtualenv]` correctly resolve to the current version of build's requirements? And also, does this always pull from the local version or does it need a connection to the package index (might also answer the question above).

layday: `check_dependency` is a thin wrapper around `importlib.metadata` and `packaging`:

```
In [1]: from build import check_dependency

In [2]: list(check_dependency('build[virtualenv]'))
Out[2]: [('build[virtualenv]', 'virtualenv>=20.0.35; extra == "virtualenv"')]

In [3]: !python -m pip install virtualenv
Collecting virtualenv
  Using cached virtualenv-20.4.2-py2.py3-none-any.whl (7.2 MB)
Requirement already satisfied: six<2,>=1.9.0 in ./venv/lib/python3.9/site-packages (from virtualenv) (1.15.0)
Requirement already satisfied: appdirs<2,>=1.4.3 in ./venv/lib/python3.9/site-packages (from virtualenv) (1.4.4)
Requirement already satisfied: filelock<4,>=3.0.0 in ./venv/lib/python3.9/site-packages (from virtualenv) (3.0.12)
Requirement already satisfied: distlib<1,>=0.3.1 in ./venv/lib/python3.9/site-packages (from virtualenv) (0.3.1)
Installing collected packages: virtualenv
Successfully installed virtualenv-20.4.2

In [4]: list(check_dependency('build[virtualenv]'))
Out[4]: []

In [5]: !python -m pip install -U 'virtualenv<20.0.35'
Collecting virtualenv<20.0.35
  Downloading virtualenv-20.0.34-py2.py3-none-any.whl (4.9 MB)
     |████████████████████████████████| 4.9 MB 2.8 MB/s
Requirement already satisfied: filelock<4,>=3.0.0 in ./venv/lib/python3.9/site-packages (from virtualenv<20.0.35) (3.0.12)
Requirement already satisfied: six<2,>=1.9.0 in ./venv/lib/python3.9/site-packages (from virtualenv<20.0.35) (1.15.0)
Requirement already satisfied: distlib<1,>=0.3.1 in ./venv/lib/python3.9/site-packages (from virtualenv<20.0.35) (0.3.1)
Requirement already satisfied: appdirs<2,>=1.4.3 in ./venv/lib/python3.9/site-packages (from virtualenv<20.0.35) (1.4.4)
Installing collected packages: virtualenv
  Attempting uninstall: virtualenv
    Found existing installation: virtualenv 20.4.2
    Uninstalling virtualenv-20.4.2:
      Successfully uninstalled virtualenv-20.4.2
Successfully installed virtualenv-20.0.34

In [6]: list(check_dependency('build[virtualenv]'))
Out[6]: [('build[virtualenv]', 'virtualenv>=20.0.35; extra == "virtualenv"')]
```

uranusjr: `importlib.metadata` only reads locally installed distributions (specifically, it only checks stuffs in `sys.path`). The overhead should be pretty low, but a lazy implementation is still a good idea IMO.

src/build/env.py
henryiii: I've cached it on the env builder. I think it should have been cached across calls to IsolatedEnvBuilder though, looking at the usage in `__main__`.

henryiii: Is d52edca9151da58da2ac89219638bd7355da949c more like what you were thinking?

gaborbernat (approved):

src/build/env.py
```diff
@@ -56,9 +58,15 @@ def install(self, requirements):  # type: (Iterable[str]) -> None
...
     """Builder object for isolated environments."""
 
+    _use_virtualenv = None  # type: Optional[bool]
+
     def __init__(self):  # type: () -> None
         self._path = None  # type: Optional[str]
 
+        if self._use_virtualenv is None:
+            unmet_deps = build.check_dependency('build[virtualenv]')
+            self.__class__._use_virtualenv = len(tuple(unmet_deps)) == 0
```
layday, line 68: This does not only check that virtualenv is installed; it checks that all dependencies are installed. Is that what we want to do?

gaborbernat: I think so. You don't want to run virtualenv when its dependencies are not satisfied, not?

layday: It checks that all of *build*'s dependencies are satisfied as well as virtualenv's.

layday:
I've made some (I hope) minor improvements, factoring the check out into a separate function, verifying that the `virtualenv` variable is not `None` before attempting to inspect the package metadata and stopping after the first match.

gaborbernat (approved):

henryiii:
@FFY00 any final comment? Okay to merge?

FFY00 (approved):

PR #247, opened 2021-03-03
setup.cfg: support direct usage from pipx run in 0.16.1.0+

henryiii:
Pipx has added a feature at our request to run packages with differently named entrypoints without having to use `--spec`. This is before:

```bash
pipx run --spec build pyproject-build
```

and this will work after this PR with Pipx 0.16.1.0+:

```bash
pipx run build
```

See pipxproject/pipx#600.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -40,6 +40,8 @@ package_dir =
 [options.entry_points]
 console_scripts =
     pyproject-build = build.__main__:entrypoint
+pipx.run =
+    build = build.__main__:entrypoint
 
 [options.extras_require]
 docs =
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,15 @@ Changelog
 +++++++++
 
 
+Unreleased
+==========
+
+- Support direct usage from pipx run in 0.16.1.0+ (`PR #247`_)
+
+.. _PR #247: https://github.com/pypa/build/pull/217
+
+
+
 0.3.0 (19-02-2021)
 ==================
 


henryiii:
CC @uranusjr

FFY00 (approved):

FFY00:
Maybe use the `setup.cfg` prefix in the commit to match the history, other than that :+1:

layday (approved):

henryiii:
Also added a changelog entry since I have the PR number now.

uranusjr (approved):

setup.cfg
```diff
@@ -40,6 +40,8 @@ package_dir =
 [options.entry_points]
 console_scripts =
     pyproject-build = build.__main__:entrypoint
+pipx.run =
+    build = build.__main__:entrypoint
```
gaborbernat, line 44: IMHO we should revert this. We should absolutely not expose different executables from different tools... That's the surest path to user confusion.

We already established we can't use the build name because that would shadow the existing build commands (there are quite a few out there).

uranusjr: `pipx.run` entry points are “namespaced” by name, and a `build` entry point here is only accessible with

```
pipx run build  # Run the "build" entry point on the package "build".
pipx run build --spec "another-package"  # Run the "build" entry point on "another-package".
```

So there is no way this can shadow another package’s `build` command.

gaborbernat: > pipx run build

I would have personally preferred pyproject-build here to be inline with the cli...

gaborbernat: So what's the point of pipx run over calling the console script directly? I don't need to add local to path?

layday: `pipx run foo` creates a temporary environment "cached for up to 2 weeks", installs the "foo" package from PyPI and calls the "foo" entry point.  If the pipx-specific entry point were `pyproject-build`, `pipx run` would have to be invoked with `--spec build pyproject-build`, just like the script, which would render it completely redundant.

henryiii: The command is: `pipx run <package>`. This is roughly the equivalent to the following commands in a temporary environment: `pip install build && build`. However, "build" isn't an executable, so this fails. The only way around in the past was to run `pipx run --spec build pyproject-build`, which is long and ugly, and forces you to know both the 'build' name _and_ the 'pyproject-build' name. Before the recent release, there was not a way to customize pipx run, so that was the best we could do. Now, we can add an entry point for pipx.run, and it will use this instead of calling `build` when you try to run the "build" entry point inside pipx run.

`pipx run pyproject-build`, besides being longer and much harder to remember, also would require that a package exist on PyPI called pyproject-build.

As for why to use pipx run, it doesn't require you polute your global python site packages, or make a new venv just for building everywhere, or all the other ugly workarounds. `pipx install build` isn't awful, but it not great; it breaks every time the system python gets updates (at least for me), and requires that I remember the weird "pyproject-build` command instead of a simple authoritative "build".

It's also perfect on CI; see [here](https://scikit-hep.org/developer/gha_pure#distribution-pure-python-wheels) - you don't need to setup Python, you just run a single line, and pipx (which is on the GHA runners) downloads, installs, and runs build for you.

gaborbernat: Fair enough 🤷🏻

henryiii: Give pipx run a try, you might like it. :) I find this job (linked above) quite beautiful, and it will get even better when GitHub updates pipx:

```yaml
 dist:
    runs-on: ubuntu-latest
    steps:

    - uses: actions/checkout@v1

    - name: Build SDist and wheel
      run: pipx run --spec build pyproject-build

    - uses: actions/upload-artifact@v2
      with:
        path: dist/*

    - name: Check metadata
      run: pipx run twine check dist/*
```

layday: I personally find this problematic as with other uses of `pipx run` because they are not project-local.  I'd love to use `pipx run` if it implied using the Python version that is active in my local development environment and were able to differentiate between development environments if it were to continue to cache packages in my `$HOME`.

gaborbernat: I personally use tox to create a build environment and publish it to PyPi, and there tox does the lifting for me, as such don't need pipx run there.

henryiii: That's kind of the point; pipx run is for applications where the python version really doesn't matter (you wouldn't use this with `build` to build a wheel for an extension, for example  - but then again, you'd use cibuildwheel for that anyway). If you are in a "local development environment", you shouldn't be using pipx run. But tools like `build`, `twine`, `cibuildwheel`, etc. are not tied to your local development environment.

henryiii: You can tie it to any Python you want, though, with --python.

gaborbernat: My solution is https://github.com/tox-dev/tox/blob/rewrite/tox.ini#L108
and runs the same in CI and locally :-)

henryiii: Looking through my history, besides build, I use `pipx run twine` a lot, along with `pipx run nox`. I install my virtual environments as "venv" directories, which breaks the script in `pip` that picks up dependency bump updates - so `pipx run nox` fixed that for me by not having a /venv directory with nox installed in it being picked up as a dependency bump update.  (When I was bumping packaging 20.9 for 21.0.1)

henryiii: Tox is also a valid solution, depending on what you are doing :) Now you just need to run `pipx run tox`. :P

henryiii: (I have tox in my homebrew, so haven't actually tried that one with pipx run)

henryiii: > Well, pipx does take after npx - npx can be run both locally

Actually, isn't this done with `--pypackages`, which uses `__pypackages__` from PEP 582?

layday: True, although I don't see PEP 582 ever gaining acceptance.

henryiii:
This is now on all the GitHub Actions runners, so `pipx run build` works out of the box in GitHub actions!

PR #248, opened 2021-03-03
Revert "setup.cfg: support direct usage from pipx run in 0.16.1.0+"

gaborbernat:
We should not expose the tool under multiple executable names. 

Reverts pypa/build#247


henryiii:
The confusion here was that this is not a "console_scripts"; no console script is created with this name. It's just an entry in the entry points available from importlib.metadata, and one that pipx run looks for.

PR #252, opened 2021-03-08
build: un-nest `chdir` context manager

layday:
An OS error in `_working_directory` would have been incorrectly
transformed into a `BuildBackendException`.


FFY00 (approved):

gaborbernat (commented):
Would be nice to have a test for it 👍🏻

gaborbernat (approved):

Issue #259, opened 2021-03-18
build does not validate that the "source directory" argument is a valid source directory


pfmoore:
If I run build in an empty directory (or any directory that isn't a Python project), rather than reporting the error, the tool appears to run setuptools and creates an empty project called UNKNOWN. If there's no `pyproject.toml`, I guess it's assuming setuptools as a fallback. But if there's no `setup.py` or `setup.cfg` this is clearly incorrect, and so should fail.

As it stands, the build runs and results in unwanted (and useless) `build`, `dist` and `UNKNOWN.egg-info` directories being created in my project. Luckily, I wasn't already using `build` or `dist` as actual directories for my code...

FFY00:
I am tempting to say this is an issue with setuptools? Though it might not be that clear. IMO setuptools itself should fail if the `setuptools.build_meta:__legacy__` hook is called and there aren't any of those files, or others that it may support in the future. What do you think?

pfmoore:
Personally, I think that build should fail early if it detects it's being run from the wrong directory. In a purist sense, that means there's no `pyproject.toml`. But because you fall back to setuptools, IMO you need to check for `setup.py` or `setup.cfg` as well.

Yes, there's a setuptools issue here, as it shouldn't blindly package up something that's obviously wrong, but as a quality of implementation feature, build should check and fail early. After all, if it's left to setuptools, you would have to wait for the whole build environment to be created before you get told "nope, this isn't a Python project"...

gaborbernat:
I'd accept a PR that fails early 👍

layday:
> But if there's no `setup.py` or `setup.cfg` this is clearly incorrect, and so should fail.

PEP 517 requires the existence of `setup.py` for legacy projects - it does not say anything about `setup.cfg` - and I don't think we should apply the fallback to `setup.py`-less projects because these projects would not be possible to build without PEP 517.

FFY00:
My argument against that is that I feel like it is not very correct, and might not work all that well in the long run. Checking for `setup.py` is fine-ish, but checking for a `setup.cfg`, or any other configration file, is setuptools-defined behavior that I do not feel good depending on. From the top of my mind, what if setuptools deprecates it and starts ignoring it?
The issue never gets properly fixed in this project, I think it should be in setuptools, which would also fix this behavior in all other PEP 517 frontends.

PEP 517 also says, if there is no `pyproject.toml`, invoke setuptools.

[...]

PR #260, opened 2021-03-18
build: validate the source directory passed to ProjectBuilder

pfmoore:


pfmoore:
Fixes #259.

Initial version, for some reason the `tox -e fix` command doesn't work on my PC (it doesn't recognise my copy of Visual Studio when building the pre-commit environments) so I can't run the linting locally. I'll let CI do it for me...

pfmoore:
I don't know why the tests are failing, they work locally.

src/build/__init__.py
```diff
@@ -59,6 +59,16 @@ class TypoWarning(Warning):
     """
 
 
+def _validate_source_directory(srcdir):
+    # type: (str) -> None
+    if not os.path.isdir(srcdir):
+        raise BuildException('Source {} is not a directory'.format(srcdir))
+    pyproject_toml = os.path.join(srcdir, "pyproject.toml")
+    setup_py = os.path.join(srcdir, "setup.py")
```
FFY00, line 67: You are not running pre-commit. Please use single quotes here :stuck_out_tongue:

src/build/__init__.py
```diff
@@ -59,6 +59,16 @@ class TypoWarning(Warning):
...
+    if not (os.path.exists(pyproject_toml) or os.path.exists(setup_py)):
```
FFY00, line 68: nitpick: I find it easier to read in the expanded form using De Morgan's laws. I do not need to keep in mind that a logical block is open and mentally negate the statement.

FFY00 (approved):
I think checking for either `pyproject.toml` or `setup.py` is *fine*, even though I don't like that people with a `setup.cfg` will continue with the current behavior, making things slightly inconsistent, so I am not gonna block this.

src/build/__init__.py
pfmoore: Thanks. Is the right way to run pre-commit to use the `tox -e fix` command? As I said, that's failing for me :-( Is there a contributing guide that explains how to set up for development, run tests, etc? I wasn't able to find one.

I tried running it by hand, which seems to have worked. So I've pushed a fix.

src/build/__init__.py
pfmoore: Also fixed. Thanks!

src/build/__init__.py
FFY00: It should, I don't know exactly what went wrong in your case :confused:. Perhaps @gaborbernat might have more insight?

FFY00: We should have a contributor guide, but right now I'm ETIME.

src/build/__init__.py
FFY00: Thanks :blush:

src/build/__init__.py
pfmoore: No worries, I completely understand.

FFY00:
You seem to have found an issue with our integration tests :sweat_smile:. We test building a few popular projects in the CI as a sanity check and it looks like we are invoking `build` in the directory where the github tarballs are extracted, not the folder that is inside them and contains the project.

I will open a separate PR, you can then rebase on top of it.

src/build/__init__.py
gaborbernat: > Thanks. Is the right way to run pre-commit to use the `tox -e fix` command? As I said, that's failing for me

Can you provide more exact details for this?

FFY00:
Btw, as the integration tests take a long time and are only included as a sanity check, you need to pass `--only-integration` or `--run-integration` to pytest. You can do this in tox as `tox -e py39 -- --only-integration`.

My PR will introduce a test dependency so you will need to recreate the tox environment, you can do so with `tox -r -e py39`

Sorry for the trouble!

gaborbernat:
> I don't know why the tests are failing, they work locally.

This is why I'm not the biggest fan of not running integration tests locally by default 😆 People think if it passed locally they are done.

gaborbernat (changes requested):
The logic looks ok, but we need tests for these cases.

src/build/__init__.py
pfmoore: I'll try to later, but basically it said I didn't have Visual C version 14 installed (I have VS 2019) when building the virtualenvs. But when I ran pre-commit outside of tox, it worked, and now `tox -e fix` works.

pfmoore:
Yep, once the current tests pass I'll add tests for the new functionality.

FFY00:
The tests should now be fixed now, @pfmoore can you rebase your branch on main?

pfmoore:
Thanks for the quick fix, rebase done.

FFY00:
Thanks, when you come back to add the tests, could you also add an entry in the changelog?

pfmoore:
Done.

By the way, the symlink from `docs/changelog.rst` to `CHANGELOG.rst` doesn't get extracted properly on Windows, and messes up the git repo. Twice, I nearly uploaded a commit which changed that file to a plain text file containing the text `../CHANGELOG.rst` 🙁 

For portability, it might be better not to include symlinks in the repository...

src/build/__init__.py
gaborbernat: Weird. Was it trying to build some c-extension package (such as black?)

FFY00:
Uh, I thought symlinks were fixed on Windows, I guess that's not the case.

CHANGELOG.rst
```diff
@@ -3,6 +3,15 @@ Changelog
 +++++++++
 
 
+0.3.2 (XX-XX-2021)
+==================
+
+- Validate that the supplied source directory is valid (`PR #260`_)
+
+.. _PR 260: https://github.com/pypa/build/pull/260
```
FFY00, line 11: Missing a `#` here, `PR #260`.

Could you also add a "Fixes #259"?

FFY00 (approved):
Other than this, it should be good! Thanks :blush:

src/build/__init__.py
pfmoore: Yes. I can't recall which one, I'll let you know once I manage to reproduce the issue.

pfmoore:
I do have symlinks enabled, and they do work. I suspect that the problem is that git might not handle them. I'm using a very vanilla git setup.

pfmoore:
@gaborbernat I've raised #262 with details of the pre-commit failure.

@FFY00 I think I've done everything needed now. I've no idea why the coverage has gone down (or what I need to do to fix that). The coverage report seems to be flaggking lines that I haven't touched as uncovered 🙁 

Let me know what (if anything) more I need to do.

FFY00:
Thanks! The coverage is a bit fiddly, it is fine now.

pfmoore:
Thanks @FFY00 and @gaborbernat for the help getting this PR through 🙂 Much appreciated.

tests/test_projectbuilder.py
```diff
@@ -136,6 +136,18 @@ def test_check_dependency(monkeypatch, requirement_string, expected):
     assert next(build.check_dependency(requirement_string), None) == expected
 
 
+def test_bad_project(test_no_project_path):
+    # Passing a nonexistent project directory
+    with pytest.raises(build.BuildException):
```
gaborbernat, line 141: Can you split each test into its own test please? 👌

gaborbernat (changes requested):
Otherwise looks good.

tests/test_projectbuilder.py
FFY00: The code already went in. This is not a big deal though.

henryiii:
Sorry I missed this. It's also valid to have setup.cfg but no setup.py or pyproject.toml.

henryiii:
See: https://github.com/joerick/cibuildwheel/blob/96aa38b56c241773f612ee702d638eb76c80576b/cibuildwheel/__main__.py#L164-L169

layday:
It's not for legacy projects - how are you going to be able to invoke `sdist` or `bdist_wheel`?  If it works in pip it must be because it calls `setup()` on the fly.

pfmoore:
If there's no `pyproject.toml` then (in certain circumstances - it's complicated with all sorts of compatibility cruft) pip will use the `setuptools.build_meta:__legacy__` backend, as suggested [in PEP 517](https://www.python.org/dev/peps/pep-0517/#source-trees).

henryiii:
Not having a pyproject.toml is not something I do, but setup.cfg only projects are valid, and currently work in `build`:

```bahs
cookiecutter gh:scikit-hep/cookie
rm pyproject.toml
rm setup.py
pipx run build
```

This correctly produces a wheel and SDist, just like pip would. (Technically, I'm not sure how you would make an SDist without build, actually). Current master would now be broken. We suggest setup.cfg only projects on https://packaging.python.org/tutorials/packaging-projects/ (with a pyproject.toml too, of course, but technically it can be omitted and pip and current build work).

henryiii:
Quick followup, pip checks this:

```bash
python3 -m pip wheel .
ERROR: Directory '.' is not installable. Neither 'setup.py' nor 'pyproject.toml' found.
```

So being just as picky in build is perfectly fine. Though it is a removal of something that used to work.

henryiii:
As of https://github.com/pypa/pip/pull/9945, we've now swapped with Pip. Pip 21.1.2 _does_ support _just_ setup.cfg now, and now we don't. 🤣

```console
$ pip wheel .
ERROR: Directory '.' is not installable. Neither 'setup.py' nor 'pyproject.toml' found.
WARNING: You are using pip version 21.1.1; however, version 21.1.2 is available.
You should consider upgrading via the '/Users/henryschreiner/tmp/quickcheck/venv/bin/python3.9 -m pip install --upgrade pip' command.
$ pip install --upgrade pip
$ pip wheel .
Processing /Users/henryschreiner/tmp/quickcheck
  DEPRECATION: A future pip version will change local packages to be built in-place without first copying to a temporary directory. We recommend you use --use-feature=in-tree-build to test your packages with this new behavior before it becomes the default.
   pip 21.3 will remove support for this functionality. You can find discussion regarding this at https://github.com/pypa/pip/issues/7555.
Collecting numpy>=1.13.3
  File was already downloaded /Users/henryschreiner/tmp/quickcheck/numpy-1.20.3-cp39-cp39-macosx_10_9_x86_64.whl
Building wheels for collected packages: quickcheck
  Building wheel for quickcheck (setup.py) ... done
  Created wheel for quickcheck: filename=quickcheck-0.0.0-py3-none-any.whl size=3372 sha256=57acfb84109eedde21404ffdea6ef0e87d6ef7663498e3220bb0c4a72b066d35
  Stored in directory: /private/var/folders/_8/xtbws09n017fbzdx9dmgnyyr0000gn/T/pip-ephem-wheel-cache-62ksr1uu/wheels/9a/34/3d/e8d9298f424ae357bac078a56990fce39937c7ed9018594f47
Successfully built quickcheck
```

PR #267, opened 2021-03-26
fix: require minimum versions of depenencies

### LINKED ISSUE #263: fix: specify minimum requirements in setup.cfg
Currently, some requirements do not have minimum versions specified, such as `importlib_metadata`. This can result in build being broken if installed into an environment with a very old version of these dependencies. All dependencies should have a minimum version specified. See https://github.com/pypa/packaging-problems/issues/476.

importlib_metadata should have a minimum of 0.22 or 1.0; there should also be minimums for the other requirements.

henryiii:
This fixes issues where an old version of requirements are cached on a users system, causing build to break on install. Testing by manually loading old versions of packages in tox. Closes #263.

Mostly selected min versions based on changelogs or just trying old versions and slowly increasing until one passed, except for packing, where I thought 19+ seemed fine and started with that.

@gaborbernat, I'm better with nox than tox, there might be ways to clean this up? Also, is there any nice trick to get pip load the oldest possible versions instead of the latest, without manually listing them?


uranusjr:
> Also, is there any nice trick to get pip load the oldest possible versions instead of the latest

Not sure I understand this line, oldest as in like packaging 1.0?

layday:
> Not sure I understand this line, oldest as in like packaging 1.0?

The minimum version that meets the dependency specification, as in https://github.com/pypa/pip/issues/8085.

henryiii:
Yes, that's exactly what I meant, will watch that issue! If you specify a version range, it's possible a user will get the oldest possible version, so it's important for testing. If no minimum version is specified, yes, that's the oldest version on PyPI, because a user could get that if it's already installed or cached. If you don't support the lowest version you allow, you can get broken installs. If you have enough users of your package, it eventually comes up.

henryiii:
Oops, I don't think the extra tests are running due to the way the GitHub Actions are set up to manually run each environment.

layday:
Why is this only being tested on 3.5 and 3.6?

henryiii:
Because the oldest possible requirements may not work on the newer Pythons, as they generally predate them. Generally, I'd put the oldest version only here, but 2.7 and 3.5 are different enough to include both, and 3.6 might become the new minimum eventually, so I added that too. Might be helped by a new "minimum" solver in Pip, but maybe not, since Pip doesn't know what version was the first to support Python x.y. Even the maintainer doesn't know that till after the Python release, sometimes. ;)

henryiii:
On a newer Python, you should have a dependency sitting around cached older than the Python version itself, so I think just checking the oldest Python supported (+/- the current special cases) is fine for this check.

tox.ini
```diff
@@ -55,6 +56,33 @@ commands =
...
 
+[testenv:py27-minversion]
+description = check minimum versions required of all dependencies
+deps =
+  importlib-metadata==0.22
+  virtualenv==20.0.35
+  typing==3.5.2.2
+  pep517==0.9.1
+  packaging==19.0
+  toml==0.10.0
```
gaborbernat, line 67: Testing with lower version dependencies is usually done via pip constraint files. Any reason why we should not use that instead of this? I think that's more universally usable.

tox.ini
```diff
@@ -55,6 +56,33 @@ commands =
...
+extras =
+    test
+commands =
+    pytest -rsx tests {posargs:-n auto}
+
+[testenv:{py35,py36}-minversion]
```
gaborbernat, line 73: IMHO we should find out what's our lowest version dependency on all versions, not just these old ones? Not/

gaborbernat (commented):
This makes the CI even slower, but is the price we have to pay for correctness I guess.

tox.ini
henryiii: That would be a lot more involved, and I don't think you get nearly as much out of it (this current patch fixes a bug seen in the wild). You are unlikely to have Python 3.9 + a version of packaging that does not support 3.9 pre-installed or cached (for example). The only way that would be likely to happen if if you locked on an old version, then loaded on a new version, or if users add upper limits to library versions (Poetry's weak point, IMO but can't convince the authors otherwise) causing something to force an old version. Generally, you shouldn't have libraries older than your copy of Python sitting around.

_build_ works with the versions listed. Python may not, but build does. :) You technically only need one test, but 2.7 and 3.x are different enough that there could be a variation. I'd rather just drop the 3.6 test, and say this remains "oldest supported" (with 2.x and 3.x versions).

tox.ini
henryiii: Sure, can do that. As long as we don't have dependabot updating requirements.txt's or something. :)

tox.ini
gaborbernat: You're making assumptions here that OS packaging people don't patch out our pip/wheel to older dependencies, and then users show up at our doors that the tool doesn't work. Would be nice to know for every python version what minimum dependencies must be met IMHO.

henryiii: It's not 0-gain, just much less gain for significantly more testing / time. I should probably at least run 3.9 and see how many of the deps still work. :)

henryiii: Looks like 3.9 works, so I can just expand this to everything. Doesn't take long to run.

henryiii:
I don't think this is testing was hoping it would test. By adding the min values to an extra, I was hoping it would be easier to see if it was out of date, and also to get an immediate error from the minversion test if the locked version was not allowed by the range in the install requires. But the installed versions do not seem to be the locked versions.

layday:
It looks like this behaviour changed sometime between pip 20.2.3 and 21.0.1; the latter does respect the extra dependency versions.

henryiii:
Intentionally? A lot changed in 20.3 with the new solver.

layday:
I can't say if it was intentional but I assume tox in CI calls an older version of pip with the old resolver.

layday:
Erm, actually, this only happens when you prefix the Python version - `tox -e minversions` installs the `lockminversions` dependencies but `tox -e py39-minversions` doesn't.  Perhaps someone familiar with tox's internals could explain what's going on here.

FFY00:
I think this is a good idea, though right now I don't have time to check the details, so I'll leave that to the other maintainers.
I just have one note, please use `setup.cfg` as the prefix, as that is the location of the core change. In our currently adopted commit message model, the prefix indicates location, not action.

henryiii:
> please use setup.cfg as the prefix

I changed to that structure before unmarking as draft. (I had to dig back a bit to see if it was "setup", "setup-cfg", or "setup.cfg", but I found a comment with the latter).

setup.cfg
```diff
@@ -49,6 +49,13 @@ docs =
     sphinx~=3.0
     sphinx-argparse-cli>=1.5
     sphinx-autodoc-typehints>=1.10
+lockminversions =
```
gaborbernat, line 52: Can we make this a constraint file rather than an extra group, please?

gaborbernat (changes requested):

setup.cfg
henryiii: Did you see my reasoning for the group? If you make it a constraint file, you can't ensure that is doesn't go out of sync, since it is not installed in the same command as the package. While if it's here, I'm pretty sure it does ensure it is in sync with the install_requires.

For example, let's say we start requiring `packaging>=20.9`, and we forget to update this. If it's a constrain file, either it will be installed first, and then the package will update packaging to 20.9 silently, or it's installed afterwords, and then it will be rolled back (and hopefully fail the tests, if we are lucky). While if it's an extra, it should immediately error out as unsolvable, with a clear description of why (`packaging>=20.9 ` and `packaging==19.0 ` in conflict). That was my thinking, anyway.

tox.ini
```diff
@@ -55,6 +56,14 @@ commands =
     mypy --python-version 3.5 src/build
     mypy --python-version 2.7 src/build
 
+[testenv:{py27, py35, py36, py37, py38, py39}-minversion]
```
henryiii, line 59: @gaborbernat is this the correct way to do this? (Also, I could add pypy and then combine the envlist  lines above, I think)

setup.cfg
gaborbernat: > For example, let's say we start requiring `packaging>=20.9`, and we forget to update this.

If we set up things correctly (which we can) the installation will fail. @uranusjr you can invoke the constraint file with a package and ensure the above case does not happen, no?

henryiii: If we can make that happen, I'm happier with a constraint file, since this is not to be used externally (though, currently, there's no way to list a project's extras).

henryiii: Yes, I have something "working" (as in, if I manually break it by forcing too-old toml, it breaks :) ). Will commit momentarily.

tox.ini
```diff
@@ -55,6 +55,14 @@ commands =
     mypy --python-version 3.5 src/build
     mypy --python-version 2.7 src/build
 
+[testenv:{py39, py38, py37, py36, py35, py27, pypy3, pypy2}-minversion]
+description = check minimum versions required of all dependencies
+extras =
+    test
+commands =
+    pip install . -r tests/requirements-min.txt
```
henryiii, line 63:
```suggestion
install_command = python -m pip install {opts} -r tests/requirements-min.txt {packages}
commands =
```

tests/requirements-min.txt
```diff
@@ -0,0 +1,6 @@
+importlib-metadata==0.22;python_version < "3.8"
```
gaborbernat, line 1: This file is generally called ``constraints.txt``.

tox.ini
```diff
@@ -55,6 +55,14 @@ commands =
     mypy --python-version 3.5 src/build
     mypy --python-version 2.7 src/build
 
+[testenv:{py39, py38, py37, py36, py35, py27, pypy3, pypy2}-minversion]
+description = check minimum versions required of all dependencies
```
gaborbernat, line 59: You want skip_install=True here.

tox.ini
```diff
@@ -3,7 +3,7 @@ envlist =
     fix
     type
     docs
-    {py39, py38, py37, py36, py35, py27, pypy3, pypy2}{, -path, -sdist, -wheel}
+    {py39, py38, py37, py36, py35, py27, pypy3, pypy2}{, -path, -sdist, -wheel, -minversion}
```
gaborbernat, line 6: IMHO -min is enough.

tox.ini
gaborbernat, line 63: We could first build a wheel and then perform the install here of that? Rather than installing the folder itself.

gaborbernat (changes requested):

tox.ini
layday: That's not how you use a constraints file:

```suggestion
    pip install -c tests/requirements-min.txt .
```

henryiii: Constraints do not install a package listed if not listed anywhere else (in pip 20.3+), but that's all. The solver still includes the combined requirements. I've verified that a mismatch does fail. The -c is better, since this is exactly a constraint file and shouldn't install anything, but it still works as a requirements file.

tox.ini
henryiii: Is that better than adding `-c tests/constraints.txt` to the install command?

tox.ini
henryiii: We don't pre-build a wheel for any of the other tests?

tox.ini
gaborbernat: I'd say so.

tox.ini
gaborbernat: We do within the suite. But probably you don't want to use that here.

henryiii:
I've included everything except pre-building a wheel. Do we want to do that?

tox.ini
henryiii: Okay.

gaborbernat (approved):

tests/constraints.txt
```diff
@@ -0,0 +1,6 @@
+importlib-metadata==0.22;python_version < "3.8"
+packaging==19.0
+pep517==0.9.1
+toml==0.10.0
+typing==3.5.2.2;python_version < "3"
```
layday, line 5: Hmm, why is the minimum version here different from the minimum version in `setup.cfg`?

henryiii: If you are referring to typing==3.5.2.2, there wasn't a `typing==3.5.2` release (took several tries apparently to get it on PyPI), so this is the actual first version available for download. If you really think it should be written out in the minimums, it can be, but a four digit minimum seems excessive, and there is no 3.5.2 or 3.5.2.1 to cause an issue here. Can change if you prefer, though.

If it's another one, then that's likely a mistake.

CHANGELOG.rst
```diff
@@ -3,12 +3,15 @@ Changelog
 +++++++++
 
 
-0.3.2 (XX-XX-2021)
-==================
+Unreleased
+==========
 
 - Validate that the supplied source directory is valid (`PR #260`_, Fixes `#259`_)
+- Set and test minimum requirements for all dependencies (`PR #267`_, Fixes `#263`_)
```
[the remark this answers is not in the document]
henryiii: Will apply after the following comment is resolved.

tests/constraints.txt
layday: It looks like the uploaded sdist for 3.5.2 has a version number of 3.5.2.2, which is the version pip is seeing on the simple API.  Ah well, as long as it works and there's no intermediate versions.

henryiii: From the tag:

> ```
> Bump version to 3.5.2.2.
> (That's the first one I successfully uploaded to PyPI.)
> ```

layday:
LGTM, thanks for working on this.

PR #274, opened 2021-04-16
env: use symlinks except on Windows

### LINKED ISSUE #271: Cannot create venv _without_ symlinks using Command Line Tools build of Python on macOS
See https://bugs.python.org/issue38705.  This seems to have bitten some users lately:

* https://github.com/pypa/packaging-problems/issues/493
* https://github.com/pypa/packaging-problems/issues/488

Note that the venv CLI defaults to using symlinks on platforms other than Windows so we could do the same in build.  We could otherwise discourage build users from relying on the macOS-provided Python.

henryiii:
Matching the `python -m venv` default, see https://github.com/python/cpython/blob/2e789a1f1d84b343a996e8654590703b5fbdd441/Lib/venv/__init__.py#L403-L406.

Closes #271.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,11 +8,14 @@ Unreleased
 
 - Validate that the supplied source directory is valid (`PR #260`_, Fixes `#259`_)
 - Set and test minimum versions of build's runtime dependencies (`PR #267`_, Fixes `#263`_)
+- Use symlinks on creating venv's when available (`PR #274`_, Fixes `#271`_)
 
-.. _PR #267: https://github.com/pypa/build/pull/267
 .. _PR #260: https://github.com/pypa/build/pull/260
-.. _#263: https://github.com/pypa/build/issues/259
+.. _PR #267: https://github.com/pypa/build/pull/267
+.. _PR #274: https://github.com/pypa/build/pull/274
 .. _#259: https://github.com/pypa/build/issues/259
+.. _#263: https://github.com/pypa/build/issues/263
+.. _#271: https://github.com/pypa/build/issues/271
 
 Breaking Changes
 ----------------
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -2,6 +2,7 @@
 Creates and manages isolated build environments.
 """
 import abc
+import functools
 import os
 import platform
 import shutil
@@ -187,6 +188,23 @@ def _create_isolated_env_virtualenv(path):  # type: (str) -> Tuple[str, str]
 # venv only exists on Python 3+
 if sys.version_info >= (3,):  # noqa: C901
 
+    @functools.lru_cache(maxsize=None)
+    def _fs_supports_symlink():  # type: () -> bool
+        """Return True if symlinks are supported"""
+        # Using definition used by venv.main()
+        if os.name != 'nt':
+            return True
+
+        # Windows may support symlinks (setting in Windows 10)
+        with tempfile.NamedTemporaryFile(prefix='build-symlink-') as tmp_file:
+            dest = '{}-b'.format(tmp_file)
+            try:
+                os.symlink(tmp_file.name, dest)
+                os.unlink(dest)
+                return True
+            except (OSError, NotImplementedError, AttributeError):
+                return False
+
     def _create_isolated_env_venv(path):  # type: (str) -> Tuple[str, str]
         """
         On Python 3 we use the venv package from the standard library.
@@ -196,7 +214,7 @@ if sys.version_info >= (3,):  # noqa: C901
         """
         import venv
 
-        venv.EnvBuilder(with_pip=True).create(path)
+        venv.EnvBuilder(with_pip=True, symlinks=_fs_supports_symlink()).create(path)
         executable, script_dir, purelib = _find_executable_and_scripts(path)
 
         # Get the version of pip in the environment
--- a/tests/test_env.py
+++ b/tests/test_env.py
@@ -153,3 +153,22 @@ def test_pip_needs_upgrade_mac_os_11(mocker, pip_version, arch):
         else:
             (uninstall_call,) = check_call.call_args_list
             assert uninstall_call[0][0][1:] == ['-m', 'pip', 'uninstall', 'setuptools', '-y']
+
+
+@pytest.mark.isolated
+@pytest.mark.skipif(IS_PY2, reason='venv module used on Python 3 only')
+@pytest.mark.skipif(IS_PYPY3 and os.name == 'nt', reason='Isolated tests not supported on PyPy3 + Windows')
+@pytest.mark.parametrize('has_symlink', [True, False] if os.name == 'nt' else [True])
+def test_venv_symlink(mocker, has_symlink):
+    if has_symlink:
+        mocker.patch('os.symlink')
+        mocker.patch('os.unlink')
+    else:
+        mocker.patch('os.symlink', side_effect=OSError())
+
+    # Cache must be cleared to rerun
+    build.env._fs_supports_symlink.cache_clear()
+    supports_symlink = build.env._fs_supports_symlink()
+    build.env._fs_supports_symlink.cache_clear()
+
+    assert supports_symlink is has_symlink


FFY00 (approved):

henryiii:
Will update changelog once tests complete.

uranusjr:
Can this be improved to “use symlinks unless you can’t”? Symlinking is actually possible on Windows ([it’s a simple toggle away on Windows 10](https://www.ghacks.net/2016/12/04/windows-10-creators-update-symlinks-without-elevation/)). IIRC @gaborbernat had some logic in virtualenv that does this.

FFY00:
We had that at some point, perhaps we should bring it back.

https://github.com/pypa/build/commit/dad5f626742b5643ae0d01e386855cdb28f34aa4#diff-4f1bb46d65de7fbb1573dfc2a2bca9f63c7b03d55e74c53339f0c8ee4fa054d8R21

src/build/env.py
```diff
@@ -187,6 +188,25 @@ def _create_isolated_env_virtualenv(path):  # type: (str) -> Tuple[str, str]
 # venv only exists on Python 3+
 if sys.version_info >= (3,):  # noqa: C901
 
+    once = functools.lru_cache(maxsize=None)
```
gaborbernat, line 191: This shouldn't be public, you probably want to prefix it with `_`

gaborbernat (changes requested):
Some tests for these, please 👍🏻

henryiii:
I thought about tests, but wasn't sure what to test (it it does run as part of the other tests). Suggestions on what would be useful? Mock a system's name / symlink support, I guess?

src/build/env.py
```diff
@@ -187,6 +188,25 @@ def _create_isolated_env_virtualenv(path):  # type: (str) -> Tuple[str, str]
...
+
+    @once
```
FFY00, line 193: Why are we assigning it to a variable? Did I missing something?

FFY00:
> I thought about tests, but wasn't sure what to test (it it does run as part of the other tests). Suggestions on what would be useful? Mock a system's name / symlink support, I guess?

I think mocking `_fs_supports_symlink` should be enough.

CHANGELOG.rst
```diff
@@ -8,11 +8,14 @@ Unreleased
 
 - Validate that the supplied source directory is valid (`PR #260`_, Fixes `#259`_)
 - Set and test minimum versions of build's runtime dependencies (`PR #267`_, Fixes `#263`_)
+- Use symlinks when making venv's on Unix (`PR #274`_, Fixes `#271`_)
```
layday, line 11:
```suggestion
- Use symlinks when creating venv's where available (`PR #274`_, Fixes `#271`_)
```

src/build/env.py
```diff
@@ -187,6 +188,25 @@ def _create_isolated_env_virtualenv(path):  # type: (str) -> Tuple[str, str]
...
+    def _fs_supports_symlink():  # type: () -> bool
+        """Return True if symlinks are supported"""
+        # Using definition used by venv.main()
+        if os.name != 'nt':
+            return True
+
+        # Windows may support symlinks (setting in Windows 10)
+        with tempfile.NamedTemporaryFile(prefix='TmP') as tmp_file:
```
layday, line 201: Use a prefix which points to build.

src/build/env.py
```diff
@@ -187,6 +188,25 @@ def _create_isolated_env_virtualenv(path):  # type: (str) -> Tuple[str, str]
...
+            temp_dir = os.path.dirname(tmp_file.name)
+            dest = os.path.join(temp_dir, '{}-{}'.format(tmp_file.name, 'b'))
```
layday, line 203: Why do we need to do all this instead of just suffiixng "-b" to the filename?

src/build/env.py
henryiii: This was discussed as an addition to functools, where it would have been called `functools.once`. It didn't make it very far since `once = functools.lru_cache(maxsize=None)` is hard to improve on in terms of performance (it's internally implemented in C) and is only one line long (and would actually be `once = functools.cache` in Python 3.9+). Happy to use it directly instead though if you want to. This highlights what it does _slightly_ better, I think - it's only evaluated once. The `maxsize=None` could be confusing without the rename; it's not actually growing, it just is needed to remove the eviction code.

src/build/env.py
henryiii: I just inherited [this code](https://github.com/pypa/build/commit/dad5f626742b5643ae0d01e386855cdb28f34aa4#diff-4f1bb46d65de7fbb1573dfc2a2bca9f63c7b03d55e74c53339f0c8ee4fa054d8R21), happy to change.

src/build/env.py
henryiii: Umm.... Because that's what was done before and I inherited it? I see no reason why we can't just do that, yes.

src/build/env.py
henryiii: Before I change this, do you want:

1. Current `once = ` name
2. Use `_once = ` name
3. Use directly, no variable
4. Use directly, no variable, but have a comment saying this is there to cache the value once on first call.

I like options 1-3 best personally; I don't really like 4 since options 1 & 2 do the same thing but use a named variable instead of a comment.

src/build/env.py
layday: Hadn't noticed it was inherited :)

src/build/env.py
henryiii: I used Option 3 in the current PR.

gaborbernat (approved):
Let's just fix the pypy3 test, suite but otherwise looks ok

henryiii:
> fix the pypy3 test

Actually, the problem is that the PyPy3 test on Windows _passes_, and this code adds an xfail strict to it. https://github.com/pypa/build/blob/c4383f0b69984c1ea6e78900b9b603b6d1149895/tests/conftest.py#L89-L93 I'm guessing that means "isolated" doesn't work, so this needs to be skipped or the isolated marker needs to be removed. What exactly is `pytest.mark.isolated` used for here?

henryiii:
PS: Of the two linked PRs in the injected xfail, https://foss.heptapod.net/pypy/pypy/-/issues/3323 and https://foss.heptapod.net/pypy/pypy/-/issues/3321, the second one is now resolved. @mattip had a suggestion for the first, but guessing it hasn't been implemented yet since it's still open.

src/build/env.py
FFY00: This is fine, though it was not obvious to me. I would just apply the decorator directly, but no need to change the commit just for that.

CHANGELOG.rst
```diff
@@ -8,11 +8,14 @@ Unreleased
 
 - Validate that the supplied source directory is valid (`PR #260`_, Fixes `#259`_)
 - Set and test minimum versions of build's runtime dependencies (`PR #267`_, Fixes `#263`_)
+- Use symlinks when creating venv's on where available (`PR #274`_, Fixes `#271`_)
```
FFY00, line 11: Maybe change to `Use symlinks when creating venv's when available`?

src/build/env.py
henryiii: That's what I did in the current commit, that was option 3 :)

CHANGELOG.rst
henryiii: Sure, a tiny bit better and the current version has an extra word anyway. Any thoughts to the PyPy3 test and the xfail? (Possibly more for @gaborbernat, who seems to have looked into it 6 months ago)

FFY00: Not really, sorry. I haven't looked at that in a while. I'll defer to @gaborbernat as he was the one who introduced this in the first place and may have more context than me right now.

tests/test_env.py
```diff
@@ -153,3 +153,22 @@ def test_pip_needs_upgrade_mac_os_11(mocker, pip_version, arch):
         else:
             (uninstall_call,) = check_call.call_args_list
             assert uninstall_call[0][0][1:] == ['-m', 'pip', 'uninstall', 'setuptools', '-y']
+
+
+@pytest.mark.isolated
+@pytest.mark.skipif(IS_PY2, reason='venv module used on Python 3 only')
+@pytest.mark.skipif(IS_PYPY3 and os.name == 'nt', reason='Isolated tests not supported on PyPy3 + Windows')
```
henryiii, line 160: Added this until @gaborbernat can comment any with other ideas.

gaborbernat: So what's the issue here with pypy3 and windows? Didn't upstream fix this by now?

mattip: I am pretty sure symlink in windows is not supported by PyPy. There are various issues open, like https://foss.heptapod.net/pypy/pypy/-/issues/3434 around that topic.

PR #276, opened 2021-04-16
ci: don't run code tests in documentation changes

FFY00:


layday:
Good thinking, should probably add the changelog and readme to ignores as well.

.github/workflows/test.yml
```diff
@@ -0,0 +1,126 @@
...
+  check:
+    runs-on: ${{ matrix.os }}-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu
+        tox_env:
+          - type
+          - docs
```
gaborbernat, line 109: This env absolutely must run on documentation changes.

gaborbernat (changes requested):

.github/workflows/test.yml
```diff
@@ -0,0 +1,126 @@
...
+        run: |
+          import os; os.rename('.tox/coverage.{}.xml'.format(os.environ['BASE']), '.tox/coverage.xml')
+        shell: python
+
+      - uses: codecov/codecov-action@v1
+        if: ${{ always() }}
+        env:
+          PYTHON: ${{ matrix.python }}
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
```
webknjaz, line 94:
```suggestion
```

.github/workflows/test.yml
webknjaz: FWIW RTD integration runs on PRs so it's roughly equivalent. I'd suggest adding a strict mode to avoid broken references completely, though...

.github/workflows/test.yml
```diff
@@ -0,0 +1,126 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu
+        tox_env:
+          - type
+          - docs
+
+    steps:
```
webknjaz, line 111: It's cleaner to just set an env var instead of passing `-e` to each step... Plus here's a few extra args that make tox output nicer in GHA in some cases:
```suggestion
    env:
      PY_COLORS: 1
      TOXENV: ${{ matrix.tox_env }}
      TOX_PARALLEL_NO_SPINNER: 1

    steps:
```

FFY00: I applied these changes, with the exception of `TOXENV`, as we no longer have a matrix for tox jobs, they are run in different workflows. Will keep in mind though, thanks!

.github/workflows/test.yml
FFY00: Ups, I have put the docs run in the check workflow.

henryiii:
I think the required checks in the settings -> branches -> main branch protection rules need to be updated right before this is merged. The check / docs and check / type names has changed.

FFY00:
Yes, I am just waiting for approvals. I think I have addressed everything that was pointed out.

layday (approved):

PR #278, opened 2021-04-16
ci: don't pass a token to the codecov action

webknjaz:
It's no longer necessary, it supports tokenless uploads for quite a while now.


FFY00 (approved):
Thanks!

PR #282, opened 2021-04-17
ci: test against Python 3.10

FFY00:
On top of #276

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,14 +28,15 @@ jobs:
           - macos
           - windows
         py:
-          - 3.9
-          - 3.8
-          - 3.7
-          - 3.6
-          - 3.5
-          - 2.7
-          - pypy3
-          - pypy2
+          - '3.10-dev'
+          - '3.9'
+          - '3.8'
+          - '3.7'
+          - '3.6'
+          - '3.5'
+          - '2.7'
+          - 'pypy3'
+          - 'pypy2'
 
     steps:
       - name: Setup python for tox
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,14 +39,6 @@ jobs:
           - 'pypy2'
 
     steps:
-      - name: Setup python for tox
-        uses: actions/setup-python@v2
-        with:
-          python-version: 3.9
-
-      - name: Install tox
-        run: python -m pip install tox
-
       - uses: actions/checkout@v2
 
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,122 @@
...
+
+      - name: Install tox
+        run: python -m pip install tox
+
+      - uses: actions/checkout@v2
+
+      - name: Setup python for test ${{ matrix.py }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.py }}
```
layday, line 55:
```suggestion
          python-version: ${{ matrix.py }}
          if: ${{ matrix.py != '3.10' }}
        - name: Setup python for test 3.10
          uses: actions/setup-python@v2
          with:
            python-version: 3.10-dev
          if: ${{ matrix.py == '3.10' }}
```

FFY00: This is not actually needed because we have an interpreter identification step. Just changing it to `3.10-dev` in the matrix works.

layday: Yeah, I forgot we don't pass the Python version from the matrix to tox.

layday:
Isn't this always testing 3.9 now? You can't have the default Python be tox if you run `python` to extract the version we're testing.

FFY00:
Ah, yes :facepalm:. I forgot to also pull the tox environment name identification routine up. Should be fixed now.

FFY00:
Okay, it seems pytest is broken on Python 3.10.

FFY00:
Should be fixed by https://github.com/pytest-dev/pytest/commit/8be16280423eafbc3c604a81419f40a8c1681fba.

layday:
Revisit after beta 1?

henryiii:
> Okay, it seems pytest is broken on Python 3.10.

Yes, sorry, I could have told you that; that's caused an issue with pybind11's test suite too.

henryiii:
It's a PyTest update that will fix it, I believe, not Python.

henryiii:
This is a possible solution if you'd like to get this in before the next release:

https://github.com/pybind/pybind11/pull/2967 (applied to setup.cfg's test extra in this case)

FFY00:
No need, I'll wait for a new pytest release.

henryiii:
Forgot about this, pytest had a release some time ago. Retriggering the CI to see if it gets picked up.

FFY00:
It's working now :blush:

PR #284, opened 2021-04-21
build: rename get_dependencies to get_requires

### LINKED ISSUE #182: Rename `ProjectBuilder.get_dependencies`
Reading https://github.com/pypa/build/issues/181#issuecomment-721579088, I think this method's name is ambiguous and should be renamed to e.g. `get_backend_dependencies`.  Would there be any objections?

FFY00:
Fixes #182

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -33,6 +33,7 @@ Breaking Changes
 - The ``skip_dependencies`` argument of ``build.__main__.build_package`` has been renamed to ``skip_dependency_check``
 - ``build.ConfigSettings`` has been renamed to ``build.ConfigSettingsType``
 - ``build.ProjectBuilder.build_dependencies`` to ``build.ProjectBuilder.build_system_requires``
+- ``build.ProjectBuilder.get_dependencies`` to ``build.ProjectBuilder.get_requires_for_build``
 
 
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -32,6 +32,7 @@ Breaking Changes
 - The ``--skip-depencencies`` option has been renamed to ``--skip-dependency-check``
 - The ``skip_dependencies`` argument of ``build.__main__.build_package`` has been renamed to ``skip_dependency_check``
 - ``build.ConfigSettings`` has been renamed to ``build.ConfigSettingsType``
+- ``build.ProjectBuilder.build_dependencies`` to ``build.ProjectBuilder.build_system_requires``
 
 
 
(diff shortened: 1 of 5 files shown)


gaborbernat (approved):

layday:
I think this adds more confusion instead of disambiguating between pyproject.toml and backend dependencies.  The methid name is still not qualified and now there are two terms used to describe the same concept (require and dependency).

FFY00:
We can rename it to `build_system_dependencies`... Do you have any better proposal?

layday:
* `declared_dependencies` iso `build_dependencies`
* `get_dependencies_from_backend` iso `get_dependencies`

FFY00:
>     * `declared_dependencies` iso `build_dependencies`

Why `declared_dependencies`? Both are declared, I think `build_system_dependencies` is clearer. 

>     * `get_dependencies_from_backend` iso `get_dependencies`

I personally don't really see a need to call it `get_dependencies_from_backend` instead of `get_dependencies`, given that it is a callable and the other is an attribute, but if others feel that this is needed, I am okay with changing it.

> We could use "requires" for both of these if we want to align them with the PEPs.

Sure.

layday:
> Why `declared_dependencies`? Both are declared, I think `build_system_dependencies` is clearer.

`get_requires` is dynamic.  When you call `get_requires_for_build_wheel` on setuptools and it returns `["wheel"]`, where is `"wheel"` declared?

> I personally don't really see a need to call it `get_dependencies_from_backend` instead of `get_dependencies`, given that it is a callable and the other is an attribute, but if others feel that this is needed, I am okay with changing it.

That would provide perhaps the slightest hint that one's coming from the backend and the other from `pyproject.toml`, based purely on programmer intuition that a backend call is less likely to be a property.

layday:
If we're adopting the term 'requires' (obscure as it might be), we could then simply shorten the hook name to `get_requires_for_build("sdist_or_wheel")`. This should be fairly unambiguous to anybody familiar with PEP 517, reads nicely and would also mirror the `build` method, cf. `build("sdist_or_wheel")`.  `build_system_requires` uses the exact same terms found in `pyproject.toml` and would be OK with me, instead of `declared_requires`.

FFY00:
I would like to go for `get_requires_for_build` and `build_system_requires`.

layday:
Sounds good to me!

FFY00:
Done. On top of #296 and #297 to avoid conflicts, should go in after both of them. Only the last two commits are relevant.

gaborbernat (approved):

PR #296, opened 2021-05-19
Better error output

FFY00:


CHANGELOG.rst
```diff
@@ -11,12 +11,14 @@ Unreleased
 - Use symlinks on creating venv's when available (`PR #274`_, Fixes `#271`_)
 - Error sooner if pip upgrade is required and fails (`PR #288`_, Fixes `#256`_)
 - Add a ``runner`` argument to ``ProjectBuilder`` (`PR #290`_, Fixes `#289`_)
+- Better error output (`PR #296`_)
```
gaborbernat, line 14: This could be expanded a bit :thinking:

src/build/__init__.py
```diff
@@ -332,14 +339,25 @@ def _call_backend(self, callback, outdir, config_settings=None, **kwargs):
         else:
             os.mkdir(outdir)
 
+        with self._handle_backend(callable.__name__):
+            basename = callback(outdir, config_settings, **kwargs)  # type: str
+            return os.path.join(outdir, basename)
```
gaborbernat, line 344: I think the return line might be de-indented :thinking:

gaborbernat (approved):

src/build/__init__.py
FFY00: Eh, doesn't matter anyway, the context manager will run perfectly fine and close after the return.

FFY00: Nevermind, we want to prevent capturing exceptions from `os.path.join` 🤦

CHANGELOG.rst
FFY00: New message: "Hide irrelevant `pep517` error traceback and improve error messages"

src/build/__init__.py
FFY00: Fixed.

gaborbernat (approved):

PR #300, opened 2021-05-21
main: disable colors on windows

FFY00:
It seems Windows has settings to enable/disable color support by default
and we'd have to use low level Windows APIs to force enable, so I am
just gonna remove colors altogether. Shouldn't really be a big deal as
they are only used for the ERROR and WARNING prefixes.

### DIFF
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -17,6 +17,14 @@ from build import BuildBackendException, BuildException, ConfigSettingsType, Pro
 from build.env import IsolatedEnvBuilder
 
 
+try:
+    import colorama
+except ImportError:
+    pass
+else:
+    colorama.init()  # fix colors on windows
+
+
 __all__ = ['build', 'main', 'main_parser']
 
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,7 @@ Unreleased
 - Error sooner if pip upgrade is required and fails (`PR #288`_, Fixes `#256`_)
 - Add a ``runner`` argument to ``ProjectBuilder`` (`PR #290`_, Fixes `#289`_)
 - Hide irrelevant ``pep517`` error traceback and improve error messages (`PR #296`_)
+- Try to use ``colorama`` to fix colors on Windows (`PR #300`_)
 
 .. _PR #260: https://github.com/pypa/build/pull/260
 .. _PR #267: https://github.com/pypa/build/pull/267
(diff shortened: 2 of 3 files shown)


uranusjr:
How about a ready-made solution like colorama? Colours on Windows is a solved problem, build is having issue only because the current implementation is inventing wheels from scratch.

henryiii:
I need a tears emoji reaction. I love colors...

henryiii:
PS: I know that @FFY00 is against adding dependencies, _but_ this would be a Windows-only dependency, which might be a lot less invasive?

FFY00:
I added it as a hard dependency on windows, but not actually a runtime dependency, people wanting to use the project to bootstrap environments may skip the dependency.

I am not 100% sure of the decision, so if it causes issues for anyone, we will reconsider.

henryiii:
I like the idea quite a bit, I think.

Note that adding comments in setup.cfg is not supported when using setup-cfg-fmt (which I love using), and the changelog entry is still unchanged, I think. (From my phone it seems to be)

FFY00:
Oops, I forgot the change log.

I tried putting a comment in a line itself and it got removed, but it seems appending it to a meaningful line works fine.

src/build/__main__.py
```diff
@@ -17,6 +17,14 @@
 from build.env import IsolatedEnvBuilder
 
 
+try:
+    import colorama
+except ImportError:
+    pass
+else:
+    colorama.init()  # fix colors on windows
```
layday, line 25: This should probably be done in `main()`, not on module import, even if that module is `__main__`.

FFY00: It should actually be done on `build_package`, but that may run many times. I think this approach is *fine*, but if you really think it would be worth to change, I can do that.

layday: I don't feel too strongly about it, more a matter of principle, that code shouldn't run on import - let's keep it as is.

gaborbernat: I also vote to do it in the main method... I don't see why it would it be done in build_package?

FFY00: If I do the following, colors will be broken.

```python
import build.__main__

build.__main__.build_package(...)
```

We already set the `showwarning` on import, so we have precedent, if we choose to not run this on import, we should move that too.
We can also change the implementation so that `build_package` to not call `_error`, but rather raise an error, and then we would call `_error` in `main`.

I would go for the PR as-is, and then one of you can make a followup one fixing what you want.

FFY00:
I am gonna merge as-is, please open a PR moving `colorama.init()` to a different place if you care about that.

PR #303, opened 2021-06-02
build: add ProjectBuilder.metadata_path

### LINKED ISSUE #301: Provide helper to get the metadata as a `pathlib.Path`-ish object
I have found myself needing this more than once. I think we should provide a helper that calls `prepare('wheel')`, then `build('wheel')` if that fails, and provides a `pathlib.Path`-like object, ready to be plugged into `importlib.metadata.PathDistribution`.

This would allow us to easily do:
```python
build = build.ProjectBuilder('.')

with tempfile.TemporaryDirectory() as tmp:
    metadata = importlib.metadata.PathDistribution(
        builder.metadata_path(tmp)
    ).metadata
```

It might require pulling `zipp` as a dependency, for `zipfile` backports, but as I don't want to add more dependencies, this functionality would only be available on Python versions where the stdlib's `zipfile` module works for this.

FFY00:
Fixes #301

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -56,6 +56,9 @@ test =
     pytest-cov>=2
     pytest-mock>=2
     pytest-xdist>=1.34
+    setuptools>=42.0.0
+    wheel>=0.36.0
+    pathlib2>=2.0.0;python_version < "3"
 typing =
     mypy==0.800
     typing-extensions>=3.7.4.3
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -9,10 +9,12 @@ import contextlib
 import difflib
 import io
 import os
+import re
 import subprocess
 import sys
 import types
 import warnings
+import zipfile
 
 from collections import OrderedDict
 from typing import AbstractSet, Any, Callable, Dict, Iterator, Mapping, Optional, Sequence, Set, Text, Tuple, Type, Union
@@ -33,6 +35,13 @@ _ExcInfoType = Union[
     Tuple[Type[BaseException], BaseException, types.TracebackType],
     Tuple[None, None, None],
 ]
+
+
+_WHEEL_NAME_REGEX = re.compile(
+    r'(?P<distribution>.+)-(?P<version>.+)'
+    r'(-(?P<build_tag>.+))?-(?P<python_tag>.+)'
+    r'-(?P<abi_tag>.+)-(?P<platform_tag>.+)\.whl'
+)
 
 
 _DEFAULT_BACKEND = {
@@ -328,6 +337,38 @@ class ProjectBuilder(object):
         build = getattr(self._hook, 'build_{}'.format(distribution))
         kwargs = {} if metadata_directory is None else {'metadata_directory': metadata_directory}
         return self._call_backend(build, output_directory, config_settings, **kwargs)
+
+    def metadata_path(self, output_directory):  # type: (str) -> str
+        """
+        Generates the metadata directory of a distribution and returns its path.
+
+        If the backend does not support the ``prepare_metadata_for_build_wheel``
+        hook, a wheel will be built and the metadata extracted.
+
+        :param output_directory: Directory to put the metadata distribution in
+        """
+        # prepare_metadata hook
+        metadata = self.prepare('wheel', output_directory)
+        if metadata is not None:
+            return metadata
+
+        # fallback to build_wheel hook
+        wheel = self.build('wheel', output_directory)
+        match = _WHEEL_NAME_REGEX.match(os.path.basename(wheel))
+        if not match:
+            raise ValueError('Invalid wheel')
+        distinfo = '{}-{}.dist-info'.format(
+            # Python 2 does not support match['group']
+            match.group('distribution'),
+            match.group('version'),
+        )
+        member_prefix = '{}/'.format(distinfo)
+        with zipfile.ZipFile(wheel) as w:
+            w.extractall(
+                output_directory,
+                (member for member in w.namelist() if member.startswith(member_prefix)),
+            )
+        return os.path.join(output_directory, distinfo)
 
     def _call_backend(self, callback, outdir, config_settings=None, **kwargs):
         # type: (Callable[...,str], str, Optional[ConfigSettingsType], Any) -> str
--- a/tests/constraints.txt
+++ b/tests/constraints.txt
@@ -4,3 +4,5 @@ pep517==0.9.1
 toml==0.10.0
 typing==3.5.3.0;python_version < "3"
 virtualenv==20.0.35;python_version < "3"
+setuptools==42.0.0
+wheel==0.36.0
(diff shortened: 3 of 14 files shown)


FFY00:
I ended up bringing `zipp` as a dependency because otherwise this method would only be available on Python >=3.8, but made it soft-fail at runtime. As Python 2 is EOL and it would require a new dependency on `pathlib2`, I decided to make this only available on Python 3.

layday:
Can we extract the `dist-info` folder from the zip file instead of pulling zipp in as a dependency, if `zipfile.Path` is not available?

src/build/__init__.py
```diff
@@ -360,6 +382,32 @@ def _handle_backend(self, hook):  # type: (str) -> Iterator[None]
...
+
+
+            .. caution::
+
+               Only available on Python 3.
+            """
+            assert zipfile, 'metadata_path is not available due to missing zipp module'
+            try:
+                return pathlib.Path(typing.cast(str, self.prepare('wheel', output_directory)))
+            except TypeError:
```
layday, line 401: Catching a type error from the `Path`'s constructor is a roundabout way to check if the return value of `prepare` is not null - I would restructure this as:

```suggestion
            metadata = self.prepare('wheel', output_directory)
            if metadata is not None:
                return pathlib.Path(metadata)
            else:
```

No cast necessary.

FFY00:
Hum, we can, but that will leave leftover files in the output directory, though I think it's probably reasonable.

FFY00:
Nevermind, that happens already with the prepare hook :facepalm:

tests/constraints.txt
```diff
@@ -4,3 +4,6 @@ pep517==0.9.1
 toml==0.10.0
 typing==3.5.3.0;python_version < "3"
 virtualenv==20.0.35;python_version < "3"
+trampolim==0.0.2;python_version < "3.7"
```
layday, line 7: This looks like a dependency that might be difficult to satisfy, e.g. for distros.  Can we have an in-tree backend which implements a barebones `build_wheel` hook for testing purposes?  Something like:

```patch
diff --git a/tests/packages/test-no-prepare/prepareless_backend.py b/tests/packages/test-no-prepare/prepareless_backend.py
new file mode 100644
index 0000000..44e932d
--- /dev/null
+++ b/tests/packages/test-no-prepare/prepareless_backend.py
@@ -0,0 +1,27 @@
+import os.path
+import zipfile
+
+
+def build_wheel(wheel_directory, config_settings=None, metadata_directory=None):
+    with zipfile.ZipFile(
+        os.path.join(wheel_directory, 'test_no_prepare-1.0.0-py3-none-any.whl'),
+        'w',
+    ) as wheel:
+        wheel.writestr(
+            'test_no_prepare-1.0.0.dist-info/METADATA',
+            '''\
+Metadata-Version: 2.2
+Name: test-no-prepare
+Version: 1.0.0
+'''
+        )
+        wheel.writestr(
+            'test_no_prepare-1.0.0.dist-info/WHEEL',
+            '''\
+Wheel-Version: 1.0
+Generator: build
+Root-Is-Purelib: true
+Tag: py3-none-any
+'''
+        )
+    return wheel.filename
diff --git a/tests/packages/test-no-prepare/pyproject.toml b/tests/packages/test-no-prepare/pyproject.toml
index 23bccd8..8795521 100644
--- a/tests/packages/test-no-prepare/pyproject.toml
+++ b/tests/packages/test-no-prepare/pyproject.toml
@@ -1,7 +1,4 @@
 [build-system]
-build-backend = 'trampolim'
-requires = ['trampolim']
-
-[project]
-name = 'test-no-prepare'
-version = '1.0.0'
+build-backend = 'prepareless_backend'
+requires = []
+backend-path = ['.']

```

FFY00: I changed this to setuptools and used an intree backend that just does `from setuptools.build_meta import build_sdist, build_wheel`.

FFY00:
Since we are not returning `zipfile.Path` anymore, I changed `metadata_path` to return the path string instead of a `pathlib.Path`-like object, making this also available on Python 2.

setup.cfg
```diff
@@ -56,6 +56,9 @@ test =
     pytest-cov>=2
     pytest-mock>=2
     pytest-xdist>=1.34
+    setuptools>=40.8.0
+    wheel>=0.36.0
+    pathlib2>=2.0.0;python_version >= "3"
```
layday, line 61: < "3"

src/build/__init__.py
```diff
@@ -329,6 +338,36 @@ def build(self, distribution, output_directory, config_settings=None, metadata_d
...
+        Generates the metadata directory of a distribution and returns its path.
+
+        If the backend does not support the ``prepare_metadata_for_build_wheel``
+        hook, a wheel will be built and the metadata extracted.
+
+        :param output_directory: Directory to put the metadata distribution in
+        """
+        # prepare_metadata hook
+        metadata = self.prepare('wheel', output_directory)
+        if metadata:
```
layday, line 352: Nitpicking but if `metadata is not None`.  In the unlikely case that the backend returns an empty string we'd be masking their error.

tests/packages/test-no-prepare/pyproject.toml
```diff
@@ -0,0 +1,4 @@
+[build-system]
+build-backend = 'backend_no_prepare'
+backend-path = ['.']
+requires = ['setuptools >= 42.0.0']
```
layday, line 4: Should specify `wheel`.

tests/packages/test-setuptools/pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[build-system]
+requires = ['setuptools >= 40.8.0', 'wheel']
```
layday, line 2: This should be `>= 42` too, `wheel >= 0.36.0`.

henryiii: Agreed. <42 does not work properly with pyproject.toml. I had a package that was broken only on my system but not CI. Took a long time to finally discover it was using a cached setuptools that was buggy under pyproject.toml, it was either 40.x or 41.x.

src/build/__init__.py
```diff
@@ -35,6 +37,13 @@
 ]
 
 
+_WHEEL_NAME_REGEX = re.compile(
+    r'(?P<distribution>.+)-(?P<version>.+)'
+    r'(-(?P<build_tag>.+))?-(?P<python_tag>.+)'
+    r'-(?P<abi_tag>.+)-(?P<platform_tag>.+).whl'
```
layday, line 43: Escape the dot: `\.whl`.

Can we use `parse_wheel_filename` from `packaging.utils`?  Also missing coverage on line 358 'cause of this.

FFY00: Because it is fairly simple and allows us to limit the runtime dependency.

layday (approved):
Looks good, thanks for bearing with me :)

FFY00:
Thanks, the feedback was very helpful.

Issue #198, opened 2020-12-07
Listing filenames of produced distributions in the CLI


layday:
Source distribution and wheel filenames are variable; they encode a variety of information (e.g. platform, Python version, distribution version) which might vary between build invocations.  We might want to think about offering a way to retrieve these from the CLI (e.g. through a new option which would create a manifest) for scripts and automation tools to refer to which would provide a minor convenience over globbing.

FFY00:
IMO this is out of scope. Maybe if this information wasn't in the file name I would agree on having a manifest with extra information.

I think it is fairly simple to write a Python module that just parses the wheel name and output json or something like that.

```python
import argparse
import json
import os.path
import re

from typing import Any, Dict, Optional


_WHEEL_NAME_REGEX = re.compile(
    r'(?P<distribution>.+)-(?P<version>.+)'
    r'(-(?P<build_tag>.+))?-(?P<python_tag>.+)'
    r'-(?P<abi_tag>.+)-(?P<platform_tag>.+).whl'
)


def parse(name: str) -> Optional[Dict[str, str]]:
    if m := re.match(_WHEEL_NAME_REGEX, name):
        return m.groupdict()
    return None


if __name__ == '__main__':
    parser = argparse.ArgumentParser()
    parser.add_argument(
        'file',
        type=str,
        help='wheel file',
    )
    args = parser.parse_args()

    if info := parse(os.path.basename(args.file)):
        print(json.dumps(info, indent=4, sort_keys=True))
    else:
        print('Inavlid wheel name, see https://www.python.org/dev/peps/pep-0427/#file-name-convention')
```

```
python -m wheel2json ~/Downloads/packaging-20.7-py2.py3-none-any.whl
{
    "abi_tag": "none",
    "build_tag": null,
    "distribution": "packaging",
    "platform_tag": "any",
    "python_tag": "py2.py3",
    "version": "20.7"
}
```

FFY00:
Perhaps something for https://github.com/pypa/wheel? `python -m wheel info ~/Downloads/packaging-20.7-py2.py3-none-any.whl`?

layday:
I think you're misunderstanding - I don't want to parse the wheel filename into its constituent tags.  I want `build` simply to return the filename of the distribution file the backend has produced in a machine-readable format.  It could then be piped into a hypothetical installer or another tool which might expect an sdist or a wheel.  For instance, if we were to produce a JSON file:

```sh
$ python -m build --write-manifest-to manifest.json
[...]

PR #304, opened 2021-06-07
main: build binary distributions via sdists by default

### LINKED ISSUE #257: Build wheel from sdist (optionally?)
The proposal is to build the wheel from an sdist rather than directly from the source.  Perhaps optionally, perhaps only when both `--sdist` and `--wheel` are specified, perhaps always.  Presently I have this implemented in towncrier as below.

```bash
# build the sdist
python -m build --sdist --outdir ${toxinidir}/dist/ ${toxinidir}
tar -xvf ${toxinidir}/dist/*
cd *
# build the wheel from the sdist
python -m build --wheel --outdir ${toxinidir}/dist/ .
```

It was suggested that perhaps this would be a useful feature for build to provide itself.  See more context and discussion below.

https://github.com/twisted/towncrier/pull/323#discussion_r594948931

Following Trio's lead, I implemented package builds for towncrier such that first an sdist is built and then a wheel is built from that sdist.  Following that, those two files are distributed throughout the CI tests and checks and (upcoming) automated publishing.  The latter bit makes sure that the specific wheel to be published is what is being tested.  The former bit makes sure that the sdist is able to generate a functional wheel.  I don't mean this to test that build and wheel and so on work correctly, but rather to test that the package being built is setup properly such that it will result in a functional sdist.  [Twisted #10103](https://twistedmatrix.com/trac/ticket/10103) was shared as an example issue that this workflow would have caught.

FFY00:
Fixes #257

### DIFF
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -140,10 +140,9 @@ def build_package(srcdir, outdir, distributions, config_settings=None, isolation
     :param isolation: Isolate the build in a separate environment
     :param skip_dependency_check: Do not perform the dependency check
     """
-    with _handle_build_error():
-        builder = ProjectBuilder(srcdir)
-        for distribution in distributions:
-            _build(isolation, builder, outdir, distribution, config_settings, skip_dependency_check)
+    builder = ProjectBuilder(srcdir)
+    for distribution in distributions:
+        _build(isolation, builder, outdir, distribution, config_settings, skip_dependency_check)
 
 
 def build_package_via_sdist(srcdir, outdir, distributions, config_settings=None, isolation=True, skip_dependency_check=False):
@@ -161,21 +160,20 @@ def build_package_via_sdist(srcdir, outdir, distributions, config_settings=None,
     if 'sdist' in distributions:
         raise ValueError('Only binary distributions are allowed but sdist was specified')
 
-    with _handle_build_error():
-        builder = ProjectBuilder(srcdir)
-        sdist = _build(isolation, builder, outdir, 'sdist', config_settings, skip_dependency_check)
+    builder = ProjectBuilder(srcdir)
+    sdist = _build(isolation, builder, outdir, 'sdist', config_settings, skip_dependency_check)
 
-        # extract sdist
-        sdist_name = os.path.basename(sdist)
-        sdist_out = tempfile.mkdtemp(dir=outdir, prefix='build-via-sdist-')
-        with tarfile.open(sdist) as t:
-            t.extractall(sdist_out)
-            builder = ProjectBuilder(os.path.join(sdist_out, sdist_name[: -len('.tar.gz')]))
-            for distribution in distributions:
-                _build(isolation, builder, outdir, distribution, config_settings, skip_dependency_check)
+    # extract sdist
+    sdist_name = os.path.basename(sdist)
+    sdist_out = tempfile.mkdtemp(dir=outdir, prefix='build-via-sdist-')
+    with tarfile.open(sdist) as t:
+        t.extractall(sdist_out)
+        builder = ProjectBuilder(os.path.join(sdist_out, sdist_name[: -len('.tar.gz')]))
+        for distribution in distributions:
+            _build(isolation, builder, outdir, distribution, config_settings, skip_dependency_check)
 
-        # remove sdist source if there was no exception
-        shutil.rmtree(sdist_out, ignore_errors=True)
+    # remove sdist source if there was no exception
+    shutil.rmtree(sdist_out, ignore_errors=True)
 
 
 def main_parser():  # type: () -> argparse.ArgumentParser
@@ -259,7 +257,7 @@ def main_parser():  # type: () -> argparse.ArgumentParser
     return parser
 
 
-def main(cli_args, prog=None):  # type: (List[str], Optional[str]) -> None
+def main(cli_args, prog=None):  # type: (List[str], Optional[str]) -> None  # noqa: C901
     """
     Parse the CLI arguments and invoke the build process.
 
@@ -298,7 +296,12 @@ def main(cli_args, prog=None):  # type: (List[str], Optional[str]) -> None
     else:
         build_call = build_package_via_sdist
         distributions = ['wheel']
-    build_call(args.srcdir, outdir, distributions, config_settings, not args.no_isolation, args.skip_dependency_check)
+    try:
+        with _handle_build_error():
+            build_call(args.srcdir, outdir, distributions, config_settings, not args.no_isolation, args.skip_dependency_check)
+    except Exception as e:  # pragma: no cover
+        print(traceback.format_exc())
+        _error(str(e))
 
 
 def entrypoint():  # type: () -> None
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -3,13 +3,18 @@
 from __future__ import print_function
 
 import argparse
+import contextlib
 import os
+import shutil
 import subprocess
 import sys
+import tarfile
+import tempfile
+import textwrap
 import traceback
 import warnings
 
-from typing import Iterable, List, Optional, Sequence, TextIO, Type, Union
+from typing import Iterable, Iterator, List, Optional, Sequence, TextIO, Type, Union
 
 import build
 
@@ -26,6 +31,16 @@ else:
 
 
 __all__ = ['build', 'main', 'main_parser']
+
+
+if sys.version_info[0] == 2:
+
+    def _indent(text, prefix):  # type: (str, str) -> str
+        return ''.join(prefix + line for line in text.splitlines(True))
+
+
+else:
+    from textwrap import indent as _indent
 
 
 def _showwarning(message, category, filename, lineno, file=None, line=None):  # pragma: no cover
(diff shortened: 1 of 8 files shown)


src/build/__main__.py
```diff
@@ -57,54 +61,43 @@ def _format_dep_chain(dep_chain):  # type: (Sequence[str]) -> str
...
-        builder.build(dist, outdir, config_settings)
 
+def _build_in_current_env(builder, outdir, distribution, config_settings, skip_dependency_check=False):
+    # type: (ProjectBuilder, str, str, Optional[ConfigSettingsType], bool) -> str
+    if not skip_dependency_check:
+        missing = builder.check_dependencies(distribution)
+        if missing:
+            _error(
+                'Missing dependencies:'
+                + ''.join('\n\t' + dep for deps in missing for dep in (deps[0], _format_dep_chain(deps[1:])) if dep)
```
gaborbernat, line 83: Assign this string into a variable so that it fits in one line. Will make it easier to read.

gaborbernat (commented):
I'm +0.5 on this, but this would like a way to force the old behavior from the cli (build a wheel from the source directly). We've basically doubled the build time for a wheel, which sometimes can be a big regression.

src/build/__main__.py
FFY00: This was the old code, which I did not touch, but we can take the opportunity to improve :blush:

FFY00:
This patch only affects invoking `python -m build` without `--sdist`/`-s` or `--wheel`/`-w`. If you don't specify any of them, we will build sdist and then a wheel from it, if you specify any, you will get the previous behavior. `python -m build -w` will keep building the wheel directly.

`python -m build` - build a sdist and then a wheel from the sdist (modified behavior)
`python -m build -s` - build a sdist (old behavior)
`python -m build -w` - build a wheel directly (old behavior)
`python -m build -s -w` - build a sdist and wheel directlt (old behavior)

Essentially, if not distribution is specified, we will build a wheel from sdist, if any are, we will just build those distributions directly.

I forgot to touch up the help messages, I should fix that.

henryiii:
I'm +0.5 on this as well. Initially it seemed like it could be problematic, but every argument I could come up with against it already has an answer or obvious workaround (`python -m build --sdist --wheel` if you really need the old behavior, packages that can't build sdists won't build sdists anyway, etc). It could help catch bugs (SDist unable to build wheel), though it might not always "catch" them and could distribute wheels with missing files, but you would already be distributing broken SDists in that case. The biggest issue remaining is making it clear in the documentation that this is how it works - as long as it is clear that "default" is to do source -> sdist -> wheel instead of being a shortcut for "--sdist --wheel", I think it's fine.

src/build/__main__.py
```diff
@@ -126,6 +119,56 @@ def build_package(srcdir, outdir, distributions, config_settings=None, isolation
...
+
+    :param srcdir: Source directory
+    :param outdir: Output directory
+    :param distribution: Distribution to build (only wheel)
+    :param config_settings: Configuration settings to be passed to the backend
+    :param isolation: Isolate the build in a separate environment
+    :param skip_dependency_check: Do not perform the dependency check
+    """
+    if 'sdist' in distributions:
+        raise ValueError('Only binary distributions are allowed but sdist was specified')
```
layday, line 153: Looks like this is only done for simplicity in `main`. Suggest renaming the function to `build_sdist_and_wheel_from_sdist` and removing the `distributions` parameter. In `main` you can use a `partial`.

src/build/__main__.py
```diff
@@ -126,6 +119,56 @@ def build_package(srcdir, outdir, distributions, config_settings=None, isolation
...
+
+    with _handle_build_error():
+        builder = ProjectBuilder(srcdir)
+        sdist = _build(isolation, builder, outdir, 'sdist', config_settings, skip_dependency_check)
+
+        # extract sdist
+        sdist_name = os.path.basename(sdist)
+        sdist_out = tempfile.mkdtemp(dir=outdir, prefix='build-via-sdist-')
```
layday, line 161: Build artefacts should not be placed in `outdir`, suggesting removing the `dir` argument.

src/build/__main__.py
```diff
@@ -126,6 +119,56 @@ def build_package(srcdir, outdir, distributions, config_settings=None, isolation
...
+        with tarfile.open(sdist) as t:
+            t.extractall(sdist_out)
+            builder = ProjectBuilder(os.path.join(sdist_out, sdist_name.rstrip('.tar.gz')))  # noqa: B005
```
layday, line 164: `str.rstrip` will mangle filenames with local version identifiers, slice the string instead: `sdist_name[: -len('.tar.gz')]`.

src/build/__main__.py
```diff
@@ -126,6 +119,56 @@ def build_package(srcdir, outdir, distributions, config_settings=None, isolation
...
+            for distribution in distributions:
+                _build(isolation, builder, outdir, distribution, config_settings, skip_dependency_check)
+
+        # remove sdist source if there was no exception
```
layday, line 168: Why not remove it always?

src/build/__main__.py
henryiii: This is why B005 exists, and [exactly why](https://www.python.org/dev/peps/pep-0616/#rationale) they added `.remove` suffix in Python 3.9.

layday:
I sympathise with not wanting to add more options to the CLI, but I don't like that `∅` has different semantics from `-sw` all of a sudden.  If the default is to build an sdist and a wheel, why would anyone expect that `-sw` does something different? I don't think the documentation can compensate for the shortcoming in usability.  Can we think of something else we can do here, even if it's simply to disallow `-sw`?

webknjaz:
@FFY00 what about `python -m build --wheel dist/some-tarball.tar.gz`? I don't see this case solved, am I missing anything? This is an important use case specifically for building platform-specific matrixes of wheels.

FFY00:
That is a different change, this PR is only for the default behavior. I know you mentioned this in #257, but can you open a separate issue for that? So that we can track it as an actionable task.

src/build/__main__.py
FFY00: I used this a bit locally and every single time the build failed, I wanted to go check the source to debug the issue. Your mileage may vary, but I'd say it is very useful to have it there after a fail.

Perhaps in a next PR we can improve the error message to let the user know where they can find the source that failed building, and perhaps we could query the user and ask if they want to remove or not.

src/build/__main__.py
FFY00: Right :facepalm:! It seems my memory failed me as I rushed through  this PR.

src/build/__main__.py
FFY00: Why not? See my comment below, I don't want to put it in a temporary path as that is subject to be removed, and leaving stuff behind in such paths is not very hygienic.

src/build/__main__.py
FFY00: No, it was a deliberate design decision to keep the API future-proof.

layday: Okay, I didn't really think of anything in `__main__` as public API.

FFY00:
The last commit will also allow us to move the colorama stuff and `warnings.showwarning` overriding to `main` in a next PR.

src/build/__main__.py
layday: Conceptually, for much the same reasons that `bdist_wheel` makes a copy of the source tree in `build` and not `dist` - the output directory's there for the final build artefacts, not the bits and bobs that go into making them.  Besides being outside of the norm, it might also trip other programs up that don't do any filtering, especially considering we don't (yet) have an answer to #198. For example, this is what happens when you `twine upload dist/*`, where `dist` contains a folder:

```
$ twine upload dist/*
Uploading distributions to https://upload.pypi.org/legacy/
InvalidDistribution: Unknown distribution format: 'foo'
```

FFY00: But build artifacts are only left there if the build fails. I agree that we should not be keeping ephemeral build artifacts there on successful builds, but IMO it would be reasonable to put them there on failed builds.
A successful run should leave you with only the final artifacts, an unsuccessful one will leave you with the information needed to debug the issue.

FFY00: Also, I'd like to point out this behavior is common in many other tools.

layday: Fair enough, let's do it that way.

FFY00:
> If the default is to build an sdist and a wheel, why would anyone expect that `-sw` does something different? I don't think the documentation can compensate for the shortcoming in usability.

The people that would be affected by that are advanced users, which should not really have much of an issue reading the documentation. I am optimizing for the normal user, which would be the one that would fall into that documentation issue. Almost all normal users should actually want to build binary distributions via the source distribution.
I acknowledge this is not an optimal solution, but so far I don't thing such exists. We already discussed this in #257 and so far no better solution that matches these requirements was presented.

gaborbernat (approved):

src/build/__main__.py
```diff
@@ -126,14 +128,84 @@ def build_package(srcdir, outdir, distributions, config_settings=None, isolation
...
+                (sdists and wheels) as defined by PEP 517.
+
+                By default, it will build a source distribution (sdist) for the
+                project and then build the binary distributions (wheels).
+                This is recommended as it will ensure your source distribution (sdist)
+                can be used to build wheels.
+
+                If any distribution flag is passed, -s/--sdist or -w/--wheel, only
+                those distributions will be built, and they will be built directly
+                from the specified source.
```
layday, line 202: I think this a little hazy and repetitive, let's cut it down. Maybe:

```suggestion
                By default, a source distribution (sdist) is built from `srcdir` and
                a binary distribution (wheel) is built from the sdist.
                This is recommended as it will ensure the sdist can be used
                to build wheels.
                Pass `-s/--sdist` or `-w/--wheel` to build a specific
                distribution. If you pass `-w/--wheel`, the wheel
                will be built from `srcdir` and not from an sdist, even if you
                combine it with `-s/--sdist`.
```

henryiii (approved):

webknjaz:
> That is a different change, this PR is only for the default behavior. I know you mentioned this in #257, but can you open a separate issue for that? So that we can track it as an actionable task.

Right. Makes sense. Done: https://github.com/pypa/build/issues/311.

src/build/__main__.py
FFY00: I have rewritten as follows.

```
By default, a source distribution (sdist) is built from {srcdir}
and a binary distribution (wheel) is built from the sdist.
This is recommended as it will ensure the sdist can be used
to build wheels.

Pass -s/--sdist and/or -w/--wheel to build a specific distribution.
If you do this, the default behavior will be disabled, and all
artifacts will be built from {srcdir} (even if you combine
-w/--wheel with -s/--sdist, the wheel will be built from {srcdir}).
```

I used the `{srcdir}` notation to match what we already had.

layday (approved):

src/build/__main__.py
layday: Sounds good, thanks.

FFY00:
That is exactly the kind of issue this change is supposed to be uncovering! It means your package cannot be built from the sdist. If there is no compatible wheel on PyPI, pip will try to build the package from the sdist.

You have two options:
1) Ship the code in the sdist, so that it can be used to build the package
2) Acknowledge that the package cannot be built from a sdist, which means the only supported platforms are the ones you are shipping wheels for. In this case, you should not be uploading a sdist to PyPI, as it is designed to be broken.

gaborbernat:
I think that exposes a valid bug though. Surely a sdist created like that would also fail, because that relative path is not exist at the users machine either 🤔

henryiii:
I'm curious, if you have compiled code, are you only building one wheel (one OS/arch/Python version), or are you building identical sdists every time? Normally, I'd expect one SDist job and multiple wheel jobs (such as in https://scikit-hep.org/developer/gha_wheels). Build #311 (and cibuildwheel pypa/cibuildwheel#597) do not support running directly from the SDist, though I think it would be a good idea to be able to do that, to encourage this structure even in that case. :)

src/build/__main__.py
henryiii: This will break a lot of assumptions - it's the cause for #320, and likely will interfere with SCM versioning, etc. These need to be placed in an isolated output directory, not in source. Most other tools have a "--dirty" command that keeps from removing the temporary build directory if needed for debugging.

If you want a failed build to produce something detectable, maybe copy the contents of the temp dir to `<outdir>` on failure? Though leaving stuff around by default doesn't seem like a good idea - most users won't know that they need to clean up and the output dir could grow.

FFY00: Right, I failed to properly consider those use-cases. Let's do the build in an isolated directory.

fuglede:
> Acknowledge that the package cannot be built from a sdist, which means the only supported platforms are the ones you are shipping wheels for. In this case, you should not be uploading a sdist to PyPI, as it is designed to be broken.

FWIW, this was exactly my use case; an sdist-less internal package designed for use in controlled environments for which the source code could not meaningfully be bundled, and for which only wheels were necessary.

gaborbernat:
So you want to only build wheel then via `python -m build -w`, that should still build the wheel via the source tree.

henryiii:
There's absolutely nothing wrong with building only wheels; there is something deeply and inherently wrong about building SDists that are lies, being known to be invalid and broken. ;) So, yes, just ask for wheel-only and everything will be fine. The default SDist + wheel is the only thing that has changed (to SDist -> wheel). `python -m build --wheel` is what you actually wanted.

henryiii:
Can I change my vote after-the-fact to a strong +1? This is finding bugs even in build backends! :)

PR #324, opened 2021-06-26
Drop Python 2 and 3.5 support

layday:


.pre-commit-config.yaml
```diff
@@ -25,7 +25,7 @@ repos:
   rev: v1.16.0
   hooks:
   - id: setup-cfg-fmt
-    args: ["--min-py3-version", "3.5", "--max-py-version", "3.9"]
+    args: ["--min-py3-version", "3.6"]
```
henryiii, line 28: This is not actually needed - it's only there because setup-cfg-fmt can't detect what minimum Python 3 to start at if it detects 2.7+. If it detects 3.6+, there's no ambiguity, so you don't have to tell it manually where you start. One less place to duplicate the supported Python version number.

henryiii:
I'm pushing a few changes heavily based on the results from `pipx run pyupgrade --py36-plus **.py` (fish syntax, too lazy to rewrite the `**` in bash). I'd be happy to add it as a pre-commit check after this PR.

Does `tests/test_projectbuilder.py` need to be able to run in Python 2? It has `try: import pathlib` and other Python 2 syntax in it. Here's the patch:

```patch
diff --git a/tests/test_projectbuilder.py b/tests/test_projectbuilder.py
index dbbb41f..34f8a67 100644
--- a/tests/test_projectbuilder.py
+++ b/tests/test_projectbuilder.py
@@ -1,13 +1,10 @@
 # SPDX-License-Identifier: MIT
-# coding=utf-8

-from __future__ import unicode_literals

 import copy
 import os
 import sys
 import textwrap
-import typing

 import pep517.wrappers
 import pytest
@@ -185,7 +182,7 @@ def test_init(mocker, test_flit_path, legacy_path, test_no_permission, test_bad_
         build.ProjectBuilder(test_bad_syntax_path)


-@pytest.mark.parametrize('value', [b'something', typing.Text('something_else')])
+@pytest.mark.parametrize('value', [b'something', str('something_else')])
 def test_python_executable(test_flit_path, value):
     builder = build.ProjectBuilder(test_flit_path)
```

henryiii:
And this would add pyupgrade to the pre-commit list:

```patch
diff --git a/.pre-commit-config.yaml b/.pre-commit-config.yaml
index 244b1f0..ad014f7 100644
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,6 +13,11 @@ repos:
   - id: end-of-file-fixer
   - id: trailing-whitespace
   - id: double-quote-string-fixer
+- repo: https://github.com/asottile/pyupgrade
+  rev: v2.19.1
+  hooks:
+  - id: pyupgrade
+    args: ["--py36-plus"]
 - repo: https://github.com/psf/black
   rev: 20.8b1
   hooks:
```

(potentially adding that one test file that I've not updated to the ignore list if it really does need to be runnable from Python 2)

layday:
No, `test_projectbuilder.py` does not need to run from Python 2, I just omitted to drop the imports.

.pre-commit-config.yaml
layday: Sure, I'll take it out.

pyproject.toml
```diff
@@ -5,7 +5,7 @@ build-backend = 'setuptools.build_meta'
 [tool.black]
 line-length = 127
 skip-string-normalization = true
-target-version = ['py38', 'py37', 'py36', 'py35', 'py27']
+target-version = ['py38', 'py37', 'py36']
```
gaborbernat, line 8: No 3.9? Can we remove these?

gaborbernat (approved):

pyproject.toml
layday: I suppose there are some reformatting rules which aren't safe to apply to older versions, like parenthesised context managers. Listing all supported versions rather than just the minimum version seems a bit pointless however.

layday: We're also going to have to bump black in pre-commit to add `py39` to the list, 20.8b1 does not have it.

FFY00:
I will have a look at this later this week. I want to rework the history but don't want to bother you with that so please don't merge, I will rework the history and then merge myself.

layday:
Is there anything else we wanna do here?

FFY00:
Not really, I just need to fix up the history. I will try to look at it next week, sorry.

layday:
That's alright, was only checking if anybody wanted anything changed.

FFY00 (approved):
Thank you very much @layday and @henryiii! And sorry for the delay getting this reviewed and touched up.

layday:
Thanks for tidying up the commit messages ;)

PR #333, opened 2021-07-27
Curate output

### LINKED ISSUE #142: Curate output

FFY00:
Fixes #142

### DIFF
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -27,6 +27,7 @@ __all__ = ['build', 'main', 'main_parser']
 _STYLES = {
     'red': '\33[91m',
     'yellow': '\33[93m',
+    'bold': '\33[1m',
     'reset': '\33[0m',
 }
 
@@ -69,6 +70,12 @@ def _error(msg: str, code: int = 1) -> None:  # pragma: no cover
     """
     _print(f'[red]ERROR[reset] {msg}')
     exit(code)
+
+
+class _ProjectBuilder(ProjectBuilder):
+    @staticmethod
+    def log(message: str) -> None:
+        _print(f'[bold]* {message}[reset]')
 
 
 def _format_dep_chain(dep_chain: Sequence[str]) -> str:
@@ -158,7 +165,7 @@ def build_package(
     :param isolation: Isolate the build in a separate environment
     :param skip_dependency_check: Do not perform the dependency check
     """
-    builder = ProjectBuilder(srcdir)
+    builder = _ProjectBuilder(srcdir)
     for distribution in distributions:
         _build(isolation, builder, outdir, distribution, config_settings, skip_dependency_check)
 
@@ -184,7 +191,7 @@ def build_package_via_sdist(
     if 'sdist' in distributions:
         raise ValueError('Only binary distributions are allowed but sdist was specified')
 
-    builder = ProjectBuilder(srcdir)
+    builder = _ProjectBuilder(srcdir)
     sdist = _build(isolation, builder, outdir, 'sdist', config_settings, skip_dependency_check)
 
     sdist_name = os.path.basename(sdist)
@@ -193,7 +200,7 @@ def build_package_via_sdist(
     with tarfile.open(sdist) as t:
         t.extractall(sdist_out)
         try:
-            builder = ProjectBuilder(os.path.join(sdist_out, sdist_name[: -len('.tar.gz')]))
+            builder = _ProjectBuilder(os.path.join(sdist_out, sdist_name[: -len('.tar.gz')]))
             for distribution in distributions:
                 _build(isolation, builder, outdir, distribution, config_settings, skip_dependency_check)
         finally:
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -153,6 +153,18 @@ def _handle_build_error() -> Iterator[None]:
         _error(str(e))
 
 
+def _natural_language_list(elements: Sequence[str]) -> str:
+    if len(elements) == 0:
+        raise IndexError('no elements')
+    elif len(elements) == 1:
+        return elements[0]
+    else:
+        return '{} and {}'.format(
+            ', '.join(elements[:-1]),
+            elements[-1],
+        )
+
+
 def build_package(
     srcdir: str,
     outdir: str,
@@ -207,6 +219,8 @@ def build_package_via_sdist(
         t.extractall(sdist_out)
         try:
             builder = _ProjectBuilder(os.path.join(sdist_out, sdist_name[: -len('.tar.gz')]))
+            if distributions:
+                builder.log(f'Building {_natural_language_list(distributions)} from sdist')
             for distribution in distributions:
                 _build(isolation, builder, outdir, distribution, config_settings, skip_dependency_check)
         finally:
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -24,6 +24,20 @@ from build.env import IsolatedEnvBuilder
 __all__ = ['build', 'main', 'main_parser']
 
 
+_STYLES = {
+    'red': '\33[91m',
+    'yellow': '\33[93m',
+    'reset': '\33[0m',
+}
+
+
+def _print(message: str) -> None:
+    color = sys.stdout.isatty()
+    for name, code in _STYLES.items():
+        message = message.replace(f'[{name}]', code if color else '')
+    print(message)
+
+
 def _showwarning(
     message: Union[Warning, str],
     category: Type[Warning],
@@ -32,10 +46,7 @@ def _showwarning(
     file: Optional[TextIO] = None,
     line: Optional[str] = None,
 ) -> None:  # pragma: no cover
-    prefix = 'WARNING'
-    if sys.stdout.isatty():
-        prefix = '\33[93m' + prefix + '\33[0m'
-    print(f'{prefix} {message}')
+    _print(f'[yellow]WARNING[reset] {message}')
 
 
 def _setup_cli() -> None:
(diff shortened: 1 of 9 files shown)


FFY00:
Preview: https://asciinema.org/a/tjnnnrs94SF3ZdOmYvnOsXzxp

src/build/env.py
```diff
@@ -68,6 +72,19 @@ def _should_use_virtualenv() -> bool:
...
+    """Invoke subprocess and output stdout and stderr if it fails."""
+    try:
+        subprocess.check_call(
+            *args,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+            **kwargs,
+        )  # type: ignore  # https://github.com/python/mypy/issues/6799
+    except Exception as e:  # pragma: no cover
+        raise e
```
gaborbernat, line 85: This catch seems to do nothing :thinking:

src/build/env.py
```diff
@@ -68,6 +72,19 @@ def _should_use_virtualenv() -> bool:
     )
 
 
+def _subprocess(*args: Any, **kwargs: Any) -> None:
```
gaborbernat, line 75: args needs to be a sequence of str, not? Also for kwargs can we just state other parameters we use?

gaborbernat (changes requested):

src/build/env.py
```diff
@@ -148,6 +192,8 @@ def install(self, requirements: Iterable[str]) -> None:
         if not requirements:
             return
 
+        self._log('Installing packages on isolated environment... ({})'.format(', '.join(sorted(requirements))))
```
layday, line 195: on -> in

src/build/env.py
layday: Don't you just want `subprocess.check_output` here?

FFY00: Yes, sorry. It accidentally got removed.

src/build/env.py
```diff
@@ -68,6 +72,15 @@ def _should_use_virtualenv() -> bool:
     )
 
 
+def _subprocess(cmd: List[str]) -> None:
+    """Invoke subprocess and output stdout and stderr if it fails."""
+    try:
+        subprocess.check_output(cmd, stderr=subprocess.STDOUT)
+    except Exception as e:  # pragma: no cover
+        print(e.output.decode(), end='')  # type: ignore  # e.output isn't type hinted but is documented
```
gaborbernat, line 80: perhaps open an issue to typeshed and link it here instead?

src/build/env.py
```diff
@@ -68,6 +72,15 @@ def _should_use_virtualenv() -> bool:
     )
 
 
+def _subprocess(cmd: List[str]) -> None:
+    """Invoke subprocess and output stdout and stderr if it fails."""
+    try:
+        subprocess.check_output(cmd, stderr=subprocess.STDOUT)
+    except Exception as e:  # pragma: no cover
```
gaborbernat, line 79: this should be just subprocess.ProcessCaledError, not?

gaborbernat: Can we also create a test for this? :thinking:

gaborbernat (changes requested):

src/build/env.py
layday: You're catching a generic `Exception`, that's why it's not type hinted. Catch `CalledProcessError` and let everything else through.

FFY00: Yes, of course, sorry. I just woke up :sweat_smile:

src/build/__main__.py
```diff
@@ -24,6 +24,24 @@
...
+    'green': '\33[92m',
+    'yellow': '\33[93m',
+    'bold': '\33[1m',
+    'dim': '\33[2m',
+    'underline': '\33[4m',
+    'reset': '\33[0m',
+}
+
+
+def _print(message: str) -> None:  # pragma: no cover
```
gaborbernat, line 38: Why does this require pragma no cover?

src/build/env.py
```diff
@@ -68,6 +72,15 @@ def _should_use_virtualenv() -> bool:
     )
 
 
+def _subprocess(cmd: List[str]) -> None:
+    """Invoke subprocess and output stdout and stderr if it fails."""
+    try:
+        subprocess.check_output(cmd, stderr=subprocess.STDOUT)
+    except subprocess.CalledProcessError as e:  # pragma: no cover
```
gaborbernat, line 79: Why does this require pragma no cover? Can we add a test to not need it?

FFY00:
Sorry about my previous fix, I had just woken up and was sleepy :sweat_smile:. I fixed this properly now and also added dim to the tracebacks.

src/build/env.py
FFY00: Like pointed out above, I am catching `Exception`, which will obviously not have `e.output` :sweat_smile:

gaborbernat (approved):

src/build/__main__.py
```diff
@@ -24,6 +24,24 @@
...
+    'dim': '\33[2m',
+    'underline': '\33[4m',
+    'reset': '\33[0m',
+}
+
+
+def _print(message: str) -> None:
+    color = sys.stdout.isatty() and 'NO_COLOR' not in os.environ
+    for name, code in _STYLES.items():
+        message = message.replace(f'[{name}]', code if color else '')
```
layday, line 41: Why not just use string formatting instead of ad hoc replaces?  Something like:

```py
def _print(message: str) -> None:  # pragma: no cover
    color = sys.stdout.isatty() and 'NO_COLOR' not in os.environ
    if color:
        message = message.format(**_STYLES)
    else:
        message = message.format(**{k: '' for k in _STYLES})
    print(message)

[...]

_print(f'{{yellow}}WARNING{{reset}} {message}')
```

FFY00: The performance difference is not relevant, and I felt this had slightly cleaner code and considerably better readability in the messages -- the color formatting is clearly distinct from f-string variables and only takes one delimiter character on each side.

```python
import os
import sys


_STYLES = {
    'a': '1',
    'b': '2',
    'c': '3',
    'd': '4',
    'e': '5',
}


def _print_replace(message: str) -> None:
    color = sys.stdout.isatty() and 'NO_COLOR' not in os.environ
    for name, code in _STYLES.items():
        message = message.replace(f'[{name}]', code if color else '')
    #print(message)


def _print_format(message: str) -> None:  # pragma: no cover
    color = sys.stdout.isatty() and 'NO_COLOR' not in os.environ
    if color:
        message = message.format(**_STYLES)
    else:
        message = message.format(**{k: '' for k in _STYLES})
    #print(message)
```

```
$ python -m timeit -s 'import test_print' -- 'test_print._print_replace("[a][b][c][d][e]")'
100000 loops, best of 5: 2.35 usec per loop
$ python -m timeit -s 'import test_print' -- 'test_print._print_format("{{a}}{{b}}{{c}}{{d}}{{e}}")'
200000 loops, best of 5: 1.93 usec per loop
```

layday: I don't think the performance difference is important either - I just felt it's the obvious way to do string substitution. Let's keep the squares then if you think readability would suffer.

FFY00:
Hum, I though the automerge would wait for @henryiii as I requested a review from him. Let me know if there are any issues.

layday:
It didn't wait for me to approve either but it's all good 😛

FFY00:
Because you commented it removed the review request anyway :smile:

src/build/__main__.py
henryiii: This looks a lot like rich's mini language - why not exactly imitate that? `[yellow]<stuff>[/yellow]`. You could loop over over styles and add `/` versions that reset. Just a thought, instead of inventing something new just for this. _Very_ minor though.

henryiii: Also what happens if the string `[yellow]` happens to be part of the traceback, since tracebacks are passed through this? Doing fstring replacement wouldn't have this problem.

FFY00: More complexity in the implementation :stuck_out_tongue:. This is private anyway, so I just went with the simple thing that would fit our needs.

src/build/__main__.py
```diff
@@ -121,25 +145,39 @@ def _handle_build_error() -> Iterator[None]:
...
                     e.exc_info[0],
                     e.exc_info[1],
                     e.exc_info[2],
                     limit=-1,
                 )
+                tb = ''.join(tb_lines)
             else:
-                print(traceback.format_exc(-1))
+                tb = traceback.format_exc(-1)
+            _print('\n[dim]{}[reset]\n'.format(tb.strip('\n')))
```
henryiii, line 157: Here's where it looks like `[<stuff>]` could get introduces in the tb and miss-processed.

src/build/__main__.py
FFY00: > Also what happens if the string `[yellow]` happens to be part of the traceback, since tracebacks are passed through this? Doing fstring replacement wouldn't have this problem.

Right, we should not pass them through.

henryiii: Personally, in plumbum.colorlib, I use `print(yellow | "str") and `yellow` wraps the string it's or'd with with color tags.

FFY00: Right, but that introduces a bit more complexity than what we need. I opened #335.

PR #335, opened 2021-07-28
main: fix replacing color codes in tracebacks

FFY00:

### DIFF
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -35,11 +35,15 @@ _STYLES = {
 }
 
 
-def _print(message: str) -> None:
+def _format(message: str) -> str:
     color = sys.stdout.isatty() and 'NO_COLOR' not in os.environ
     for name, code in _STYLES.items():
         message = message.replace(f'[{name}]', code if color else '')
-    print(message)
+    return message
+
+
+def _print(message: str) -> None:
+    print(_format(message))
 
 
 def _showwarning(
@@ -154,7 +158,7 @@ def _handle_build_error() -> Iterator[None]:
                 tb = ''.join(tb_lines)
             else:
                 tb = traceback.format_exc(-1)
-            _print('\n[dim]{}[reset]\n'.format(tb.strip('\n')))
+            print(_format('\n[dim]{}[reset]\n').format(tb.strip('\n')))
         _error(str(e))
 
 
@@ -367,7 +371,7 @@ def main(cli_args: Sequence[str], prog: Optional[str] = None) -> None:  # noqa: 
             artifact_list = _natural_language_list([f'[underline]{artifact}[reset][bold][green]' for artifact in built])
             _print(f'[bold][green]Successfully built {artifact_list}')
     except Exception as e:  # pragma: no cover
-        _print('\n[dim]{}[reset]\n'.format(traceback.format_exc().strip('\n')))
+        print(_format('\n[dim]{}[reset]\n').format(traceback.format_exc().strip('\n')))
         _error(str(e))
 
 
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -33,17 +33,8 @@ _STYLES = {
     'underline': '\33[4m',
     'reset': '\33[0m',
 }
-
-
-def _format(message: str) -> str:
-    color = sys.stdout.isatty() and 'NO_COLOR' not in os.environ
-    for name, code in _STYLES.items():
-        message = message.replace(f'[{name}]', code if color else '')
-    return message
-
-
-def _print(message: str) -> None:
-    print(_format(message))
+if not sys.stdout.isatty() or 'NO_COLOR' in os.environ:
+    _STYLES = {color: '' for color in _STYLES}
 
 
 def _showwarning(
@@ -54,7 +45,7 @@ def _showwarning(
     file: Optional[TextIO] = None,
     line: Optional[str] = None,
 ) -> None:  # pragma: no cover
-    _print(f'[yellow]WARNING[reset] {message}')
+    print('{yellow}WARNING{reset} {}'.format(message, **_STYLES))
 
 
 def _setup_cli() -> None:
@@ -75,20 +66,20 @@ def _error(msg: str, code: int = 1) -> None:  # pragma: no cover
     :param msg: Error message
     :param code: Error code
     """
-    _print(f'[red]ERROR[reset] {msg}')
+    print('{red}ERROR{reset} {}'.format(msg, **_STYLES))
     exit(code)
 
 
 class _ProjectBuilder(ProjectBuilder):
     @staticmethod
     def log(message: str) -> None:
-        _print(f'[bold]* {message}[reset]')
+        print('{bold}* {}{reset}'.format(message, **_STYLES))
 
 
 class _IsolatedEnvBuilder(IsolatedEnvBuilder):
     @staticmethod
     def log(message: str) -> None:
-        _print(f'[bold]* {message}[reset]')
+        print('{bold}* {}{reset}'.format(message, **_STYLES))
 
 
 def _format_dep_chain(dep_chain: Sequence[str]) -> str:
@@ -158,7 +149,7 @@ def _handle_build_error() -> Iterator[None]:
                 tb = ''.join(tb_lines)
             else:
                 tb = traceback.format_exc(-1)
-            print(_format('\n[dim]{}[reset]\n').format(tb.strip('\n')))
+            print('\n{dim}{}{reset}\n'.format(tb.strip('\n'), **_STYLES))
         _error(str(e))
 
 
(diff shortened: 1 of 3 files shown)


gaborbernat (commented):
Can we add a test to validate the change?

FFY00:
I think adding a test for this that would actually be helpful in development would be very tricky, and I feel like it might not be worth the time. The code that we would be covering would only generate formatting errors, not functional ones which would prevent people from working.
How do you test this kind of stuff in tox?

gaborbernat:
Why would it be tricky? If now the contract is that the output is nicely colored, why would a test validating that be not worth it? tox threats output similarly. We run the code in a subprocess, and then assert the stdout/stderr contains the right series of bytes. If you can run it without needing a new subprocess you can assert on capsys/capfd and do the same. Doesn't have to test color output for all functional cases, just one run that covers all cases. Bugs related to the coloring helper function can be tested by directly calling the helper function from your test.

src/build/__main__.py
```diff
@@ -365,9 +369,9 @@ def main(cli_args: Sequence[str], prog: Optional[str] = None) -> None:  # noqa:
                 args.srcdir, outdir, distributions, config_settings, not args.no_isolation, args.skip_dependency_check
             )
             artifact_list = _natural_language_list([f'[underline]{artifact}[reset][bold][green]' for artifact in built])
-            _print(f'[bold][green]Successfully built {artifact_list}')
+            _print(f'[bold][green]Successfully built {artifact_list}[reset]')
     except Exception as e:  # pragma: no cover
-        _print('\n[dim]{}[reset]\n'.format(traceback.format_exc().strip('\n')))
+        print(_format('\n[dim]{}[reset]\n').format(traceback.format_exc().strip('\n')))
```
henryiii, line 374: > a bit more complexity than what we need

How is this not complex? In six months, will you remember the difference between `print`, `_print`, `format`, and `_format`?

I'd do either:

```python
        print('\n{dim}{msg}{reset}\n').format(msg=traceback.format_exc().strip('\n'), **_STYLES)
```

And remove `_print` and `_format` entirely. If NO_COLOR is specified, the dict would just be empty strings.

Or at the very least have `_format` pass-through to format, so you could do this:

```python
        print(_format('\n[dim]{}[reset]\n', traceback.format_exc().strip('\n'))
```

FFY00: I changed to

```python
print('{red}ERROR{reset} {}'.format(msg, **_STYLES))
print('\n{dim}{}{reset}\n'.format(tb.strip('\n'), **_STYLES))
...
```

src/build/__main__.py
```diff
@@ -364,10 +359,16 @@ def main(cli_args: Sequence[str], prog: Optional[str] = None) -> None:  # noqa:
...
+                ['{underline}{}{reset}{bold}{green}'.format(artifact, **_STYLES) for artifact in built]
+            )
+            print('{bold}{green}Successfully built {}{reset}'.format(artifact_list, **_STYLES))
     except Exception as e:  # pragma: no cover
-        _print('\n[dim]{}[reset]\n'.format(traceback.format_exc().strip('\n')))
+        colored_tb = '\n{dim}{}{reset}\n'.format(
+            traceback.format_exc().strip('\n'),
+            **_STYLES,
+        )
+        print(colored_tb)
```
henryiii, line 371: Personally, I'd make black happy this way:

```suggestion
        tb_str = traceback.format_exc().strip('\n')
        print('\n{dim}{}{reset}\n'.format(tb_str, **_STYLES))
```

henryiii (approved):

src/build/__main__.py
```diff
@@ -33,13 +33,8 @@
...
     'reset': '\33[0m',
 }
-
-
-def _print(message: str) -> None:
-    color = sys.stdout.isatty() and 'NO_COLOR' not in os.environ
-    for name, code in _STYLES.items():
-        message = message.replace(f'[{name}]', code if color else '')
-    print(message)
+if not sys.stdout.isatty() or 'NO_COLOR' in os.environ:
```
gaborbernat, line 36: I like that we have an environment variable to disable color, however can we have also one that force enables it? (e.g. tox 4 cannot forward the tty state into a subprocess yet, so is handy in such cases to force color mode). Perhaps FORCE_COLOR? And if neither set then check the tty state :blush:

gaborbernat (changes requested):
In general looks good, let's just add an environmetn variable to enable color mode :blush:

src/build/__main__.py
FFY00: Sure. What happens with both `FORCE_COLOR` and `NO_COLOR`?

gaborbernat: In my implementation NO_COLOR wins, but :shrug:

gaborbernat: See e.g. how this would be used https://github.com/tox-dev/tox/blob/rewrite/tox.ini#L79 (this shows how mypy handles it).

FFY00: Done.

gaborbernat: Shouldn't empty values translate as false for these flags?

FFY00: https://no-color.org/ says the mere presence enables the behavior, the value is irrelevant. I think we should have the same behavior in `FORCE_COLOR`.

henryiii: I'm a fan of having a FORCE_COLOR too, as often on GHA (which supports color) things can be missing color. I think no-color.org went overboard - it's a pretty well established practice for an empty environment variable to be the same as unsetting it. It's not always easy to unset an environment variable, and the disconnect there is irritating. But I'd follow what the spec dictates. :(

src/build/__main__.py
```diff
@@ -33,13 +33,8 @@
...
     'reset': '\33[0m',
 }
-
-
-def _print(message: str) -> None:
-    color = sys.stdout.isatty() and 'NO_COLOR' not in os.environ
-    for name, code in _STYLES.items():
-        message = message.replace(f'[{name}]', code if color else '')
-    print(message)
+if 'FORCE_COLOR' not in os.environ and not sys.stdout.isatty() or 'NO_COLOR' in os.environ:
```
gaborbernat, line 36: This is for me is hard to read and reason about :thinking: Can we do instead something like:

```python
def _init_colors():
	  use_colors = sys.stdout.isatty()
	  if os.environ.get('NO_COLOR', '').strip():
	     use_colors = False
	  elif os.environ.get('FORCE_COLOR', '').strip():
	     use_colors = True
	  
	  styles = {..}
	  if use_colors = False:
	     styles = {color: '' for color in _STYLES}
      return styles

_STYLES = _init_colors
```

This will make testing it easier to. No need to do the importlib reload tricks.

layday:
It's not entirely clear what this is fixing. Are you trying to avoid replacing "[dim]" and "[reset]" in the traceback itself?

FFY00:
Yes, as @henryiii pointed out.

src/build/__main__.py
FFY00: I went with

```python
def _init_colors() -> Dict[str, str]:
    if 'NO_COLOR' in os.environ:
        if 'FORCE_COLOR' in os.environ:
            warnings.warn(
                'Both NO_COLOR and FORCE_COLOR environment '
                'variables are set, disabling color'
            )
        return _NO_COLORS
    elif 'FORCE_COLOR' in os.environ or sys.stdout.isatty():
        return _COLORS
    return _NO_COLORS


_STYLES = _init_colors()
```

I don't really consider the module reload a trick, in fact, it helps us make sure the styles are set correctly at load :stuck_out_tongue:

gaborbernat: ok 👍 I still don't like that we consider empty value as set (as @henryiii pointed out that's in our opinion a bad practice, but 🤷 )

gaborbernat (approved):

src/build/__main__.py
FFY00: I understand the concern and share a similar opinion. IMO, coloring should be if `NO_COLOR` is present and it not `no`/`false`. But given that it is already a pretty well established standard, I think the best option is to follow what is defined.

FFY00:
The tests are failing on Windows and I don't really get why. Since I don't have any Windows setup right now, it might be a little while before I am able to set one up and debug the issues. If anyone is able to look into it, I would really appreciate it.
@henryiii you use Windows, right? Perhaps you could help out if you have time.

FFY00:
It was colorama. It's a bit weird since this test was passing previously.

henryiii:
> The tests are failing on Windows and I don't really get why. Since I don't have any Windows setup right now, it might be a little while before I am able to set one up and debug the issues. If anyone is able to look into it, I would really appreciate it.
> @henryiii you use Windows, right? Perhaps you could help out if you have time.

Sorry, wasn't around a computer much this weekend. I technically have a Windows box I boot up once in a while when I'm stuck debugging Windows issues in various packages. I'm perfectly fine to keep that a secret, though. 😅

PR #336, opened 2021-08-01
build: switch to tomli for TOML v1 compliant parser

### LINKED ISSUE #308: Consider switching to TOML v1 compliant parser
With several other projects (Poetry, by virtue of using tomlkit, [ward](https://github.com/darrenburns/ward/pull/267)) having switched or soon to switch ([pip](https://github.com/pypa/pip/pull/10035), [Black](https://github.com/psf/black/pull/2301)) to a parser able to handle v1 of the spec, we might want to consider doing the same (or assisting with updating toml).

FFY00:
Fixes #308

### DIFF
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -17,10 +17,35 @@ import warnings
 import zipfile
 
 from collections import OrderedDict
-from typing import AbstractSet, Any, Callable, Dict, Iterator, Mapping, Optional, Sequence, Set, Tuple, Type, Union
+from typing import (
+    AbstractSet,
+    Any,
+    Callable,
+    Dict,
+    Iterator,
+    Mapping,
+    MutableMapping,
+    Optional,
+    Sequence,
+    Set,
+    Tuple,
+    Type,
+    Union,
+)
 
 import pep517.wrappers
-import toml
+
+
+TOMLDecodeError: Type[Exception]
+toml_loads: Callable[[str], MutableMapping[str, Any]]
+
+
+try:
+    from tomli import TOMLDecodeError
+    from tomli import loads as toml_loads
+except ModuleNotFoundError:  # pragma: no cover
+    from toml import TomlDecodeError as TOMLDecodeError  # type: ignore
+    from toml import loads as toml_loads  # type: ignore
 
 
 RunnerType = Callable[[Sequence[str], Optional[str], Optional[Mapping[str, str]]], None]
@@ -185,13 +210,13 @@ class ProjectBuilder:
         spec_file = os.path.join(srcdir, 'pyproject.toml')
 
         try:
-            with open(spec_file, encoding='UTF-8') as f:
-                spec = toml.load(f)
+            with open(spec_file, 'rb') as f:
+                spec = toml_loads(f.read().decode())
         except FileNotFoundError:
             spec = {}
         except PermissionError as e:
             raise BuildException(f"{e.strerror}: '{e.filename}' ")
-        except toml.TomlDecodeError as e:
+        except TOMLDecodeError as e:
             raise BuildException(f'Failed to parse {spec_file}: {e} ')
 
         build_system = spec.get('build-system')
--- a/tests/test_projectbuilder.py
+++ b/tests/test_projectbuilder.py
@@ -2,6 +2,7 @@
 
 
 import copy
+import importlib
 import os
 import sys
 import textwrap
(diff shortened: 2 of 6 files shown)


layday:
* Reading from a text stream is deprecated in tomli and will generate a warning which can be reinterpreted as an error in test suites.
* Please replace `toml` with `tomli` in `tests/constraints.txt`.
* Remove `types-toml` from `setup.cfg`.

src/build/__init__.py
```diff
@@ -20,7 +20,17 @@
...
+
+try:
+    import tomli as toml
+
+    TOMLDecodeError: Type[Exception] = toml.TOMLDecodeError
+except ModuleNotFoundError:  # pragma: no cover
+    import toml as _toml
+
+    toml = _toml  # type: ignore
+    TOMLDecodeError = _toml.TomlDecodeError
```
hukkin, line 33:
```suggestion
try:
    from tomli import loads as toml_loads
    from tomli import TOMLDecodeError
except ImportError:  # pragma: no cover
    from toml import loads as toml_loads
    from toml import TomlDecodeError as TOMLDecodeError
```

Something like this plus changing these lines below
```python
with open(spec_file, encoding='UTF-8') as f:
    spec = toml.load(f)
```
to
```python
with open(spec_file, 'rb') as f:
        spec = toml_loads(f.read().decode())
```
will
1. avoid a [deprecation warning](https://github.com/hukkin/tomli/blob/6c3ae6b0836d4af2d3b434b594312d78af0b116a/tomli/_parser.py#L71) in tomli
2. avoid making illegal line ending conversions when opening the file

FFY00: I generally prefer to keep full imports, but since we were already doing it for the error, we can have `loads` too.

src/build/__init__.py
```diff
@@ -17,10 +17,35 @@
...
+TOMLDecodeError: Type[Exception]
+toml_loads: Callable[[str], MutableMapping[str, Any]]
+
+
+try:
+    from tomli import TOMLDecodeError
+    from tomli import loads as toml_loads
+except ModuleNotFoundError:  # pragma: no cover
+    from toml import TomlDecodeError as TOMLDecodeError  # type: ignore
+    from toml import loads as toml_loads  # type: ignore
```
gaborbernat, line 48: Eh, I'm not that eager on this. I'd prefer to just drop this. If we must keep on this we should at least validate that build works with toml and not tomli by adding a test env for this :shrug:

gaborbernat (changes requested):

src/build/__init__.py
layday: It does seem a bit pointless considering pep517 has switched to tomli for Python 3.6 and above.

FFY00: This is still relevant for bootstraping. Building with older version is common if it makes things eaier.

The code won't affect any use that installed the package the normal way, just people provisioning environments.

gaborbernat: But considering pep517 switched to tomli, and we depend on tomli wouldn't people need to provision tomli anyway? If they only provide toml, pep517 will break...

tests/test_projectbuilder.py
```diff
@@ -520,3 +521,28 @@ def test_metadata_invalid_wheel(tmp_dir, test_bad_wheel_path):
 
     with pytest.raises(ValueError, match='Invalid wheel'):
         builder.metadata_path(tmp_dir)
+
+
+@pytest.fixture()
+def mock_tomli_not_available(mocker):
+    loads = mocker.patch('tomli.loads')
+    tomli = sys.modules['tomli']
+    try:
```
gaborbernat, line 530: Here you want

```python
patch.dict(sys.modules, {'tomli': None})
```

gaborbernat (changes requested):

src/build/__init__.py
FFY00: Sure, but currently it's easier to build an older version on pep517 and then update it instead of building tomli directly when bootstraping.

gaborbernat: I see we were missing this piece of information. Perhaps worth adding this as comment in the catch 👍

layday: Can't they also bootstrap an older version of build and then update it, what's the difference?

tests/test_projectbuilder.py
FFY00: Thanks, I did not know about it :stuck_out_tongue:

src/build/__init__.py
FFY00: > I see we were missing this piece of information. Perhaps worth adding this as comment in the catch +1

I improved the comment in `setup.cfg`, let me know if you want me to clarify it more, and if so how/what exactly.

> Can't they also bootstrap an older version of build and then update it, what's the difference?

They won't get bugfixes or improvements. As long as we have a usable version of pep517, things should be good, I find it much more likely that bugs will happen on our part.
I'd also like for pep517 to fallback to toml, which should be pretty straightforward as they have the code already because they still support Python < 3.6.

FFY00:
Of course :facepalm:

FFY00:
Nevermind, the teardown does not run after I yield the other fixture. Does anyone have any proposal on how to do this cleanly? pytest-mock is very much against giving us access over when the mock is removed, we would use `unittest.mock` directly.

layday:
You are reloading the module before sys.modules is unpatched.

layday:
Okay, I tried running the tests locally and I think I see what's going on. Try calling `stopall` on the mocker:

```py
@pytest.fixture
def mock_tomli_not_available(mocker):
    loads = mocker.patch('tomli.loads')
    mocker.patch.dict(sys.modules, {'tomli': None})
    importlib.reload(build)
    try:
        yield
    finally:
        loads.assert_not_called()
        mocker.stopall()
        importlib.reload(build)
```

FFY00:
That works too, and our use-case is simple enough that using stopall is fine.

layday (approved):

gaborbernat (approved):

FFY00:
:blush:

PR #337, opened 2021-08-03
tests: enable pytest strict mode

graingert:


layday:
`test_colors_conflict` appears to be erroring because of `filterwarnings = error`.

gaborbernat (approved):

layday:
Shouldn't the filtering be done in `test_colors_conflict` instead of the fixture?

graingert:
@gaborbernat you pressed the wrong approval button, the CI needs to run xD

layday:
I would prefer if we used a monkeypatching context. Remove the filter from the fixture and in `test_colors_conflict`:

```py
with monkeypatch.context() as m:
    m.setenv('NO_COLOR', '')
    ,.setenv('FORCE_COLOR', '')

    with pytest.warns(Warning, match='Both NO_COLOR and FORCE_COLOR environment variables are set, disabling color'):
        importlib.reload(build.__main__)
```

graingert:
> I would prefer if we used a monkeypatching context. Remove the filter from the fixture and in `test_colors_conflict`:
> 
> ```python
> with monkeypatch.context() as m:
>     m.setenv('NO_COLOR', '')
>     m.setenv('FORCE_COLOR', '')
> 
>     with pytest.warns(Warning, match='Both NO_COLOR and FORCE_COLOR environment variables are set, disabling color'):
>         importlib.reload(build.__main__)
> ```

That would make sense in a test case, but not in the cleanup

layday:
That is for the test case.

graingert:
The importlib reload will happen again in the fixture and raise a new warning

layday:
That's what the monkeypatch context is there for, it won't raise a warning outside of the context.

graingert:
Oh! I see what you mean now

layday (approved):

layday:
Great, thanks! Will let @FFY00 take a look before merging.

FFY00 (approved):
LGTM, thanks! :blush:

PR #339, opened 2021-08-04
build: restore compatibility with Python 3.6 and 3.7

### LINKED ISSUE #338: v0.6 is not compatible with python 3.6 due to use of `stacklevel` in log calls
# Problem

V0.6 fails on python 3.6 with 

<details>
  <summary>
TypeError: _log() got an unexpected keyword argument 'stacklevel'
</summary>
 
``` python
/home/jenkins/workspace/_core_dependabot_pip_build-0.6.0/project/inmanta-core/.tox/py36/lib/python3.6/site-packages/build/env.py:100: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

message = 'Creating virtualenv isolated environment...'

    @staticmethod
    def log(message: str) -> None:
        """
        Prints message
    
        The default implementation uses the logging module but this function can be
        overwritten by users to have a different implementation.
    
        :param msg: Message to output
        """
>       _logger.log(logging.INFO, message, stacklevel=2)

/home/jenkins/workspace/_core_dependabot_pip_build-0.6.0/project/inmanta-core/.tox/py36/lib/python3.6/site-packages/build/env.py:138: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <Logger build.env (DEBUG)>, level = 20
msg = 'Creating virtualenv isolated environment...', args = ()
kwargs = {'stacklevel': 2}

    def log(self, level, msg, *args, **kwargs):
        """
        Log 'msg % args' with the integer severity 'level'.
    
        To pass exception information, use the keyword argument exc_info with
        a true value, e.g.
    
        logger.log(level, "We have a %s", "mysterious problem", exc_info=1)
        """
        if not isinstance(level, int):
            if raiseExceptions:
                raise TypeError("level must be an integer")
            else:
                return
        if self.isEnabledFor(level):
>           self._log(level, msg, args, **kwargs)
E           TypeError: _log() got an unexpected keyword argument 'stacklevel'
```
</details>

# Cause 

this change https://github.com/pypa/build/commit/32398749d06b19bd2cf21be278e8ceb871411dbb#diff-4f1bb46d65de7fbb1573dfc2a2bca9f63c7b03d55e74c53339f0c8ee4fa054d8R129
makes use of the stacklevel argument, which is only supported from python 3.8
[ref](https://docs.python.org/3/library/logging.html#logging.Logger.debug)

wouterdb:
fixes #338

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -135,7 +135,7 @@ class IsolatedEnvBuilder:
 
         :param msg: Message to output
         """
-        _logger.log(logging.INFO, message, stacklevel=2)
+        _logger.log(logging.INFO, message)
 
 
 class _IsolatedEnvVenvPip(IsolatedEnv):
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -135,7 +135,10 @@ class IsolatedEnvBuilder:
 
         :param msg: Message to output
         """
-        _logger.log(logging.INFO, message)
+        if sys.version_info >= (3, 8):
+            _logger.log(logging.INFO, message, stacklevel=2)
+        else:
+            _logger.log(logging.INFO, message)
 
 
 class _IsolatedEnvVenvPip(IsolatedEnv):
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -135,12 +135,10 @@ class IsolatedEnvBuilder:
 
         :param msg: Message to output
         """
-        # use %s as message to force formatting to run
-        # this prevents failure if the message contains a % sign
         if sys.version_info >= (3, 8):
-            _logger.log(logging.INFO, "%s", message, stacklevel=2)
+            _logger.log(logging.INFO, message, stacklevel=2)
         else:
-            _logger.log(logging.INFO, "%s", message)
+            _logger.log(logging.INFO, message)
 
 
 class _IsolatedEnvVenvPip(IsolatedEnv):
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -135,10 +135,12 @@ class IsolatedEnvBuilder:
 
         :param msg: Message to output
         """
+        # use %s as message to force formatting to run
+        # this prevents failure if the message contains a % sign
         if sys.version_info >= (3, 8):
-            _logger.log(logging.INFO, message, stacklevel=2)
+            _logger.log(logging.INFO, "%s", message, stacklevel=2)
         else:
-            _logger.log(logging.INFO, message)
+            _logger.log(logging.INFO, "%s", message)
 
 
 class _IsolatedEnvVenvPip(IsolatedEnv):
--- a/tests/test_env.py
+++ b/tests/test_env.py
@@ -28,9 +28,7 @@ def test_isolation():
 
 
 @pytest.mark.isolated
-def test_isolated_environment_install(mocker, caplog):
-    # ensure the log code is used
-    caplog.set_level(logging.INFO)
+def test_isolated_environment_install(mocker):
     with build.env.IsolatedEnvBuilder() as env:
         mocker.patch('build.env._subprocess')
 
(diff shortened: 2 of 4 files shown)


src/build/env.py
```diff
@@ -135,7 +135,7 @@ def log(message: str) -> None:
 
         :param msg: Message to output
         """
-        _logger.log(logging.INFO, message, stacklevel=2)
+        _logger.log(logging.INFO, message)
```
gaborbernat, line 138: Rather than blindly removing it can we keep it for 3.8+?

gaborbernat (changes requested):

gaborbernat (commented):
Can you add a test that validates/catches this issue?

src/build/env.py
```diff
@@ -135,7 +135,12 @@ def log(message: str) -> None:
 
         :param msg: Message to output
         """
-        _logger.log(logging.INFO, message, stacklevel=2)
+        # use %s as message to force formatting to run
+        # this prevents failure if the message contains a % sign
+        if sys.version_info >= (3, 8):
+            _logger.log(logging.INFO, "%s", message, stacklevel=2)
```
wouterdb, line 141: I spotted a second problem, which I also fixed

layday:
How are the tests passing for 3.6 and 3.7?

src/build/env.py
wouterdb: (never mind, problem doesn't occur in this setup)

wouterdb:
> How are the tests passing for 3.6 and 3.7?

if the log level is below INFO, the problem doesn't manifest.

If you would execute the tests with `--log-cli-level=INFO` they would fail

layday:
Okay, we are not testing the logging directly.  We are only testing the CLI's output and the CLI overloads the log methods.

This should've also been caught by the type checker unless the stubs in the typesheds are wrong.

layday:
mypy selects the correct overload but does not produce an error when it's called with `stacklevel` because `Logger.log` accepts `**kwargs` (see https://github.com/python/typeshed/blob/b2e429cd41347ce48a96feaa1681726d8f77fbfb/stdlib/logging/__init__.pyi#L218).  This is incorrect.

tests/test_env.py
```diff
@@ -27,7 +28,9 @@ def test_isolation():
 
 
 @pytest.mark.isolated
-def test_isolated_environment_install(mocker):
+def test_isolated_environment_install(mocker, caplog):
+    # ensure the log code is used
+    caplog.set_level(logging.INFO)
```
FFY00, line 33: Instead of doing this, could you add a test in `test_projectbuilder.py` to verify the default `ProjectBuilder.log` implementation is correct?

FFY00: Sorry, I end up doing it myself as I has the time and I want this fixed ASAP :sweat_smile:

FFY00:
@gaborbernat or @layday please take a look.

gaborbernat (commented):
Where should we see the stacklevel=2 impact? 🤔

FFY00:
`record.lineno`, it makes the log be registered to the `ProjectBuilder.log` caller, instead of the line where we actually call the log. Right now, I am not sure how to write tests for this without having to fix them every time we touch the code. The best we could do is mocking the call and make sure `stacklevel` is passed, although that is kinda just duplicating the code. I would merge this as-is and if welcome better tests in the future.

gaborbernat (commented):
I'd prefer to have a test that validates the stack level, it's alright that we need to change the test when line numbers change.

FFY00:
Sure, but if this becomes an annoyance, we will remove it.

gaborbernat (approved):

henryiii:
> mypy selects the correct overload but does not produce an error when it's called with `stacklevel` because `Logger.log` accepts `**kwargs` (see https://github.com/python/typeshed/blob/b2e429cd41347ce48a96feaa1681726d8f77fbfb/stdlib/logging/__init__.pyi#L218). This is incorrect.

Interesting, you pass keyword arguments via dict to *args, so yes, this is true - it's only used to pass things through to `self._log`, which does not accept arbitrary keyword arguments. I can make a PR to type shed unless you plan to do it, @layday. Honestly, this sort of shortcut seems like it should also be made explicit in the stdlib, too.

layday:
> I can make a PR to type shed unless you plan to do it, @layday.

I'd appreciate it if you could, might not get around to it for a few days.

PR #340, opened 2021-08-05
util: add project_wheel_metadata

FFY00:

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -60,6 +60,7 @@ test =
     toml>=0.10.0
     wheel>=0.36.0
 typing =
+    importlib-metadata>=4.6.4
     mypy==0.910
     typing-extensions>=3.7.4.3
 virtualenv =
--- a/src/build/util.py
+++ b/src/build/util.py
@@ -1,0 +1,60 @@
+# SPDX-License-Identifier: MIT
+
+import os
+import pathlib
+import sys
+import tempfile
+
+from typing import Union
+
+import pep517
+
+import build
+import build.env
+
+
+if sys.version_info >= (3, 8):
+    import importlib.metadata as importlib_metadata
+else:
+    import importlib_metadata
+
+
+def _project_wheel_metadata(builder: build.ProjectBuilder) -> 'importlib_metadata.PackageMetadata':
+    with tempfile.TemporaryDirectory() as tmpdir:
+        path = pathlib.Path(builder.metadata_path(tmpdir))
+        # https://github.com/python/importlib_metadata/pull/343
+        return importlib_metadata.PathDistribution(path).metadata  # type: ignore
+
+
+def project_wheel_metadata(
+    srcdir: Union[str, 'os.PathLike[str]'],
+    isolated: bool = True,
+) -> 'importlib_metadata.PackageMetadata':
+    """
+    Return the wheel metadata for a project.
+
+    Uses the ``prepare_metadata_for_build_wheel`` hook if availablable,
+    otherwise ``build_wheel``.
+
+    :param srcdir: Project source directory
+    :param isolated: Wether or not to run invoke the backend in the current
+                     environment or to create an isolated one and invoke it
+                     there.
+    """
+    builder = build.ProjectBuilder(
+        os.fspath(srcdir),
+        runner=pep517.quiet_subprocess_runner,
+    )
+
+    if not isolated:
+        return _project_wheel_metadata(builder)
+
+    with build.env.IsolatedEnvBuilder() as env:
+        builder.python_executable = env.executable
+        builder.scripts_dir = env.scripts_dir
+        env.install(builder.build_system_requires)
+        env.install(builder.get_requires_for_build('wheel'))
+        return _project_wheel_metadata(builder)
+
+
+__all__ = ('project_wheel_metadata',)
--- a/tests/packages/test-metadata/pyproject.toml
+++ b/tests/packages/test-metadata/pyproject.toml
@@ -1,0 +1,14 @@
+[build-system]
+requires = []
+build-backend = 'backend'
+backend-path = ['.']
+
+[project]
+name = 'test-metadata'
+version = '1.0.0'
+description = 'hello!'
+
+[tool.black]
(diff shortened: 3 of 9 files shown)


setup.cfg
```diff
@@ -86,6 +87,7 @@ extend-ignore = E203
 [mypy]
 ignore_missing_imports = True
 strict = True
+implicit_reexport = True
```
layday, line 90: This is gonna break `--strict` downstream. Just add `__all__` in `build.util`.

src/build/util.py
```diff
@@ -0,0 +1,64 @@
...
+    import importlib.metadata as importlib_metadata
+else:
+    import importlib_metadata
+
+
+if typing.TYPE_CHECKING:
+    from importlib_metadata import PackageMetadata, SimplePath
+
+
+def _project_wheel_metadata(builder: build.ProjectBuilder) -> 'PackageMetadata':
```
layday, line 27: Does it return `PackageMetadata` if you don't have `importlib-metadata >= 4.1.0`?

src/build/util.py
```diff
@@ -0,0 +1,64 @@
...
+    with tempfile.TemporaryDirectory() as tmpdir:
+        path = pathlib.Path(builder.metadata_path(tmpdir))
+        if typing.TYPE_CHECKING:
+            cast_path = typing.cast(SimplePath, path)
```
layday, line 31: Why? `Path` should satisfy the `SimplePath` protocol.

src/build/util.py
```diff
@@ -0,0 +1,64 @@
...
+        runner=pep517.quiet_subprocess_runner,
+    )
+
+    if not isolated:
+        return _project_wheel_metadata(builder)
+
+    with build.env.IsolatedEnvBuilder() as env:
+        builder.python_executable = env.executable
+        builder.scripts_dir = env.scripts_dir
+        env.install(builder.build_system_requires)
```
layday, line 63: Also need to call `get_requires_for_build_wheel`.

gaborbernat: Add a test for this by using an in place backend that tries to import something required via get_requires_for_build_wheel

src/build/util.py
layday: https://github.com/python/importlib_metadata/issues/334

setup.cfg
FFY00: This is for `SimplePath`, which is not explicitly re-exported in importlib.

src/build/util.py
FFY00: This is a case of me shooting myself in the foot :upside_down_face:. https://github.com/python/importlib_metadata/commit/fa20a0a330175502b61de89e5dd8e10f3f77ede7

setup.cfg
gaborbernat: If so you want to set that option for only that lib, not globally. Or add a type ignore at the location of the import.

layday: We shouldn't need to import `SimplePath` after https://github.com/python/importlib_metadata/pull/335 is merged.

layday:
There's a new version of `importlib-metadata` out if you have time to revisit this PR @FFY00.

src/build/util.py
FFY00: Done.

src/build/util.py
```diff
@@ -0,0 +1,61 @@
...
+
+from typing import Union
+
+import pep517
+
+import build
+import build.env
+
+
+if sys.version_info >= (3, 8) and not typing.TYPE_CHECKING:
```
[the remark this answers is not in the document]
FFY00: This protocol is still not fixed in importlib_metadata (needs https://github.com/python/importlib_metadata/pull/342, see https://github.com/python/importlib_metadata/pull/343), so I think we should just go ahead with the `# type: ignore` as we know the code is definitely correct.

src/build/util.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def _project_wheel_metadata(builder: build.ProjectBuilder) -> 'importlib_metadata.PackageMetadata':
+    with tempfile.TemporaryDirectory() as tmpdir:
+        path = pathlib.Path(builder.metadata_path(tmpdir))
+        # https://github.com/python/importlib_metadata/pull/343
+        return importlib_metadata.PathDistribution(path).metadata  # type: ignore
+
+
+def project_wheel_metadata(
```
gaborbernat, line 29: Add an ``__all__``  with this in it to mark this method as public :+1: (and hides importlib_metadata)

tests/packages/test-metadata/pyproject.toml
```diff
@@ -0,0 +1,14 @@
+[build-system]
+requires = []
+build-backend = 'backend'
+backend-path = ['.']
+
+[project]
```
gaborbernat, line 6: By not making this tool.build you're technically violating pep-518... :+1:

layday: These are PEP 621 metadata.

src/build/util.py
henryiii: I also usually recommend:

```python
def __dir__() -> Tuple[Any, ...]:
    return __all__
```

This fixes tab completion the REPL or IPython in Python 3.7+ to only show public methods.

FFY00: I am gonna do `__all__` to keep in sync with the other modules, and then you can open a PR moving to that form.

FFY00:
Does anyone have other comments?

src/build/util.py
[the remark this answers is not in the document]
henryiii: You should get a message about an unnecessary type ignore when this does get fixed, so don't see any harm.

setup.cfg
henryiii: I think we should move to pyproject.toml mypy config, it's much cleaner and makes per-lib settings prettier if we need them. Not in this PR, though.

FFY00: Sure, feel free to open a PR.

layday (approved):

layday:
LGTM!

PR #351, opened 2021-09-14
tests: integration: use PEP 440 version as setuptools-scm fallback

FFY00:
I guess it now dropped support for legacy versions.


FFY00:
This is a fairly simple change, so I am merging.

PR #360, opened 2021-09-28
env: change `IsolatedEnv.install` parameter type

layday:
`_IsolatedEnvVenvPip` expects this not to be an iterator.


gaborbernat (approved):

FFY00 (approved):
Please add a changelog entry. Better safe than sorry.

layday:
I'll fix this in #361 instead.

PR #361, opened 2021-09-28
env: refactor `IsolatedEnv`

layday:
The `IsolatedEnv` is made responsible for env creation to allow pip
to implement custom env creation logic.

~~`IsolatedEnvBuilder` takes an `IsolatedEnv` class as an argument
so that bringing your own `IsolatedEnv` won't require re-implementing
the builder.~~

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -65,5 +65,5 @@ html_title = f'build {version}'
 # html_static_path = ['_static']
 
 autoclass_content = 'both'
-
+autodoc_member_order = 'bysource'
 autodoc_preserve_defaults = True
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -65,3 +65,5 @@ html_title = f'build {version}'
 # html_static_path = ['_static']
 
 autoclass_content = 'both'
+
+autodoc_preserve_defaults = True
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,6 +31,7 @@ install_requires =
     tomli>=1.0.0 # toml can be used instead -- in case it makes bootstrapping easier
     colorama;os_name == "nt" # not actually a runtime dependency, only supplied as there is not "recommended dependency" support
     importlib-metadata>=0.22;python_version < "3.8"
+    typing-extensions>=3.8;python_version < "3.8"
 python_requires = >=3.6
 package_dir =
     =src
@@ -64,7 +65,6 @@ test =
 typing =
     importlib-metadata>=4.6.4
     mypy==0.910
-    typing-extensions>=3.7.4.3
 virtualenv =
     virtualenv>=20.0.35
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,7 +31,6 @@ install_requires =
     tomli>=1.0.0 # toml can be used instead -- in case it makes bootstrapping easier
     colorama;os_name == "nt" # not actually a runtime dependency, only supplied as there is not "recommended dependency" support
     importlib-metadata>=0.22;python_version < "3.8"
-    typing-extensions>=3.8;python_version < "3.8"
 python_requires = >=3.6
 package_dir =
     =src
@@ -65,6 +64,7 @@ test =
 typing =
     importlib-metadata>=4.6.4
     mypy==0.910
+    typing-extensions>=3.8;python_version < "3.8"
 virtualenv =
     virtualenv>=20.0.35
 
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -19,7 +19,6 @@ import zipfile
 
 from typing import (
     TYPE_CHECKING,
-    AbstractSet,
     Any,
     Callable,
     Dict,
@@ -27,7 +26,6 @@ from typing import (
     Mapping,
     MutableMapping,
     Optional,
-    Sequence,
     Set,
     Tuple,
     Type,
@@ -36,10 +34,12 @@ from typing import (
 
 import pep517.wrappers
 
+from . import env
+from ._helpers import ConfigSettingsType, PathType, check_dependency
+
 
 if TYPE_CHECKING:
-    # Avoid import cycle.
-    from . import env
+    from ._helpers import RunnerType
 
 
 TOMLDecodeError: Type[Exception]
@@ -54,27 +54,7 @@ except ModuleNotFoundError:  # pragma: no cover
     from toml import loads as toml_loads  # type: ignore
 
 
-ConfigSettingsType = Mapping[str, Union[str, Sequence[str]]]
-PathType = Union[str, 'os.PathLike[str]']
 _ExcInfoType = Union[Tuple[Type[BaseException], BaseException, types.TracebackType], Tuple[None, None, None]]
-
-
-if TYPE_CHECKING:
-    from typing_extensions import Protocol
-
-    class RunnerType(Protocol):
-        def __call__(
-            self, cmd: Sequence[str], cwd: Optional[PathType] = None, extra_environ: Optional[Mapping[str, str]] = None
-        ) -> None:
-            """
-            Run a command in a Python subprocess.
-
-            The parameters mirror those of ``subprocess.run``.
-
-            :param cmd: The command to execute
-            :param cwd: The working directory
-            :param extra_environ: Variables to be exported to the environment
-            """
 
 
 _WHEEL_NAME_REGEX = re.compile(
@@ -143,49 +123,6 @@ def _working_directory(path: str) -> Iterator[None]:
         yield
     finally:
         os.chdir(current)
-
-
-def check_dependency(
-    req_string: str, ancestral_req_strings: Tuple[str, ...] = (), parent_extras: AbstractSet[str] = frozenset()
-) -> Iterator[Tuple[str, ...]]:
-    """
-    Verify that a dependency and all of its dependencies are met.
-
-    :param req_string: Requirement string
-    :param parent_extras: Extras (eg. "test" in myproject[test])
-    :yields: Unmet dependencies
-    """
-    import packaging.requirements
-
-    if sys.version_info >= (3, 8):
-        import importlib.metadata as importlib_metadata
-    else:
-        import importlib_metadata
-
-    req = packaging.requirements.Requirement(req_string)
-
-    if req.marker:
-        extras = frozenset(('',)).union(parent_extras)
-        # a requirement can have multiple extras but ``evaluate`` can
-        # only check one at a time.
-        if all(not req.marker.evaluate(environment={'extra': e}) for e in extras):
-            # if the marker conditions are not met, we pretend that the
-            # dependency is satisfied.
-            return
-
-    try:
-        dist = importlib_metadata.distribution(req.name)  # type: ignore[no-untyped-call]
-    except importlib_metadata.PackageNotFoundError:
-        # dependency is not installed in the environment.
-        yield ancestral_req_strings + (req_string,)
-    else:
-        if req.specifier and not req.specifier.contains(dist.version, prereleases=True):
-            # the installed version is incompatible.
-            yield ancestral_req_strings + (req_string,)
-        elif dist.requires:
-            for other_req_string in dist.requires:
-                # yields transitive dependencies that are not satisfied.
-                yield from check_dependency(other_req_string, ancestral_req_strings + (req_string,), req.extras)
 
 
 def _parse_source_dir(source_dir: PathType) -> str:
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -17,20 +17,7 @@ import types
 import warnings
 import zipfile
 
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    Callable,
-    Dict,
-    Iterator,
-    Mapping,
-    MutableMapping,
-    Optional,
-    Set,
-    Tuple,
-    Type,
-    Union,
-)
+from typing import TYPE_CHECKING, Any, Callable, Dict, Iterator, Mapping, MutableMapping, Optional, Set, Tuple, Type, Union
 
 import pep517.wrappers
 
@@ -39,7 +26,7 @@ from ._helpers import ConfigSettingsType, PathType, check_dependency
 
 
 if TYPE_CHECKING:
-    from ._helpers import RunnerType
+    from ._helpers import Distribution, RunnerType, WheelDistribution
 
 
 TOMLDecodeError: Type[Exception]
@@ -260,12 +247,14 @@ class ProjectBuilder:
         """
         return self._requires
 
-    def get_requires_for_build(self, distribution: str, config_settings: Optional[ConfigSettingsType] = None) -> Set[str]:
+    def get_requires_for_build(
+        self, distribution: 'Distribution', config_settings: Optional[ConfigSettingsType] = None
+    ) -> Set[str]:
         """
         Get the build dependencies requested by the backend for
         a given distribution.
 
-        :param distribution: Distribution (``sdist`` or ``wheel``)
+        :param distribution: Distribution to build
         :param config_settings: Config settings passed to the backend
         """
         self.log(f'Getting dependencies for {distribution}...')
@@ -276,14 +265,14 @@ class ProjectBuilder:
             return set(get_requires(config_settings))
 
     def check_dependencies(
-        self, distribution: str, config_settings: Optional[ConfigSettingsType] = None
+        self, distribution: 'Distribution', config_settings: Optional[ConfigSettingsType] = None
     ) -> Set[Tuple[str, ...]]:
         """
         Check that the :attr:`build_system_requires` and :meth:`get_requires_for_build`
         dependencies for a given distribution are satisfied and return the dependency
         chain of those which aren't.  The unmet dependency is the last value in the chain.
 
-        :param distribution: Distribution (``sdist`` or ``wheel``)
+        :param distribution: Distribution to build
         :param config_settings: Config settings passed to the backend
         :returns: Unmet dependencies in the PEP 508 format
         """
@@ -291,12 +280,15 @@ class ProjectBuilder:
         return {u for d in dependencies for u in check_dependency(d)}
 
     def prepare(
-        self, distribution: str, output_directory: PathType, config_settings: Optional[ConfigSettingsType] = None
+        self,
+        distribution: 'WheelDistribution',
+        output_directory: PathType,
+        config_settings: Optional[ConfigSettingsType] = None,
     ) -> Optional[str]:
         """
         Prepare metadata for a distribution.
 
-        :param distribution: Distribution (must be ``wheel``)
+        :param distribution: Distribution to build
         :param output_directory: Directory to put the prepared metadata in
         :param config_settings: Config settings passed to the backend
         :returns: The path of the metadata directory
@@ -316,7 +308,7 @@ class ProjectBuilder:
 
     def build(
         self,
-        distribution: str,
+        distribution: 'Distribution',
         output_directory: PathType,
         config_settings: Optional[ConfigSettingsType] = None,
         metadata_directory: Optional[str] = None,
@@ -324,7 +316,7 @@ class ProjectBuilder:
         """
         Build a distribution.
 
-        :param distribution: Distribution (``sdist`` or ``wheel``)
+        :param distribution: Distribution to build
         :param output_directory: Directory to put the built distribution in
         :param config_settings: Config settings passed to the backend
         :param metadata_directory: If provided, should be the return value of a
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -479,7 +479,7 @@ class ProjectBuilder:
             _logger.log(logging.INFO, message)
 
 
-__all__ = (
+__all__ = [
     '__version__',
     'ConfigSettingsType',
     'RunnerType',
@@ -489,4 +489,4 @@ __all__ = (
     'TypoWarning',
     'check_dependency',
     'ProjectBuilder',
-)
+]
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -73,7 +73,7 @@ _DEFAULT_BACKEND = {
 }
 
 
-_logger = logging.getLogger('build')
+_logger = logging.getLogger(__name__)
 
 
 class BuildException(Exception):
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -78,13 +78,13 @@ _logger = logging.getLogger(__name__)
 
 class BuildException(Exception):
     """
-    Exception raised by ProjectBuilder
+    Exception raised by :class:`ProjectBuilder`.
     """
 
 
 class BuildBackendException(Exception):
     """
-    Exception raised when the backend fails
+    Exception raised when a backend operation fails.
     """
 
     def __init__(
@@ -112,7 +112,7 @@ class BuildSystemTableValidationError(BuildException):
 
 class TypoWarning(Warning):
     """
-    Warning raised when a potential typo is found
+    Warning raised when a possible typo is found.
     """
 
 
@@ -240,8 +240,9 @@ class ProjectBuilder:
         runner: RunnerType = pep517.wrappers.default_subprocess_runner,
     ) -> None:
         """
-        :param srcdir: The source directory
-        :param python_executable: The python executable where the backend lives
+        :param srcdir: The project source directory
+        :param python_executable: Path of Python executable used to invoke
+            PEP 517 hooks
         :param runner: An alternative runner for backend subprocesses
 
         The 'runner', if provided, must accept the following arguments:
@@ -309,20 +310,18 @@ class ProjectBuilder:
     @property
     def build_system_requires(self) -> Set[str]:
         """
-        The dependencies defined in the ``pyproject.toml``'s
-        ``build-system.requires`` field or the default build dependencies
-        if ``pyproject.toml`` is missing or ``build-system`` is undefined.
+        The dependencies specified in the project's ``build-system.requires``
+        field or the default build dependencies if unspecified.
         """
         return self._requires
 
     def get_requires_for_build(self, distribution: str, config_settings: Optional[ConfigSettingsType] = None) -> Set[str]:
         """
-        Return the dependencies defined by the backend in addition to
-        :attr:`build_system_requires` for a given distribution.
+        Get the build dependencies requested by the backend for
+        a given distribution.
 
-        :param distribution: Distribution to get the dependencies of
-            (``sdist`` or ``wheel``)
-        :param config_settings: Config settings for the build backend
+        :param distribution: Distribution (``sdist`` or ``wheel``)
+        :param config_settings: Config settings passed to the backend
         """
         self.log(f'Getting dependencies for {distribution}...')
         hook_name = f'get_requires_for_build_{distribution}'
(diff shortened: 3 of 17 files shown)


src/build/env.py
```diff
@@ -82,10 +90,13 @@ def _subprocess(cmd: List[str]) -> None:
 
 
 class IsolatedEnvBuilder:
-    """Builder object for isolated environments."""
+    """Create and dispose of isolated build environments."""
 
-    def __init__(self) -> None:
-        self._path: Optional[str] = None
+    def __init__(self, isolated_env_class: Optional[Type[IsolatedEnv]] = None) -> None:
```
uranusjr, line 95:
```suggestion
    def __init__(self, isolated_env_class: Type[IsolatedEnv] = _DefaultIsolatedEnv) -> None:
```

why not?

layday: Because `_DefaultIsolatedEnv` is defined underneath `IsolatedEnvBuilder` and I didn't wanna swap them around in the draft.

layday:
Yep, that's right.  Typical usage would be along the lines of:

```py
with build.env.IsolatedEnvBuilder(PipIsolatedEnv) as env:
    builder = build.ProjectBuilder(
        source_dir,
        python_executable=env.python_executable,
        scripts_dir=env.scripts_dir,
    )
    env.install(builder.build_system_requires)
    env.install(builder.get_requires_for_build('wheel'))
    builder.build('wheel', output_dir)
```

FFY00 (commented):
Overall, the API looks OK. @gaborbernat what do you think?

@layday don't forget to add an entry to the changelog about these breaking changes in the final version, it doesn't need to be anything too descriptive, just something on the lines of "reworked the IsolatedEnv interface and added an isolated_env_class argument to IsolatedEnvBuilder to make it easier to use cutsom environments".

layday:
Yep, I just wanted to garner some feedback before diving any deeper.

layday:
(I probably won't have time to work on this from next week so I'd appreciate any reviews over the weekend.)

setup.cfg
```diff
@@ -46,6 +46,7 @@ pipx.run =
 
 [options.extras_require]
 docs =
+    build[typing]
```
gaborbernat, line 49: Shouldn't we do here a self install? A reference to ourselves feels odd :thinking:

src/build/_helpers.py
```diff
@@ -0,0 +1,104 @@
...
+    subprocess.run(cmd, check=True, cwd=cwd, env=env)
+
+
+def quiet_subprocess_runner(
+    cmd: Sequence[str], cwd: Optional[PathType] = None, env: Optional[Mapping[str, str]] = None
+) -> None:
+    subprocess.run(cmd, check=True, cwd=cwd, env=env, stdout=subprocess.DEVNULL)
+
+
+def pep517ify_subprocess_runner(values: Union['RunnerType', Tuple['RunnerType', Optional[Mapping[str, str]]]]) -> Any:
```
gaborbernat, line 45: What does pep517ify means?

gaborbernat: Also any reason we did not annotate the return value correctly?

src/build/env.py
```diff
@@ -1,312 +1,313 @@
...
+        """
+        Create the isolated environment.
+
+        This method should be idempotent.
+        """
         raise NotImplementedError
 
     @abc.abstractmethod
-    def install(self, requirements: Iterable[str]) -> None:
+    def install_packages(self, requirements: Iterable[str]) -> None:
```
gaborbernat, line 45: Any reason this needed to change from install to install packages? I don't see any other install methods here to need to specify? What else could you install?

src/build/env.py
```diff
@@ -1,312 +1,313 @@
...
-        The default implementation uses the logging module but this function can be
-        overwritten by users to have a different implementation.
 
-        :param msg: Message to output
-        """
-        if sys.version_info >= (3, 8):
-            _logger.log(logging.INFO, message, stacklevel=2)
-        else:
-            _logger.log(logging.INFO, message)
+@_cache()
```
gaborbernat, line 134: This looks to me more cryptic than if we would just have put here the entire cache annotation.

src/build/env.py
```diff
@@ -1,312 +1,313 @@
...
-    @property
-    def scripts_dir(self) -> str:
-        return self._scripts_dir
 
-    def install(self, requirements: Iterable[str]) -> None:
-        """
-        Install packages from PEP 508 requirements in the isolated build environment.
+class _DefaultIsolatedEnv(IsolatedEnv):
+    """
+    An isolated environment class which combines venv and virtualenv with pip.
```
gaborbernat, line 169: I think you wanted to say here `or` and not `and`. Only one can be active at the same time.

src/build/env.py
```diff
@@ -1,312 +1,313 @@
...
-    with tempfile.NamedTemporaryFile(prefix='build-symlink-') as tmp_file:
-        dest = f'{tmp_file}-b'
         try:
-            os.symlink(tmp_file.name, dest)
-            os.unlink(dest)
-            return True
-        except (OSError, NotImplementedError, AttributeError):
-            return False
+            self._log(f'Installing build dependencies... ({", ".join(req_list)})')
+            self._run(
```
gaborbernat, line 229: This would read better if we'd declare first the command, and fit it in two lines. the comment can go after definining the cmd variable.

tox.ini
```diff
@@ -65,7 +65,7 @@ commands =
 
 [testenv:docs]
 description = build documentations
-basepython = python3.8
+basepython = python3.9
```
gaborbernat, line 68: Any reason we can't make it 3.10? :thinking:

src/build/_helpers.py
layday: It means to make the subprocess callback compatible with `pep517`. `pep517` expects there to be an `extra_environ` parameter but we need to be able to modify the environment, not just add values to it. I did not annotate the return value because `pep517` itself is not type hinted and mypy's support for function-altering decorator return types is not great and it would've probably involved one or several `cast`s or `type: ignore`s.

src/build/env.py
layday: It's a backport of the `functools` function called `cache`, which works the same way.

src/build/env.py
layday: Methods without a noun can be assumed to operate on the class, so an unqualified `install` would install the environment.

src/build/_helpers.py
gaborbernat: > mypy's support for function-altering decorator return types is not great and it would've probably involved one or several `cast`s or `type: ignore`s.

Such as, I never had any problems with this :thinking: 

Perhaps rename it to pep517_pkg_ify, because currently I thought you're referring to something in PEP-517 and not the package interface we're using.

src/build/env.py
gaborbernat: Ah, but then can we make it an actual backport, on 3.8/9 this exists let's use it from stdlib than our definition.

src/build/env.py
gaborbernat: Interesting, I never made that assumption myself :thinking:

tox.ini
layday: We should probably keep it synced with RTD, dunno if they support 3.10 yet.

src/build/env.py
layday: The `--pep-517` comment? But that's specific to that option?

src/build/env.py
layday: Alright.

src/build/env.py
gaborbernat: I know but it's kind obvious what you're talking about it, doesn't need IMHO to be right beside it.

tox.ini
gaborbernat: I see :thinking:

src/build/env.py
layday: Done.

src/build/env.py
layday: Done.

src/build/_helpers.py
layday: Done. It was actually okay because it's not used as a decorator.

setup.cfg
layday: It's supposed to be a supported/encouraged idiom (though it didn't work properly before pip 21.2).

layday: https://github.com/pypa/pip/issues/10393#issuecomment-942771191

gaborbernat: So from where is this build pulled? PyPi or the local source tree?

gaborbernat: So this is a way to make some extras depend on other extras :thinking: within the package?

src/build/_helpers.py
```diff
@@ -0,0 +1,117 @@
+import functools
```
gaborbernat, line 1: `_helpers` is the new `_utils`? :laughing:

gaborbernat (approved):

src/build/_helpers.py
layday: It seemed like the best place to put it 😛

setup.cfg
layday: > So from where is this build pulled? PyPi or the local source tree?

From the local source tree when installing locally (with `.`).



> So this is a way to make some extras depend on other extras 🤔 within the package?

Indeed.

layday:
@FFY00 Would you like to review before merging?

src/build/env.py
```diff
@@ -1,312 +1,322 @@
...
-        Install packages from PEP 508 requirements in the isolated build environment.
+        Install packages in the isolated environment.
 
-        :param requirements: PEP 508 requirements
+        :param requirements: PEP 508-style requirements
         """
         raise NotImplementedError
 
+    @abc.abstractmethod
+    def prepare_environ(self) -> Dict[str, str]:
```
layday, line 51: Possibly allow this to be `None` so that implementers aren't forced to copy the environ if they are not modifying it.

src/build/__init__.py
```diff
@@ -16,29 +17,16 @@
...
 
 import pep517.wrappers
 
+from . import env
+from ._helpers import ConfigSettingsType, PathType, check_dependency, default_runner, rewrap_runner_for_pep517_lib
 
-TOMLDecodeError: Type[Exception]
-toml_loads: Callable[[str], MutableMapping[str, Any]]
 
+if TYPE_CHECKING:
```
layday, line 28: Add to default coverage excludes.

src/build/env.py
```diff
@@ -1,312 +1,322 @@
...
+        """
+        Create the isolated environment.
+
+        This method should be idempotent.
+        """
         raise NotImplementedError
 
     @abc.abstractmethod
-    def install(self, requirements: Iterable[str]) -> None:
+    def install_packages(self, requirements: Iterable[str]) -> None:
```
layday, line 42: Make variadic for pip to be able to customise subprocess args or offer some other way for pip to do that?

layday: Removed `install_packages` from the interface.  pip can implement a custom installation method.

FFY00:
Sorry, I'll try to review this today!

For the lineno checks, let's get the function lineno and add the relative lines inside the function.

```python
>>> import build.env
>>> build.env.IsolatedEnvBuilder.log.__code__.co_firstlineno + 10
138
```

layday:
Done, thanks for the suggestion.

src/build/env.py
```diff
@@ -1,312 +1,293 @@
...
 class IsolatedEnv(metaclass=abc.ABCMeta):
-    """Abstract base of isolated build environments, as required by the build project."""
+    """ABC for isolated build environments."""
 
-    @property
     @abc.abstractmethod
-    def executable(self) -> str:
-        """The executable of the isolated build environment."""
-        raise NotImplementedError
+    def create(self, path: str) -> None:
```
FFY00, line 28: Not all implementation need to be tied to a path, this API is not very nice :confused:

FFY00 (changes requested):
I have some issues with the API redesign. I think we should get rid of `IsolatedEnv.create` and replace `IsolatedEnvManager` with a `_DefaultIsolatedEnv`-specific context manager, which will require us to make it public.

src/build/env.py
```diff
@@ -1,312 +1,293 @@
...
-    return executable, script_dir
+    @overload
+    def __init__(self, isolated_env: IsolatedEnvType) -> None:
+        ...
 
+    def __init__(self, isolated_env: Optional[IsolatedEnvType] = None) -> None:
+        """
+        :param isolated_env: The isolated environment
+        """
+        self._isolated_env = isolated_env
```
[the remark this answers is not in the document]
FFY00: We might need parameters in `__init__`.

src/build/env.py
```diff
@@ -1,312 +1,293 @@
...
-    @staticmethod
-    def log(message: str) -> None:
-        """
-        Prints message
+    def _run(self, cmd: Sequence[str]) -> None:
+        return default_runner(cmd, env=self.prepare_environ())
 
-        The default implementation uses the logging module but this function can be
-        overwritten by users to have a different implementation.
+    def create(self, path: str) -> None:
```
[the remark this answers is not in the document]
FFY00: We need a different creation step because we might need parameters in `__init__`.

docs/conf.py
```diff
@@ -10,19 +10,20 @@
...
 project = 'build'
 copyright = '2020, Filipe Laíns'
 author = 'Filipe Laíns'
 
 # The short X.Y version
-version = build.__version__
+version = importlib.metadata.version('build')
 # The full version, including alpha/beta/rc tags
-release = build.__version__
+release = version
```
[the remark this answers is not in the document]
layday: Importing build trips up `sphinx_autodoc_typehints`; `build` would have already been imported with `TYPE_CHECKING` being `False`.  I suppose we could remove it from `sys.modules` after reading the version or something of the sort but this seemed cleaner.

src/build/__main__.py
```diff
@@ -235,11 +238,17 @@ def build_package_via_sdist(
     with tarfile.open(sdist) as t:
         t.extractall(sdist_out)
         try:
-            builder = _ProjectBuilder(os.path.join(sdist_out, sdist_name[: -len('.tar.gz')]))
             if distributions:
-                builder.log(f'Building {_natural_language_list(distributions)} from sdist')
+                _ProjectBuilder.log(f'Building {_natural_language_list(distributions)} from sdist')
```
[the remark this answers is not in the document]
layday: Because we can't pre-initialise the builder for isolated builds.

src/build/_helpers.py
```diff
@@ -0,0 +1,117 @@
...
+        local_env = os.environ.copy() if env is None else dict(env)
+        if extra_environ:
+            local_env.update(extra_environ)
+
+        runner(cmd, cwd=cwd, env=local_env)
+
+    return inner
+
+
+def check_dependency(
```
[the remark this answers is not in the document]
layday: > I assume there's a reason you move this here, right? Do we have issues with circular imports?

Yes, `__init__` would import `env` and `env` import `check_dependency` and `PathType` from `__init__`.  This manifests as an actual import cycle when building the docs with `TYPE_CHECKING` being `True`.

src/build/env.py
```diff
@@ -1,312 +1,293 @@
...
+                    # of a project which does not use setuptools.
+                    '--use-pep517',
+                    '-r',
+                    req_file.name,
+                ]
+            )
         finally:
             os.unlink(req_file.name)
 
+    def prepare_environ(self) -> Dict[str, str]:
```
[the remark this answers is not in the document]
layday: > Same thing. Why do we need a `prepare_environ`? Why can't we do this in `__init__`?

`prepare_environ` is part of the interface, it's used to construct a builder from an isolated env.

src/build/env.py
```diff
@@ -1,312 +1,293 @@
...
 
-try:
-    import virtualenv
-except ModuleNotFoundError:
-    virtualenv = None
 
+_logger = logging.getLogger(__name__)
 
-_logger = logging.getLogger('build.env')
+IsolatedEnvType = TypeVar('IsolatedEnvType', bound='IsolatedEnv')  #: :class:`IsolatedEnv` type alias.
```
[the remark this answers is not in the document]
layday: I didn't want to change it to a protocol precisely to avoid the extra dependency but if you're ok with it I can do that.

src/build/env.py
```diff
@@ -1,312 +1,293 @@
...
 
-    if platform.system() == 'Darwin' and int(platform.mac_ver()[0].split('.')[0]) >= 11:
-        # macOS 11+ name scheme change requires 20.3. Intel macOS 11.0 can be told to report 10.16 for backwards
-        # compatibility; but that also fixes earlier versions of pip so this is only needed for 11+.
-        is_apple_silicon_python = platform.machine() != 'x86_64'
-        minimum_pip_version = '21.0.1' if is_apple_silicon_python else '20.3.0'
-    else:
-        # PEP-517 and manylinux1 was first implemented in 19.1
-        minimum_pip_version = '19.1.0'
+class IsolatedEnvManager(Generic[IsolatedEnvType]):
```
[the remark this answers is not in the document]
layday: Both are fairly generic; "builder" has a more precise meaning in `ProjectBuilder` and using the same term to mean two different things as part of the same API can be confusing.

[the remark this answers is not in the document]
layday: > And I am starting to think this would be better as a `classmethod` context manager in the implementation class.

I agree, but I was under the impression that we didn't want to expose the default isolated env because we consider using pip to be an 'implementation detail'.

src/build/env.py
layday: The `IsolatedEnv{Builder,Manager}` needs to pass the path to the isolated env somehow.

src/build/__main__.py
```diff
@@ -181,11 +186,11 @@ def _natural_language_list(elements: Sequence[str]) -> str:
 def build_package(
     srcdir: PathType,
     outdir: PathType,
-    distributions: Sequence[str],
+    distributions: Sequence['Distribution'],
     config_settings: Optional[ConfigSettingsType] = None,
     isolation: bool = True,
     skip_dependency_check: bool = False,
-) -> Sequence[str]:
+) -> List[str]:
```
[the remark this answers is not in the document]
layday: Just following the 'robustness principle' which is espoused by mypy/Python typing. Returning abstract types is fairly unusual.  Do you think list is too restrictive or that it might change in the future?

src/build/__init__.py
```diff
@@ -227,112 +185,79 @@ class ProjectBuilder:
...
 
+    _default_rewrapped_runner = staticmethod(rewrap_runner_for_pep517_lib(default_runner))
+
     def __init__(
         self,
         srcdir: PathType,
         python_executable: str = sys.executable,
-        scripts_dir: Optional[str] = None,
-        runner: RunnerType = pep517.wrappers.default_subprocess_runner,
+        runner: Optional[Union['RunnerType', Tuple['RunnerType', Optional[Mapping[str, str]]]]] = None,
```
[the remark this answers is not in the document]
layday: Because `rewrap_runner_for_pep517_lib` needs access to the environ to be able to update it with values from `extra_environ`.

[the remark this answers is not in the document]
layday: I agree that this is a bit unwieldy, do you have a better idea for how to preserve the environ?

src/build/__init__.py
```diff
@@ -187,57 +187,51 @@ def _parse_build_system_table(pyproject_toml: Mapping[str, Any]) -> Dict[str, An
...
-            PEP 517 hooks
+        :param srcdir: Project source directory
+        :param python_executable: Path of Python executable used to invoke PEP 517 hooks
+        :param isolated_env: An :class:``env.IsolatedEnv`` instance
         :param runner: Callback for executing PEP 517 hooks in a subprocess
         """
         self._srcdir = _parse_source_dir(srcdir)
-        self._python_executable = python_executable
+
+        if isolated_env is not None:
```
layday, line 216: Could use `**kwargs` and raise `TypeError` if `isolated_env` is passed along with `python_executable` but I thought it'd be overkill.

docs/conf.py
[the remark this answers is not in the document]
layday: Reverted since we're depending on `typing-extensions` now and have no `TYPE_CHECKING` imports as a result.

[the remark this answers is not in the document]
FFY00: Thanks. Before merging can you rebase and remove the commits?

src/build/__main__.py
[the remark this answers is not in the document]
FFY00: I believe it's better to return well scoped protocols than to return concrete implementations. If we need to change the container here, that would be an API breakage. IMO this is a good practice when designing APIs. There's no reason here why we would need to tie the return value to a list object.
Even in internal APIs, I think limiting the promises is generally good practice, as it prevents us from depending too much on concrete implementations and then having to refactor the code if we need to adjust something.

src/build/__main__.py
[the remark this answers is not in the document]
FFY00: Yes, sorry, I missed that.

src/build/__main__.py
```diff
@@ -235,11 +234,11 @@ def build_package_via_sdist(
     with tarfile.open(sdist) as t:
         t.extractall(sdist_out)
         try:
-            builder = _ProjectBuilder(os.path.join(sdist_out, sdist_name[: -len('.tar.gz')]))
             if distributions:
-                builder.log(f'Building {_natural_language_list(distributions)} from sdist')
+                _ProjectBuilder.log(f'Building {_natural_language_list(distributions)} from sdist')
             for distribution in distributions:
-                out = _build(isolation, builder, outdir, distribution, config_settings, skip_dependency_check)
+                srcdir = os.path.join(sdist_out, sdist_name[: -len('.tar.gz')])
```
FFY00, line 240: This can be outside the loop.

src/build/_helpers.py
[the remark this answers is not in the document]
FFY00: How were you importing? That should not happen when using absolute imports.

src/build/env.py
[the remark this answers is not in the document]
FFY00: I think it was just so that to get people to use `IsolatedEnvBuilder` instead IIRC.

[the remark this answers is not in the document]
FFY00: I am looking at your changes and I think you misunderstand me. The context manager should not be the instance, but instead a `classmethod` constructor. See the example I linked above. This should allow us to move the `create` and `prepare_environ` contents to `__init__`.

src/build/env.py
FFY00: Hence I think this should be tied to that specific class, which has that requirement. This can be done via a `classmethod` context manager.

src/build/__init__.py
```diff
@@ -223,116 +187,77 @@ def _parse_build_system_table(pyproject_toml: Mapping[str, Any]) -> Dict[str, An
...
+        ...
 
     def __init__(
         self,
         srcdir: PathType,
+        *,
         python_executable: str = sys.executable,
-        scripts_dir: Optional[str] = None,
-        runner: RunnerType = pep517.wrappers.default_subprocess_runner,
+        isolated_env: Optional[env.IsolatedEnv] = None,
```
FFY00, line 205: I don't like this. IMO it has a bad API separation. I think constructing from an isolated environment should be done via a custom constructor, `from_isolated_env`.

FFY00:
Thank you for the changes! I provided some more feedback. This is starting to look better, but still has some things I think should be changed :sweat_smile:

src/build/env.py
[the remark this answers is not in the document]
layday: That's a bit cleaner I suppose though I generally don't like having side effects in `__init__`.

docs/conf.py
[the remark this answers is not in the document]
layday: I will, I just find it easier to follow reviews without overwriting commits.

src/build/_helpers.py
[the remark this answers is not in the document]
layday: I don't think absolute imports is a good way to avoid import cycles. They're still there, you're just imposing artificial limitations on your package in a manner which is obscure and probably implementation specific.

src/build/env.py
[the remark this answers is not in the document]
layday: Yeah, I'd prefer to keep `create` separate - I really don't want to go about performing I/O in the constructor.  The trouble is that you'd then need to guard the object's methods prior to the creation of the env.  If you think about it, a proxy object like `IsolatedEnvManager` is the obvious solution to this problem.

src/build/__init__.py
layday: Is it bad because you consider the isolated env to be a 'higher-level' construct, compared to the other params? Just trying to understand.

gaborbernat:
@FFY00 I'm approving this, anything else left to do?

FFY00:
I'll try to review the new changes this weekend.

FFY00:
Just wanted to let you know that I haven't forgotten about this, I just have a bunch of things going on right now, and the holidays and end of year are making me even busier. I don't think there is any particular reason to rush this, as this is not critical to the affected parties, and they likely are on the same boat as me :sweat_smile:. I will try to get back to this in January.

layday:
To recap what has changed here in a major way:

* The env builder is no more. `IsolatedEnv` has been stripped down to the bare essentials - currently that's just the two properties, `environ` and `python_executable`.  This _will_ allow pip to slot in its own isolated env but it does not make our own `IsolatedEnv` amenable to reuse by pip. This is because our requirements are fundamentally different. pip will not want pip to be installed by `ensurepip` or `virtualenv`; they will want to expose the running pip to the isolated env. pip maintainers have also not expressed interested in supporting `virtualenv` in place of or in addition to the stdlib venv like we do.  Furthermore, pip supports a large number of installation options which our single-minded environment does not and that we would not be able to abstract as part of the isolated env protocol. 
* The runner has been modified to allow removing env vars from the environ - not just replacing them - as a precursor #406. This has rendered the callbacks from `pep517` incompatible with the project builder.

gaborbernat (approved):

layday:
Judging by https://github.com/pypa/pip/pull/10720, pip isn't interested in using build, so I'm closing this.

pradyunsg:
> Judging by [pypa/pip#10720](https://github.com/pypa/pip/pull/10720), pip isn't interested in using build, so I'm closing this.

Wait, what?

pradyunsg:
As I see it, moving to a venv-based build environment in pip is the first step in making it possible for pip to move to using build, so I'm a genuinely confused by the closing note on this issue.

layday:
I was given the impression that pip would be using venv directly in unison with pep517 (the library). The pip PR would have closed the pip issue where this PR was conceived and I had not heard from pip maintainers that they were interested in pursuing this further.  If there is interest, all the code is still here :)

pradyunsg:
In that case, do you want to reopen this? :)

FFY00:
This will be one of my priorities for the PyCon sprints.

pradyunsg:
What can I do to help move this forward? :)

layday:
Review my assumptions in https://github.com/pypa/build/pull/361#issuecomment-1007244397 and provide feedback on the isolated env interface - if there are any obvious methods or properties missing, if something doesn't feel right, etc.  We are gonna need to rebase but the isolated env interface hasn't changed fundamentally since this PR.

henryiii:
Personally stuck on the rebase. I started it, it's a pretty hefty one - the comments were not really very organized / squashed, so it was a lot of manual work. I think I was doing it in steps, or doing a manual merge. I then lost my spot and haven't restarted working on it. I've been traveling for a month and am pretty busy for the next week (workshop), but can probably work on this after that.

pradyunsg:
> * This is because our requirements are fundamentally different.

That is correct, and not a problem. pip will likely create an environment with `venv.EnvBuilder` and use its internal runner script to run itself within that environment to install packages.

> * The runner has been modified to allow removing env vars from the environ

That seems fine? I guess I don't understand the implications of this fully though I imagine that pip's not gonna handle anything about this differently than it does today.

layday:
> That seems fine? I guess I don't understand the implications of this fully though I imagine that pip's not gonna handle anything about this differently than it does today.

We wanted to be able to clear a bunch of `PYTHON*` env vars like `PYTHONHOME` and `PYTHONPATH`.  You can't do that with the stock `pep517` subprocess wrappers.  I don't remember the specifics anymore, if I'm honest.  But it's a breaking change to the public build API.

FFY00:
Yep, we need to create a subprocess with a controlled environment, where we remove *some* of these keys.

layday:
Superseded by #537.

PR #365, opened 2021-09-29
build: validate `build-system` table schema

### LINKED ISSUE #364: Validate build-system table schema
I just spend a bunch of time debugging an error due to me mistakenly setting `build-backend = ['something']` in the `build-system` table.

We should verify the schema (types) and provide a helpful error message if something is wrong.

https://www.python.org/dev/peps/pep-0518/#build-system-table
https://www.python.org/dev/peps/pep-0518/#json-schema

layday:
Closes #364.

### DIFF
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -94,10 +94,31 @@ class BuildBackendException(Exception):
         return f'Backend operation failed: {self.exception!r}'
 
 
+class BuildSystemTableValidationError(BuildException):
+    """
+    Exception raised when the ``[build-system]`` table in pyproject.toml is invalid.
+    """
+
+    def __str__(self) -> str:
+        return f'Failed to validate `build-system` in pyproject.toml: {self.args[0]}'
+
+
 class TypoWarning(Warning):
     """
     Warning raised when a potential typo is found
     """
+
+
+@contextlib.contextmanager
+def _working_directory(path: str) -> Iterator[None]:
+    current = os.getcwd()
+
+    os.chdir(path)
+
+    try:
+        yield
+    finally:
+        os.chdir(current)
 
 
 def _validate_source_directory(srcdir: str) -> None:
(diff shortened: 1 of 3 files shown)


gaborbernat (approved):

src/build/__init__.py
```diff
@@ -94,12 +94,33 @@ def __str__(self) -> str:
         return f'Backend operation failed: {self.exception!r}'
 
 
+class BuildSystemTableValidationError(BuildException):
+    """
+    Exception raised when the ``[build-system]`` table in pyproject.toml is invalid.
+    """
+
+    def __str__(self) -> str:
+        return f'Failed to validate `[build-system]`: {self.args[0]}'
```
FFY00, line 103:
```suggestion
        return f'Failed to validate `build-system` in pyproject.toml: {self.args[0]}'
```

tests/test_projectbuilder.py
```diff
@@ -570,4 +570,74 @@ def test_log(mocker, caplog, test_flit_path):
         ('INFO', 'something'),
     ]
     if sys.version_info >= (3, 8):  # stacklevel
-        assert [(record.lineno) for record in caplog.records] == [305, 305, 338, 368, 368, 562]
+        assert caplog.records[-1].lineno == 562
+
+
+def test_parse_build_system_table_type_validation():
```
FFY00, line 576: Can we parametrize this test instead?

henryiii (approved):
Once existing comments are addressed, LGTM.

tests/test_projectbuilder.py
layday: I would've done but it's kind of awkward when mixing test cases which raise.

layday: done

layday: The new parametrised tests fail unless I exclude `test_toml_instead_of_tomli`...

tests/test_projectbuilder.py
```diff
@@ -573,71 +579,51 @@ def test_log(mocker, caplog, test_flit_path):
...
+        (
             {'build-system': {}},
-        )
-
-    with pytest.raises(
-        build.BuildSystemTableValidationError,
-        match='`requires` must be an array of strings',
-    ):
-        build._parse_build_system_table(
+            pytest.raises(build.BuildSystemTableValidationError, match='`requires` is a required property'),
```
FFY00, line 587: We can just parameterize the match.

https://github.com/FFY00/python-pep621/blob/f374fa2fe8f53fc3b24ef9e08b00870fef8fce30/tests/test_standard_metadata.py#L479

tests/test_projectbuilder.py
```diff
@@ -570,4 +576,54 @@ def test_log(mocker, caplog, test_flit_path):
...
+        ),
+        (
+            {'build-system': {'requires': ['foo'], 'unknown-prop': False}},
+            pytest.raises(build.BuildSystemTableValidationError, match='Unknown properties: unknown-prop'),
+        ),
+    ],
+)
+def test_parse_build_system_table_type_validation(pyproject_toml, raises):
+    with raises:
+        build._parse_build_system_table(pyproject_toml)
```
FFY00, line 629: I would rather instantiate a `ProjectBuilder` here.

tests/test_projectbuilder.py
layday: Not all of them raise.

FFY00: Then split the test into one for the errors and one for the case it should pass.

layday: Do you not like the way it's structured now or why do you want me to do that?

FFY00: It's to remove duplication and complexity. If you split the test into two you can easily parameterize only the error string.

tests/test_projectbuilder.py
layday: I'd have to serialise them all as TOML, create a tmp dir for each and write them on disk - I don't think I can disentangle the disk I/O with the mocker.

tests/test_projectbuilder.py
layday: I hope https://github.com/pypa/build/pull/365/commits/1dd2d353382978d394919cb278b9941510142d66 is acceptable.

tests/test_projectbuilder.py
FFY00: I think you can, as you should be able to patch `builtins.open`, but I'll leave that up to you. This is just a nitpick.

layday: I think I might leave that for another day then 😛

tests/test_projectbuilder.py
FFY00: I still think it would be better to split the test into two, and validate the return of `_parse_build_system_table`, but I am not gonna block the PR because of that.

FFY00 (approved):

tests/test_projectbuilder.py
layday: Done in https://github.com/pypa/build/pull/365/commits/8bab23845d9f002977324a8b928dc1f97fd71c70.

PR #375, opened 2021-10-10
env: clear environment variables that interfere with Python

### LINKED ISSUE #373: local imports to setup.py alias with PYTHONPATH when run from ../pep517/in_process/_in_process.py
I am using a helper module "setup_helper.py" to add some consistency checking to my setup.py build process. The setup_helper.py is in the same folder as setup.py and is imported correctly when using "python setup.py". 

When using python -m build, I find that another package that I have that is on my PYTHONPATH with its own setup_helper.py in the sys.path is getting imported and called rather than the local one. I've tried adding back "." to the sys.path and haven't yet found a workaround. I've found that this is happening when sys.argv[0] or `__main__.__file__` point to this _in_process.py runner.
`
It's a subtle and unexpected error to have the import path handling change with this build system. I don't understand why this is happening to help it though.

FFY00:
Some environment variables, like PYTHONPATH, interfere with Python.
They should be cleared when constructing an isolated environment so that
they do not cause anything from the original environment to leak into
our new environment.

Fixes #373

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -13,7 +13,7 @@ import sysconfig
 import tempfile
 
 from types import TracebackType
-from typing import Callable, Iterable, List, Optional, Tuple, Type
+from typing import Callable, Dict, Iterable, List, Optional, Tuple, Type
 
 import packaging.requirements
 import packaging.version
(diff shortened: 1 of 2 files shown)


layday:
Is it ok to close this PR or will you be picking it up again? https://github.com/pypa/build/pull/406 would supersede it should some part of https://github.com/pypa/build/pull/361 be merged.

FFY00:
Let's close it.

PR #376, opened 2021-10-12
ci: add Python 3.11 dev

henryiii:
Followup to https://github.com/pypa/build/pull/374.


henryiii:
The bug on Python 3.11 looks like an issue with PyTest, perhaps? I've reduced this to just ubuntu for now.

PR #379, opened 2021-10-15
ci: use supported version of PyPy

henryiii:
For backward compatibility reasons, `pypy3` is stuck on an unsupported PyPy version. Using the modern spelling here instead.


FFY00:
I updated the branch rules, apparently it does not trigger the auto-merge :see_no_evil:

Issue #380, opened 2021-10-15
Consider turning off coverage for PyPy tests


layday:
Coverage (i.e. tracing) is supposed to slow down execution considerably on PyPy.  We could try turning it off for the PyPy tests to see how much of a difference it makes.

henryiii:
We can add concurrently groups and auto cancel if you'd like.

PR #386, opened 2021-10-21
Run slower PyPy/Windows/macOS first for faster builds

hugovk:
This would help towards the problems mentioned in https://github.com/pypa/build/issues/380.

Running the slower ones first gives them a head start, so when the quick ones have finished, we haven't wasted time waiting for the slow one to even begin.

It especially helps when there's a queue and not all build runners are initially available.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,9 +28,9 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - ubuntu
-          - macos
           - windows
+          - macos
+          - ubuntu
         py:
           - 'pypy-3.7'
           - '3.11-dev'


layday:
I'm still seeing the Windows runners go last, possibly something to do with runner availability within PyPA, judging by this message:

```
Found online and idle hosted runner in the current repository's organization account that matches the required labels: 'windows-latest'
Waiting for a Hosted runner in the 'organization' to pick this job...
Job was not picked up by the preferred runner, try requesting additional runners to run this job
Waiting for a self-hosted runner to pickup this job...
```

hugovk:
Not much we can do about that, and I'm also seeing it for first-in-list Ubuntu on `main`. I see that message for all those not running, and in fact second-in-list Mac runners are going first.

Anyway, happy to close this if there's no observable benefit in this instance.

layday:
Still appreciate that you opened a PR, thanks.

PR #388, opened 2021-10-21
tests: require latest py for Python 3.11

henryiii:
This should fix the errors we are seeing in the tests for Python 3.11 by using the latest master for py, which includes the needed fixes.


layday:
I'll close this one in favour of https://github.com/pypa/build/pull/390 which adds 3.10 and 3.11 to the tox matrix.

PR #390, opened 2021-10-21
Fix 3.11: Temporarily install py from GitHub pending release

hugovk:
The Python 3.11 build is failing due to a problem (https://github.com/pytest-dev/py/issues/273) in the py dependency of pytest.

This has been fixed but not yet released.

To turn the CI green, we can first install py from GitHub, so it contains the fix.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -3,7 +3,7 @@ envlist =
     fix
     type
     docs
-    {py39, py38, py37, py36, pypy3}{, -path, -sdist, -wheel, -min}
+    {py311, py310, py39, py38, py37, py36, pypy3}{, -path, -sdist, -wheel, -min}
 isolated_build = true
 skip_missing_interpreters = true
 minversion = 3.14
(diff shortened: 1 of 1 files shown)


henryiii:
I'll add my further fixes here.

henryiii:
Actually, I like my original syntax better, it actually worked, while this is missing the -path, etc versions. :)

henryiii:
Looked like it was going to pass, so pushed one more change to the extras requirements, to bring in inline with the constraints.txt. Needs to be auto squashed and then split into two commits, then it should be ready.

henryiii:
Resquashed into three commits, fixing the minimum setuptools for python 3.10, using the dev version of py on 3.11, and adding 3.10 and 3.11 to the default tox matrix.

henryiii:
Back to only PyPy on Windows being red. @FFY00 I think we can add Python 3.10 to the required checks after this fix for the "-min" tests.

hugovk:
@henryiii Apologies if I stepped on your work elsewhere, thanks for finishing things up here! This is now a much nicer way of doing it.

PR #391, opened 2021-10-21
ci: refactor tox targets into matrix for faster builds via more parallelism

hugovk:
This would help towards the problems mentioned in #380, as suggested at https://github.com/pypa/build/pull/386#issuecomment-948738999:

Another option to get it down from [~1h 13m](https://github.com/pypa/build/actions/runs/1367055009) is to factor out the five tox runs (tox, path, sdist, wheel, min) into the test matrix.

So instead of (say):

* 4m + 4m + 4m + 4m + 4m in sequence = 20m for a good case (Ubuntu/3.6), and
* 9m + 9m + 9m + 9m + 9m = 45m for a bad case (macOS/PyPy),

let those run in parallel. The main pre/post steps are only a few seconds, so not much overhead.

Here's an example which took [32m 37s](https://github.com/hugovk/build/actions/runs/1367752095) - https://github.com/hugovk/build/commit/36e2540f28cb28f3795058d8fa8038b69ce1adec.

Quick illustration!

* Before: two jobs before with five sequential tox runs each.
* After: ten jobs with one tox run each, shown here with a maximum four runners in parallel.

image

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,15 +32,22 @@ jobs:
           - macos
           - windows
         py:
+          - 'pypy-3.7'
+          - '3.11-dev'
           - '3.10'
           - '3.9'
           - '3.8'
           - '3.7'
           - '3.6'
-          - 'pypy-3.7'
-        include:
-          - os: ubuntu
-            py: '3.11-dev'
+        tox-target:
+          - 'tox'
+          - 'path'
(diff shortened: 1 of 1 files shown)


hugovk:
Rebased to include merged #390 for passing 3.11.

henryiii:
What happened to PyPy?

PR #398, opened 2021-10-30
Fix coverage data collection

### LINKED ISSUE #397: Coverage reports broken on Codecov since 10 Oct
Since #371, Codecov is only reporting test coverage, see https://app.codecov.io/gh/pypa/build/compare/371.

layday:
Closes #397.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -91,6 +91,7 @@ ignore_missing_imports = True
 strict = True
 
 [coverage:run]
+source = build, tests
 omit =
     setup.py
     *bin/pyproject-build


setup.cfg
```diff
@@ -91,6 +91,7 @@ ignore_missing_imports = True
 strict = True
 
 [coverage:run]
+source = build, tests
```
gaborbernat, line 94: Just a note that this does not work for subprocess calls that change the cwd... https://twitter.com/gjbernat/status/1452419280394461195 and we might have some of those around...

gaborbernat (approved):

setup.cfg
layday: Good to know, though looking at the Codecov report, I don't think we're affected by it.

PR #399, opened 2021-10-30
config: migrate tool config to `pyproject.toml`

layday:
On top of #398.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -53,7 +53,7 @@ docs =
 test =
     filelock>=3
     pytest>=6.2.4
-    pytest-cov>=2
+    pytest-cov>=2.12
     pytest-mock>=2
     pytest-rerunfailures>=9.1
     pytest-xdist>=1.34
@@ -71,42 +71,3 @@ virtualenv =
 [options.package_data]
 build =
     py.typed
-
-[tool:pytest]
-addopts = --strict-config --strict-markers
-xfail_strict = True
-junit_family = xunit2
-norecursedirs = tests/integration/*
-markers =
-    isolated
-filterwarnings = error
-
-[flake8]
-max-line-length = 127
-max-complexity = 10
-extend-ignore = E203
-
-[mypy]
-ignore_missing_imports = True
-strict = True
-
-[coverage:run]
-source = build, tests
-omit =
-    setup.py
-    *bin/pyproject-build
-    *bin\pyproject-build.exe
-
-[coverage:report]
-exclude_lines =
-    \#\s*pragma: no cover
-    ^\s*raise NotImplementedError\b
-
-[coverage:paths]
-source =
-    src
-    */site-packages
-    *\site-packages
-
-[coverage:html]
-show_contexts = true
--- a/tests/constraints.txt
+++ b/tests/constraints.txt
@@ -1,10 +1,9 @@
-importlib-metadata==0.22;python_version < "3.8"
+importlib-metadata==0.22
 packaging==19.0
 pep517==0.9.1
+setuptools==42.0.0; python_version < "3.10"
+setuptools==56.0.0; python_version >= "3.10"
 toml==0.10.0
 tomli==1.0.0
-typing==3.5.3.0;python_version < "3"
-virtualenv==20.0.35;python_version < "3"
-setuptools==42.0.0;python_version < "3.10"
-setuptools==56.0.0;python_version >= "3.10"
+virtualenv==20.0.35
 wheel==0.36.0
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,11 +1,56 @@
 [build-system]
-requires = ['setuptools >= 40.8.0', 'wheel']
-build-backend = 'setuptools.build_meta'
+requires = ["setuptools >= 40.8.0", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[tool.coverage.run]
+source = [
+  "build",
+  "tests",
+]
+
+[tool.coverage.report]
+exclude_lines = [
+  '\#\s*pragma: no cover',
+  '^\s*raise NotImplementedError\b',
+]
+
+[tool.coverage.paths]
+build = [
+  "src",
+  "*/site-packages",
+  '*\site-packages',
+]
+
+[tool.coverage.html]
+show_contexts = true
+
+[tool.pytest.ini_options]
+addopts = "--strict-config --strict-markers"
+xfail_strict = true
+junit_family = "xunit2"
+norecursedirs = "tests/integration/*"
+markers = [
+  "isolated",
+]
+filterwarnings = [
+  "error",
+]
+
+[tool.mypy]
+strict = true
+
+[[tool.mypy.overrides]]
+module = [
+  "colorama",   # Optional dependency
+  "pep517.*",   # Untyped
+  "virtualenv", # Optional dependency
+]
+ignore_missing_imports = true
 
 [tool.black]
 line-length = 127
 skip-string-normalization = true
-target-version = ['py39', 'py38', 'py37', 'py36']
+target-version = ["py39", "py38", "py37", "py36"]
 
 [tool.isort]
 profile = "black"
--- a/tox.ini
+++ b/tox.ini
@@ -32,7 +32,7 @@ extras =
 deps =
      py@git+https://github.com/pytest-dev/py@master;python_version>="3.11"
 commands =
-    pytest -rsx --cov --cov-config setup.cfg \
+    pytest -rsx --cov --cov-config pyproject.toml \
       --cov-report=html:{envdir}/htmlcov --cov-context=test \
       --cov-report=xml:{toxworkdir}/coverage.{envname}.xml \
       tests {posargs:-n auto}
@@ -105,3 +105,8 @@ commands =
     coverage html -d {toxworkdir}/htmlcov -i
     python -m diff_cover.diff_cover_tool --compare-branch {env:DIFF_AGAINST:origin/main} {toxworkdir}/coverage.xml
 depends = {py311, py310, py39, py38, py37, py36, pypy3}{, -path, -sdist, -wheel}
+
+[flake8]
+max-line-length = 127
+max-complexity = 10
+extend-ignore = E203


.flake8
```diff
@@ -0,0 +1,4 @@
+[flake8]
+max-line-length = 127
+max-complexity = 10
+extend-ignore = E203
```
gaborbernat, line 4: IMHO this should be in tox.ini or setup.cfg. Considering we want to move away from setuptools I'd vote for tox.ini.

gaborbernat (changes requested):

.flake8
layday: That doesn't sit right with me, in the same way that `setup.cfg` I feel should not have had other tools appropriate it for their configuration - I give more weight to having a logical hierarchy than the desire to cut down on the number of configuration files. The case is stronger against `tox.ini` because (a) the file unambiguously belongs to tox and (b) there's less precedent for it being used for third-party configuration. I don't want to encourage that usage, especially in a package that Python programmers might refer to for good configuration practices.

gaborbernat: > The case is stronger against `tox.ini` because (a) the file unambiguously belongs to tox and (b) there's less precedent for it being used for third-party configuration

Maybe the way you used it. But both pytest and flake8 support it out of the box. Guess we'll need a third reviewer to break the tie here.

layday: > If you put flake8 on its own then you should do the same with mypy and coverage.

I'm not following, why would I pull every other tool out of the project-wide, standardised configuration file because one tool doesn't support it?

gaborbernat: > standardised configuration file because

I think this is where you're mistaken. `pyproject.toml` is not standardised or mandated as a configuration file. Per PEP-518:

> The [tool] table is where any tool related to your Python project, not just build tools, can have users specify configuration data as long as they use a sub-table within [tool], e.g. the flit tool would store its configuration in [tool.flit].

Note the usage of the word `can`. It's optional supporting that file and tools are not mandated to do so. The only thing standardized in `pyproject.toml` is the build definitions per PEP-517/518/621. Everything else is just incidental.

layday: It's the closest thing we've got to "standardised" and the file the community is converging on for configuration.  I still don't see how it follows that I should pull everything else out of `pyproject.toml` on account of flake8.

gaborbernat: I think we just need to accept we don't have a `standardised` configuration file and go from that premise. Then if you're going to bundle up configuration files to minimize the number of configuration files lying around you might as well group flake8 with tox, as you already grouped mypy/coveragepy with the build configuration file. If minimizing the number of configuration files is not a goal, then we might as well have a dedicated one for each. The way I see it, it's either of those two a viable path ahead.

layday: My aim is to move the configuration out of `setup.cfg` to a place where it makes sense to have it, in a well-defined format with very few surprises, native support for nested structures, and with actual provision for tool configuration.  We don't need to be dealing in absolutes, that either minimisation must be maximised or it must be minimised.  I don't have anything more to add other that I haven't said already.  If `.flake8` is a deal breaker for you we can use `tox.ini`.  I'd rather do that than keep debating four lines of configuration, especially for a tool whose author has blocked me on GH.  Being reminded me of that put me in a foul mood.

gaborbernat: > My aim is to move the configuration out of `setup.cfg` to a place where it makes sense to have it

I agree with that aim, but if it makes sense to have mypy in `pyproject.toml` I think it's alright to have flake8 in `tox.ini`, per my reasoning above.

setup.cfg
```diff
@@ -53,7 +53,7 @@ docs =
 test =
     filelock>=3
     pytest>=6.2.4
-    pytest-cov>=2
+    pytest-cov>=2.12.0
```
gaborbernat, line 56: I think the `.0` is redundent.

tests/constraints.txt
```diff
@@ -1,10 +1,9 @@
-importlib-metadata==0.22;python_version < "3.8"
+importlib-metadata==0.22
```
gaborbernat, line 1: Hm, how come we still use this on newer Python versions?

tests/constraints.txt
```diff
@@ -1,10 +1,9 @@
...
 packaging==19.0
 pep517==0.9.1
+setuptools==42.0.0; python_version < "3.10"
+setuptools==56.0.0; python_version >= "3.10"
 toml==0.10.0
 tomli==1.0.0
```
gaborbernat, line 7: Do we still need both of these?

layday: Not sure honestly, do the tests run with toml?

tests/constraints.txt
layday: We don't but this is a constraint file so importlib-metadata won't be installed unless it's required by build.

tests/constraints.txt
layday: I guess it's a transitive dependency. If an older version of pep517 is installed then it'll require toml, not tomli, so we're testing that an old version of toml works with an old version of pep517?

.flake8
henryiii: Guess I'm not going to be at a computer any time soon, so here goes on my phone...

`pyproject.toml` is intended to be useable as a common config location for python packages. There are several advantages to having one config location; it reduces clutter in the main directory (very nice on GitHub where larger directories push the readme down the page), it makes it easier to see all configuration in one place, and reduces hunting around for configurations for each tool. It is not required for all tools to support it; it's just usable as such. There are cases where you might want config for a tool in a dedicated file, like if it is large. But most of the time, I think it's the right place. The only issue is a few tools don't support it (flake8 being the largest remaining one after mypy gained support).

tox.ini is a dedicated file for tox. I don't see why I'd expect to look there for configuration for a random tool (that is not tox or a tox plugin, that is). If we were to switch from tox to nox, for example, this would require moving the coverage config which is unrelated to the ?ox switch.

Same thing is technically true for setup.cfg - it's a setuptools specific config, but for a long time, setuptools was "special" and this was the best spot for random tools. I think it's best now to avoid config here unrelated to setuptools, but for backwards compat I've left PyTest config in here in a few Py2 compatible projects.

From a user perspective, you should look in pyproject.toml or a tool specific location for config. For historical reasons, setup.cfg too sometimes. Not in nox.ini for non-Nox tools. I literally missed the coverage config in build because of this a while back.

gaborbernat: > `pyproject.toml` is intended to be useable as a common config location for python packages.

I think this is the wish of some of the python community (including you), but we never agreed on this. There's no PEP/common discussion where (at least the PyPA project members) agreed on this.

> it makes it easier to see all configuration in one place

This is debatable. Once your configuration file does not fit into one screen I'd argue it makes it harder to reason about configuration, because when you see flake8 configs you also get side-tracked with mypy.

> reduces hunting around for configurations for each tool

From a UX POV individual files make it easier to locate the configuration. With everything in one file, you have to scroll a lot to discover the config for a given tool.

> tox.ini is a dedicated file for tox.

Not really, as tox maintainer/author I can say, tox never claimed exclusive rights to that file. pytest/flake8/devpi is happy to put its configuration into tox.ini file.

> If we were to switch from tox to nox, for example, this would require moving the coverage config which is unrelated to the ?ox switch.

We can cross that bridge when we get there. We optimize for the current state of the world, not some potential future of YAGNI.

> Same thing is technically true for setup.cfg - i

Just as tox.ini never requested exclusive rights to setup.cfg the same way setuptools did not either. It's fair play to put other stuff in there. And just as tox.ini or setup.cfg was never standardized as the one true location for tool configurations the same way pyproject.toml was never either (not explicitly via a PEP or pyproject.toml). So IMHO their on equivalent footing. There's a momentum for `pyproject.toml` but that's about its advtange. 

> From a user perspective, you should look in pyproject.toml or a tool specific location for config.

This is not true. As stated before we never agreed on this. Until we do IMHO this a week argument. No PEP makes this statement, and as such is the non-normative user wish.

gaborbernat (approved):

.flake8
henryiii: > The [tool] table is where any tool related to your Python project, not just build tools, **can**

This means that this file can be used by arbitrary tools, and it's the only generic configuration file with this property. There is also a clear namespace rule for arbitrary tools. tox.ini literally has the name Tox in it. setup.cfg sounds a little more general, at least. I know some tools allow using tox.ini, and if everybody did, it would be consistent at least. But why would a random user look in tox.ini for flake8 config, and why would it go there of a project doesn't use tox? I think that makes it an invalid choice for general configuration. Pyproject.toml is not tied to a tool, but rather a project.

I don't see what a community decision would add that's not already in the current PEP. What would a PEP for this say that's not already been said? "Please don't use other configuration files"? I don't think that would be a good idea - a tool with a complex setup might be better in a custom file. I wouldn't go so far as to say everything should be on a single page; I'd even say you can't have a rule about the total file length, but only the lengths of the configuration for each tool inside. I'd define "short" configurations under a page long as a good fit (users can scroll, and seeing the next configuration encourages scrolling). But I'd understand if someone said "a few lines" instead.

I would consider having many tiny files much harder to navigate, and it affects less advanced users who are just trying to get to the readme or understand the package, unlike the users who are opening the pyproject.toml file explicitly looking for configuration.

gaborbernat: > and it's the only generic configuration file with this property.

I'd argue against this. tox.ini or setup.cfg might not explicitly spell it out, but this is also true. If you want it to be spelt out I'm happy to add a note in the documentation that tox.ini can be used by other tools, but that was implied until now.

> why would a random user look in tox.ini for flake8

Because they read in the flake8 documentation and the tool documents it explicitly from where documentation is pulled from? The tools documentation is the only way to know where configurations are pulled from, as we don't have a mandated configuration file.

> why would it go there of a project doesn't use tox

If you go by that logic pyproject.toml is also a bad idea. Strictly speaking pyproject.toml is a configuration file for python build specification. Following your logic why would anyone store their tool configuration in there? What if they no longer want to build the project (e.g. now is an inline script).

> Pyproject.toml is not tied to a tool, but rather a project.

You're reading things into this. `pyproject.toml` is tied to the project build logic. It's possible to have a project without build, at which point your likely shouldn't use it?

> I would consider having many tiny files much harder to navigate,

This depends on how you interact with the code, and personal taste. Can make the argument both ways.

henryiii: > This is not true

Yes it is true today. If I want to see the PyTest config, I can look in pyproject.toml or PyTest.ini. I can also look in nox.ini or setup.cfg, but I'm arguing above those two locations are less general than pyproject.toml or less specific than pytest.ini. For black, it's pyproject.toml only (I'd be happy with black.toml as an alternative, because I believe adding a pyproject.toml changes still the build procedure in pip even though the PEP wording seems to suggest otherwise - at least I'm sure it used to). I don't know of any non-tox plugins that only support tox.ini as their config.

This is an argument about ideals, after all - obviously it works to put flake8 config in tox.ini, and to be complete you have to check all locations for config to find it, and having options is not bad. For a modern project, I'd likely only support pyproject.toml and `<package>.toml` (cibuildwheel supports pyproject.toml or you can manually pass `--config-file=somefile.toml` to look somewhere else). Older packages that have an ini config are welcome to support tox.ini and setup.cfg.

gaborbernat: > Yes it is true today. If I want to see the PyTest config, I can look in pyproject.toml or PyTest.ini. I can also look in nox.ini or setup.cfg, but I'm arguing above those two locations are less general than pyproject.toml or less specific than pytest.ini.

There's nothing mentioning here that those other sources are lesser: https://docs.pytest.org/en/6.2.x/customize.html So I think you're just projecting your wishes in here.

> For a modern project

This is such an overloaded trait. What is a modern project? What makes some practices modern? In my mind encouragement by a PEP. So while I can agree using pyproject.toml for stuff described in PEP-517/518/621 is preferable, where to place configuration files in not anywhere, so IMHO is not less modern to store something in setup.cfg/tox.ini than any other place.

> I don't see what a community decision would add that's not already in the current PEP.

Likely something around the line of `Python tools are encouraged to favour using pyproject.toml over other configuration files`. That would then follow my thought from previous paragraph and make it modern to put it there.

> Older packages that have an ini config are welcome to support tox.ini and setup.cfg.

This remains just your idealistic wish, without the PEP I just mentioned.

henryiii: > What is a modern project?

I thought the definition was implied. A modern project is one that only supports toml configuration. If you were writing a package today, toml is the language of choice. Why would you support ini and toml in a new project? While I dislike several decisions that TOML made (inline tables can't have newlines or ending commas, argh), at least it's a defined standard, unlike ini, and more powerful. Before pyproject.toml, ini was the language of choice, and "classic" projects are stuck supporting both.

> This remains just your idealistic wish, without the PEP I just mentioned.

Yes, it is. There is no mandate. And at best, I think this would only be a one line suggestion as you mention. That doesn't mean the community can't or shouldn't support pyproject.toml, and it is explicitly mentioned and allowed by the PEP, even for not build related tools.

> pyproject.toml is tied to the project build logic.

If you don't build a *package*, then it doesn't hurt to have this file. I admit this is a little weird and when I'm working on a project that is not a package, I would probably tend toward separate configuration files. (Just checked, I use setup.cfg, but that was also before mypy supported pyproject.toml) I wish the build backend was required for packages.

gaborbernat: > A modern project is one that only supports toml configuration.

lol, no; definitely no. I strongly object to this simplistic definition.

> If you were writing a package today, toml is the language of choice.

I'd argue against this too. For specifying your python build dependencies, yes. For other stuff, maybe, but often not.

henryiii: I like your (new) two line suggestion. I do think it should only be a suggestion; some tools might not target Python only and so use YAML configuration; I don't think that's to be discouraged; this is still for tools mostly targeting being used in python projects, like PyTest, mypy, etc. This should clearly be not applicable for a tool like pre-commit.

henryiii: Why would you support toml and ini for a new package today? Tools like mypy have made good use of TOML features not in ini (like arrays for overrides). Something like bump2version has weird caveats to configuration because ini doesn't support repeated headers. It's more work and possibly limits expressiveness to also support ini.

gaborbernat: > Why would you support toml and ini for a new package today?

ini is more flexible and allows more succinct configurations.

layday: I'm sure you are aware but INIs are untyped.  They afford about as much flexibility as strings do in TOML.  So if you want your users not knowing what the grammar that governs your pretend nested structures is, then yes, INIs are super flexible.

PR #401, opened 2021-11-02
Only import importlib_metadata when needed

jmroot:
Similar to #395, this helps with bootstrap issues with python < 3.8.

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -18,11 +18,6 @@ from typing import Callable, Collection, List, Optional, Tuple, Type
 import build
 
 
-if sys.version_info < (3, 8):
-    import importlib_metadata as metadata
-else:
-    from importlib import metadata
-
 try:
     import virtualenv
 except ModuleNotFoundError:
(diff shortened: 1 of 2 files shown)


layday:
The `importlib_metadata` import was removed in #361 which is awaiting feedback.

jmroot:
Ah, OK. Feel free to close then, unless the other PR is going to take a while (would be nice to get this in a release).

layday:
I don't have an estimate, sorry.

layday:
`test_pip_needs_upgrade_mac_os_11` needs fixing if you'd still like for this to be merged.

jmroot:
I'm afraid I don't know what the test that is now failing is even supposed to be testing, let alone how to fix it.

henryiii:
Ahh, that’s our self checking source test. It’s a pain. Basically you just have to adjust the line numbers if they change. We improved it a little, but it’s still very sensitive to any source changes. I can fix it later if that’s easiest.

jmroot:
@henryiii CI passed; is there any remaining reason not to approve this?

gaborbernat (approved):

henryiii (approved):
No, I just forgot to return. Gentle reminders always appreciated!

PR #407, opened 2021-11-14
util: return distribution from `project_wheel_metadata`

layday:
To provide access to all of the distribution folder's contents, e.g. entry points.

On top of #361.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -29,5 +29,5 @@ jobs:
       - name: Setup run environment
         run: tox -vv --notest -e docs
 
-      - name: Run check for type
+      - name: Run check for docs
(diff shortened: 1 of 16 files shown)


FFY00 (commented):
This will break some of my code, and it seems a bit weird having `project_wheel_metadata` return a distribution, but OTOH I don't know a better name. Acknowledging it's not very nice, I think this might be the best solution.

Can't merge right now as we are blocked by #361.

layday:
`prepare_metadata_for_build_wheel` also returns a distribution (a `dist` folder) so we're following the PEP's lead here.

layday:
Feel free to cherry-pick whatever makes sense.

PR #416, opened 2021-12-17
tests: use a relative lineno assert in test_log

jameshilliard:
This should prevent spurious test failures when changing unrelated tests.

### DIFF
--- a/tests/test_projectbuilder.py
+++ b/tests/test_projectbuilder.py
@@ -570,7 +570,9 @@ def test_log(mocker, caplog, package_test_flit):
         ('INFO', 'something'),
     ]
     if sys.version_info >= (3, 8):  # stacklevel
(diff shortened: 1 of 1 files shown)


layday:
We did something similar to this in https://github.com/pypa/build/pull/361 which needs to be resolved before we can accept any other PRs.

jameshilliard:
Think this approach or that one is preferred? I guess this is a negative relative computation while that is a positive one.

layday:
Well, this'll only work if the log message is emitted in the current frame's module - we have a similar test in `test_env` with line numbers from the `env` module. We're overtesting anyway.  What we want to know is that we're using line nos in 3.8+ but instead we're checking that the stdlib logging module's is emitting line nos correctly.

layday:
I'll close this one in favour of #361. Thanks for taking the time to open a PR :)

PR #420, opened 2021-12-24
env: prevent warning from being emitted on venv creation

### LINKED ISSUE #413: Failing Windows 3.9.9 tests
Can someone with Windows experience take a look at why the Windows 3.9.9 tests are failing? Some context might be found at https://github.com/actions/setup-python/issues/267.

layday:
After https://github.com/python/cpython/commit/6811fdaec825bd6ab64e358a4b480108f5634d2d
the venv module produces spurious warnings for venv paths which contain
DOS-encoded parts e.g. "USER\~1" in "C:\Users\USER~1".
`tempfile.gettempdir()` returns legacy paths like these for
user temp dirs.

MRE:

    python -c "import tempfile
    import venv

    venv.create(tempfile.mkdtemp())"

    Actual environment location may have moved due to redirects, links or junctions.
    Requested location: "C:\Users\RUNNER~1\AppData\Local\Temp\tmpfoobar\Scripts\python.exe"
    Actual location:    "C:\Users\runneradmin\AppData\Local\Temp\tmpfoobar\Scripts\python.exe"

Closes #413.

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -92,7 +92,9 @@ class IsolatedEnvBuilder:
 
         :return: The isolated build environment
         """
-        self._path = tempfile.mkdtemp(prefix='build-env-')
+        # ``realpath`` is used to prevent warning from being emitted that
+        # the venv location has moved on Windows.
+        self._path = os.path.realpath(tempfile.mkdtemp(prefix='build-env-'))
         try:
             # use virtualenv when available (as it's faster than venv)
             if _should_use_virtualenv():
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -92,8 +92,11 @@ class IsolatedEnvBuilder:
 
         :return: The isolated build environment
         """
-        # ``realpath`` is used to prevent warning from being emitted that
-        # the venv location has moved on Windows.
+        # Call ``realpath`` to prevent spurious warning from being emitted
+        # that the venv location has changed on Windows. The username is
+        # DOS-encoded in the output of tempfile - the location is the same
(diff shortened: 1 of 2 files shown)


layday:
@gaborbernat @henryiii could you take a look? This will unblock several other PRs.

gaborbernat (approved):

henryiii (approved):

uranusjr:
I feel this should be better fixed at the _warning_ site to use more robust path comparison logic, instead of calling realpath eagerly (and potentially prematurely).

src/build/env.py
```diff
@@ -92,7 +92,9 @@ def __enter__(self) -> IsolatedEnv:
 
         :return: The isolated build environment
         """
-        self._path = tempfile.mkdtemp(prefix='build-env-')
+        # ``realpath`` is used to prevent warning from being emitted that
+        # the venv location has moved on Windows.
```
uranusjr, line 96: This is to vague (and potentially misleading because the location does not actually move; the warning is due to a _misidentification_ of the location.

I’d say something like _Call realpath so venv does not incorrectly identify equivalent path forms as different locations on Windows_ with a reference to the b.p.o. issue entry.

layday:
Well, the warning site is CPython, so what do you suggest?

Sent with [ProtonMail](https://protonmail.com/) Secure Email.

src/build/env.py
layday: Expanded the text and linked to the issue.

layday: Thanks for the feedback :)

Issue #433, opened 2022-01-31
Custom default sysconfig INSTALL_SCHEME leads to superfulous "Virtual environment creation failed"


hroncok:
Hello,

With the proposal in https://bugs.python.org/issue43976 we are using an approach in Fedora, that changes the default install scheme to use `/usr/local` paths. Essentially, we set:

```python
if (not (hasattr(sys, 'real_prefix') or
    sys.prefix != sys.base_prefix) and
    'RPM_BUILD_ROOT' not in os.environ):
        _INSTALL_SCHEMES['posix_prefix'] = {
            'stdlib': '{installed_base}/{platlibdir}/python{py_version_short}',
            'platstdlib': '{platbase}/{platlibdir}/python{py_version_short}',
            'purelib': '{base}/local/lib/python{py_version_short}/site-packages',
            'platlib': '{platbase}/local/{platlibdir}/python{py_version_short}/site-packages',
            'include':
                '{installed_base}/include/python{py_version_short}{abiflags}',
            'platinclude':
                '{installed_platbase}/include/python{py_version_short}{abiflags}',
            'scripts': '{base}/local/bin',
            'data': '{base}/local',
        }
```

However, we have realized this breaks standard assumptions about virtual environments created by `virtualenv`.

In `virtualenv`, we have fixed this in https://github.com/pypa/virtualenv/issues/2208 by using a dedicated installation scheme called `"venv"`. The standard library `venv` module does not read install schemes at all, but we have opened https://bugs.python.org/issue45413 anyway.

In `build`, this causes problems that @vstinner reported to me on Fedora's Python IRC channel. tl;dr in here:

https://github.com/pypa/build/blob/96f9188ad181907fbd3e0efdf32dd3dc959d39c3/src/build/env.py#L306

The default install scheme is used. Hence on Fedora, `build` errors with:

```
RuntimeError: Virtual environment creation failed, executable /tmp/build-env-.../local/bin/python missing
```

That is correct, `/tmp/build-env-.../local/bin/python` is missing, but there is `/tmp/build-env-.../bin/python` which is completely OK.

We'd like `build` to use the `"venv"` scheme if it exists, as does `virtualenv` itself:

https://github.com/pypa/virtualenv/blob/bb2eeaca5a5c9818bf3b8b047f0e859c16905d26/src/virtualenv/discovery/py_info.py#L76-L80

[...]

PR #434, opened 2022-01-31
env._find_executable_and_scripts(): Prefer the venv installation scheme if it exists

hroncok:
Python distributors with custom default installation scheme can set a
scheme that can't be used to expand the paths in a venv.
This can happen if build itself is not installed in a venv.
The distributors are encouraged to set a "venv" scheme to be used for this.
See https://bugs.python.org/issue45413
and https://github.com/pypa/virtualenv/issues/2208

Since Python that ships with the macOS developer tools does not have the "venv"
scheme yet, we keep the special case below. Once it gains the "venv" scheme,
it will be preferred.

Fixes https://github.com/pypa/build/issues/433

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -293,22 +293,21 @@ def _find_executable_and_scripts(path: str) -> Tuple[str, str, str]:
     """
     config_vars = sysconfig.get_config_vars().copy()  # globally cached, copy before altering it
     config_vars['base'] = path
-    # Python distributors with custom default installation scheme can set a
-    # scheme that can't be used to expand the paths in a venv.
-    # This can happen if build itself is not installed in a venv.
-    # The distributors are encouraged to set a "venv" scheme to be used for this.
-    # See https://bugs.python.org/issue45413
-    # and https://github.com/pypa/virtualenv/issues/2208
-    #
-    # The Python that ships with the macOS developer tools varies the
-    # default scheme depending on whether the ``sys.prefix`` is part of a framework.
-    # But it does not (yet) set the "venv" scheme.
-    # If the Apple-custom "osx_framework_library" scheme is available but "venv"
-    # is not, we use "posix_prefix" instead which is venv-compatible there.
     scheme_names = sysconfig.get_scheme_names()
     if 'venv' in scheme_names:
+        # Python distributors with custom default installation scheme can set a
+        # scheme that can't be used to expand the paths in a venv.
+        # This can happen if build itself is not installed in a venv.
+        # The distributors are encouraged to set a "venv" scheme to be used for this.
+        # See https://bugs.python.org/issue45413
+        # and https://github.com/pypa/virtualenv/issues/2208
         paths = sysconfig.get_paths(scheme='venv', vars=config_vars)
     elif 'osx_framework_library' in scheme_names:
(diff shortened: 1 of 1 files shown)


src/build/env.py
```diff
@@ -293,14 +293,21 @@ def _find_executable_and_scripts(path: str) -> Tuple[str, str, str]:
...
-    # can happen if build itself is not installed in a venv.
-    # If the Apple-custom "osx_framework_library" scheme is available
-    # we enforce "posix_prefix", the venv scheme, for isolated envs.
-    if 'osx_framework_library' in sysconfig.get_scheme_names():
+    # But it does not (yet) set the "venv" scheme.
+    # If the Apple-custom "osx_framework_library" scheme is available but "venv"
+    # is not, we use "posix_prefix" instead which is venv-compatible there.
+    if 'venv' in sysconfig.get_scheme_names():
+        paths = sysconfig.get_paths(scheme='venv', vars=config_vars)
+    elif 'osx_framework_library' in sysconfig.get_scheme_names():
```
layday, line 310:
```suggestion
    scheme_names = sysconfig.get_scheme_names()
    if 'venv' in scheme_names:
        paths = sysconfig.get_paths(scheme='venv', vars=config_vars)
    elif 'osx_framework_library' in scheme_names:
```

src/build/env.py
```diff
@@ -293,14 +293,21 @@ def _find_executable_and_scripts(path: str) -> Tuple[str, str, str]:
     """
     config_vars = sysconfig.get_config_vars().copy()  # globally cached, copy before altering it
     config_vars['base'] = path
+    # Python distributors with custom default installation scheme can set a
+    # scheme that can't be used to expand the paths in a venv.
+    # This can happen if build itself is not installed in a venv.
```
layday, line 298: I don't think this is relevant or that something distributors should be doing, varying the scheme at runtime depending on the prefix.

hroncok: That is the reality. We need to do that. See mostly https://bugs.python.org/issue43976

FFY00: Unfortunately, this was made possible with the introduction of `sysconfig._get_preferred_schemes` in 3.10, so something like this is needed.

FFY00 (approved):
This change looks alright, and is indeed necessary to deal with distributors overriding the default scheme.

@hroncok would you be able to move each part of the comment to above its relevant `if`/`elif` clause? I think that makes the code easier to read, instead of having this big comment.

hroncok:
> @hroncok would you be able to move each part of the comment to above its relevant if/elif clause? I think that makes the code easier to read, instead of having this big comment.

Done.

hroncok:
In https://github.com/python/cpython/pull/31034 the name of the installation scheme has been changed. We can either land this now and adapt later or wait a bit until the actual way forward is finalized there.

hroncok:
https://github.com/python/cpython/pull/31034 was merged, hence this is no longer a draft

FFY00:
Thanks!

FFY00:
Oops, forgot to squash merge :confused:

hroncok:
> Oops, forgot to squash merge confused

:man_shrugging:

hroncok:
Also, https://github.com/pypa/build/pull/448

PR #436, opened 2022-02-04
ci: fix broken tests

henryiii:
WIP at the moment. These are the 11 Windows PyPy failures:

```
FAILED tests/test_env.py::test_isolated_env_log - subprocess.CalledProcessErr...
FAILED tests/test_main.py::test_build_package - subprocess.CalledProcessError...
FAILED tests/test_main.py::test_build_package_via_sdist - subprocess.CalledPr...
FAILED tests/test_main.py::test_build_package_via_sdist_cant_build - subproce...
FAILED tests/test_main.py::test_output[via-sdist-isolation] - SystemExit: 1
FAILED tests/test_main.py::test_output[wheel-direct-isolation] - SystemExit: 1
FAILED tests/test_util.py::test_wheel_metadata[True] - subprocess.CalledProce...
FAILED tests/test_util.py::test_wheel_metadata_isolation - subprocess.CalledP...
FAILED tests/test_util.py::test_with_get_requires - subprocess.CalledProcessE...
FAILED tests/test_main.py::test_output_env_subprocess_error[no-color] - asser...
FAILED tests/test_main.py::test_output_env_subprocess_error[color] - assert [...
```

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -29,7 +29,7 @@ setenv =
 extras =
     test
 commands =
-    pytest -rsx --cov --cov-config pyproject.toml \
+    pytest -ra --cov --cov-config pyproject.toml \
       --cov-report=html:{envdir}/htmlcov --cov-context=test \
       --cov-report=xml:{toxworkdir}/coverage.{envname}.xml \
       tests {posargs:-n auto}
@@ -58,7 +58,7 @@ description = check minimum versions required of all dependencies
 skip_install = true
 commands =
     pip install .[test] -c tests/constraints.txt
-    pytest -rsx tests {posargs:-n auto}
+    pytest -ra tests {posargs:-n auto}
 
 [testenv:docs]
 description = build documentations
--- a/tox.ini
+++ b/tox.ini
@@ -23,6 +23,7 @@ passenv =
 setenv =
     COVERAGE_FILE = {toxworkdir}/.coverage.{envname}
     TEST_STATUS_DIR = {envtmpdir}
+    PYPY3323BUG = 1
     path: TEST_MODE = path
     sdist: TEST_MODE = sdist
     wheel: TEST_MODE = wheel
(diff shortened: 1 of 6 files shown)


henryiii:
The main issue with most of the failures is `python -IM ensure-up --upgrade --default-pip` is failing with `no module named encodings`. Our error reporting still needs work, as that error is completely hidden.

henryiii:
I believe all problems are due to a single bug, and that bug is getting fixed in https://foss.heptapod.net/pypy/pypy/-/issues/3323 for PyPy 7.3.9, which is planned to be released in a couple of weeks or so. I'll try to test with with a nightly build in a few days if possible. For now, I'm thinking we can xfail strict these tests, and then remove the workarounds once the new PyPy hits.

layday (approved):

FFY00 (approved):
Thank you so much for looking into this!

> I'll try to test with with a nightly build in a few days if possible.

If it fixes it, we should update the code. We can do that later.

tox.ini
```diff
@@ -23,13 +23,14 @@ passenv =
...
     TEST_STATUS_DIR = {envtmpdir}
+    PYPY3323BUG = 1
     path: TEST_MODE = path
     sdist: TEST_MODE = sdist
     wheel: TEST_MODE = wheel
 extras =
     test
 commands =
-    pytest -rsx --cov --cov-config pyproject.toml \
+    pytest -ra --cov --cov-config pyproject.toml \
```
layday, line 33: What does`-a` do?

layday: Right, that's the value of `-r`.  I would've separated them with a space for clarity, otherwise it looks like you're chaining multiple flags.

henryiii:
> We can do that later.

We have to do that later, I'm quite sure this morning's fix is not in the nightlies yet. :D

PR #442, opened 2022-02-25
fix: better error printouts

henryiii:
This is the remainder of the changes madw while working on #436. The key
change is the reworking of the error messages; if the venv builder fails,
there's much nicer error message now. Stdout/stderr are no longer hidden, which
revealed the true cause of the failure[^1]; a `from None` was added to reduce
the useless part of the printout for the builder exception (which already
"remembers" the parent exception in the constructor).

The other changes are minor cleanups noticed along the way, and are organized
by commit, feel free to look at the commits to see them individually; combined
here because our CI is large and clogs other PyPA projects.

Typing:

- chore(types): clean up over typing
- chore: bump mypy version
- chore: use `sys.platform` over `os.name` (supported by mypy)

Modernization:
- env: use `subprocess.run` instead of `check_output`
- env: update old mention of Python 2

- env: better error handling for failures creating venv
- main: nicer printout of errors, including out/err
- env: from None added, reducing useless printout

[^1]: This does change the type of the exception given when venv creation
  fails, but this should be a rare and broken thing anyway, so having a new
  custom exception for it seems fine for a 0.8 release.


pyproject.toml
```diff
@@ -39,6 +39,9 @@ filterwarnings = [
 ]
 
 [tool.mypy]
+files = "src"
+python_version = "3.7"
```
henryiii, line 43: Oops, sorry.

```suggestion
python_version = "3.6"
```

src/build/__main__.py
```diff
@@ -71,7 +71,7 @@ def _setup_cli() -> None:
         colorama.init()  # fix colors on windows
 
 
-def _error(msg: str, code: int = 1) -> None:  # pragma: no cover
+def _error(msg: str, code: int = 1) -> NoReturn:  # pragma: no cover
```
henryiii, line 74: This allows the function to serve as a guard below.

src/build/env.py
```diff
@@ -260,7 +260,12 @@ def _create_isolated_env_venv(path: str) -> Tuple[str, str]:
 
     import packaging.version
 
-    venv.EnvBuilder(with_pip=True, symlinks=_fs_supports_symlink()).create(path)
+    simlinks = _fs_supports_symlink()
```
henryiii, line 263: Moved here to make --showlocals more useful in pytest.

pyproject.toml
```diff
@@ -39,6 +39,9 @@ filterwarnings = [
 ]
 
 [tool.mypy]
+files = "src"
+python_version = "3.6"
```
layday, line 43: Eh, I don't think that testing against 3.6 only is a particularly good idea and I'd prefer to type check for the Python version I'm developing against by default.

src/build/__main__.py
```diff
@@ -250,9 +251,7 @@ def main_parser() -> argparse.ArgumentParser:
     """
     Construct the main parser.
     """
-    # mypy does not recognize module.__path__
-    # https://github.com/python/mypy/issues/1422
-    paths: Iterable[str] = build.__path__  # type: ignore
+    paths = build.__path__
```
layday, line 254: We don't need the assignment here anymore, we only had it so we could replace the type of `__path__`.

pyproject.toml
henryiii: I think the best thing is to test for the lowest supported python version. I'd rather always avoid the version currently being used, because you normal tests might already catch that! What I like doing normally (and thought it was done here, but it's not) is to set the minimum in the pyproject.toml, then run the maximum manually also in ~~pre-commit~~ ~~nox~~ tox. That way you are protected from using features that were introduced after 3.6, as well as protected from features that were removed in 3.10 (less common).

henryiii: 95% of the time, a feature was added since the (minimum version here) and you doesn't notice it when developing or testing, but mypy can catch it because you are checking against the earliest supported versions.

There are cases were this simply doesn't work, though - like using numpy, which can't type check against 3.6 or even very well against 3.7, and in those cases, you just have to trust your test matrix to catch these sorts of bugs.

henryiii:
That hardcoded line number that might move is very annoying...

pyproject.toml
layday: The version in use is the only version you can reliably type check against because dependencies can (and do) differ. If we want to type check 3.6 we should install Python 3.6 in CI - no gimmicks.

Although, as I recall we limited ourselves to 3.6 because we couldn't get importlib-metadata to behave across all Python versions that we support.

henryiii: This only becomes a problem when something starts dropping the oldest version so the versions mismatch. You can type check quite well against 3.6 from 3.10 as long your dependencies (as installed) still support that version. Typeshed is properly guarded for different versions. This is exactly what I described for NumPy above - it dropped < 3.8, so now you can't typecheck with the latest NumPy and <3.8 as a target, and no NumPy supported <3.7 & types.

Type checking is not runtime, and it's[^1] developer only. If there's a problem, you have to look at it and see if it's real, runtime problem, or a false positive, etc. Practicality beats purity here; you are trying to maximize your ability to find problems. If something comes up (which is rare, _especially with minimal dependencies_, but it can happen), then we can adjust. But for most cases, the Python targeting works extremely well in mypy from other versions.

One difference - I always do type checking in pre-commit, with a (somewhat) locked and very limited environment. This gives more control than assuming some random environment can run type checking. If a dependency doesn't support 3.6, a limit could be included there. And I always specify the minimum here, even if I also test the maximum in CI sometimes.

I'm okay to move this back to CI only, but I think your chances of finding a real error are higher if you have the minimum set than if you don't. Yes, it's not "perfect", but it's rate of failure is lower than the rate of errors caught that would otherwise be missed[^2]. And I am not interested in installing Python 3.6 locally - in fact, on my AS machine, I _can't_, even though I can still type check for 3.6.

[^1]: minus external signatures, but that should not be affected here.
[^2]: I like newer Python features, so maybe my rate of mistakes here is high.

henryiii: Ahh, I see why newer versions were not also run, due to the type ignores. In that case, this really should always be set, because a newer version running the "newer" target will likely tell you to remove type ignores that still need to be there for CI.

src/build/__main__.py
henryiii: Thanks!

henryiii:
(You can click "force-pushed" above to see what changed.) I've added tests for broken venv, simplified the printout code a tiny bit, removed the unneeded line pointed out by @layday, and tried to fix the moving code issue in the tests.

pyproject.toml
layday: We can revisit this when we've got static typing to work with all versions that we support. I'm convinced that overriding the default Python version is the better choice for the time being.

henryiii:
Any reviews?

src/build/env.py
layday: Typo, should be symlinks.

layday (approved):

gaborbernat (approved):

PR #443, opened 2022-03-08
Use tomllib on Python 3.11

hauntsaninja:


src/build/__init__.py
```diff
@@ -40,13 +40,16 @@
...
 
-
 try:
-    from tomli import TOMLDecodeError
-    from tomli import loads as toml_loads
-except ModuleNotFoundError:  # pragma: no cover
-    from toml import TomlDecodeError as TOMLDecodeError  # type: ignore
-    from toml import loads as toml_loads  # type: ignore
+    from tomllib import TOMLDecodeError  # type: ignore
+    from tomllib import loads as toml_loads  # type: ignore
```
gaborbernat, line 45: Rather than this try-except let's use if/else and check based on the python version what we need to do. This is also mypy friendly.

gaborbernat (dismissed):

henryiii:
It was added to CPython 5 hours ago. :) https://github.com/python/cpython/tree/main/Lib/tomllib

Sadly just before the latest alpha, so we won’t be able to test this for now. Maybe we should wait to merge till the next dev release? (It will fail once fixed to use sys.version_info until the next dev release)

layday:
Yeah, let's just wait.

henryiii:
Waiting on https://github.com/actions/python-versions/pull/155 to merge and deploy.

henryiii:
I'm guessing we are now waiting for https://github.com/actions/python-versions/pull/157.

hauntsaninja:
Test looks broken on main as well, looking into it...

hauntsaninja:
https://github.com/pypa/build/pull/459 for the Python 3.11 failures

layday (approved):

layday:
@gaborbernat please unblock this :)

FFY00:
Thank you!

PR #459, opened 2022-04-16
Fix test_executable_missing_post_creation for Python 3.11, PyPy

hauntsaninja:
As a result of https://github.com/python/cpython/pull/31034,
venv now calls sysconfig.get_paths. This messes up the mock call
count.

The change to using Signature.bind isn't strictly necessary, but fixes
the initial symptom (since venv calls sysconfig.get_paths without
kwargs). This also means we can use sysconfig.get_paths however we want.

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -289,10 +289,6 @@ def _create_isolated_env_venv(path: str) -> Tuple[str, str]:
     return executable, script_dir
 
 
-# Alias is used for mocking in tests
-_sysconfig_get_paths = sysconfig.get_paths
-
-
 def _find_executable_and_scripts(path: str) -> Tuple[str, str, str]:
     """
     Detect the Python executable and script folder of a virtual environment.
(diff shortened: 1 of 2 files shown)


hauntsaninja:
cc @hroncok

henryiii (approved):
LGTM. The PyPy 3.7 Windows builds broke 4-11 days ago; guessing it might be the PyPy 7.3.9 release. Look like pytest-rerun-failures is breaking it. The other libraries I checked haven't had releases in that time frame.

src/build/env.py
```diff
@@ -289,6 +289,9 @@ def _create_isolated_env_venv(path: str) -> Tuple[str, str]:
     return executable, script_dir
 
 
+_sysconfig_get_paths = sysconfig.get_paths
```
hroncok, line 292: As somebody who reads this code, I believe this assignment would benefit from a comment.

layday:
I feel like this is a lot of unnecessary work just to test that a file which does not exist will produce an error. Can we mock `venv.EnvBuilder.create` instead?

layday:
I suppose we can also get rid of the skip decorator since the same is true for Python 3.11.

layday (approved):
Thanks!

FFY00:
Forgot to change the commit message again 😣

FFY00:
Wait, I thought I was the was the one merging this, perhaps you were too fast 😅

layday:
Yeah, I forgot too.

FFY00:
Yeah, no worries.

PR #460, opened 2022-04-18
tests: faster testing with less duplication

henryiii:
This replaces the SDist and wheel tests with tests of the file contents of the SDist and wheel. Every file is checked - over-including or forgetting a file will be detected. This also simplifies the path test and runs it on only one matrix element (all OS's) in order to reduce our testing load - currently we fill and block the entire PyPI's quota when running. If something is added that breaks build's ability to run from the path, it should break that matrix element.

With the newly reduced testing load, I've then enabled testing PyPy 3.8 and 3.9 as well. We are still broken on all Windows PyPy builds, but this should at least give us better coverage on linux and macOS, and hopefully I'll debug the Windows failure soon (it seems to be a clash between the latest release of PyPy and pytest-rerunfailures).


gaborbernat (commented):
You'll need to get the CI to pass but otherwise looks alright.

tox.ini
```diff
@@ -48,18 +42,42 @@ commands =
...
+    pytest>=6.2.4
+    pytest-cov>=2.12
+    pytest-mock>=2
+    pytest-rerunfailures>=9.1
+    pytest-xdist>=1.34
+    toml>=0.10.0
+    wheel>=0.36.0
+    setuptools>=42.0.0;python_version < "3.10"
+    setuptools>=56.0.0;python_version >= "3.10"
+commands =
```
layday, line 66: Can we install build and then uninstall it before running the tests instead of re-listing all of its dependencies here?

henryiii:
> You'll need to get the CI to pass but otherwise looks alright.

This is broken on main too - pypy + pytest-rerunfailures + Windows blows up. I'll need to play with this and probably report it to PyPy. Maybe we can turn off rerun failure on Windows PyPy...

I didn't realize you could link issues directly in the UI without mentioning them, interesting.

tox.ini
layday: Looking at the output from CI:

```
path run-test: commands[0] | pip uninstall -y build
Found existing installation: build 0.7.0
Not uninstalling build at d:\a\build\build\src, outside environment D:\a\build\build\.tox\path
Can't uninstall 'build'. No files were found to uninstall.
```

I wonder what the exit code from pip is and if it isn't 0 why it isn't being reinterpreted as an error.

layday:
```
~/c/p/build (main|✔) $ PYTHONPATH=src python -m pip uninstall -y build
Found existing installation: build 0.7.0
Not uninstalling build at /[...]/build/src, outside environment /[...]/build/venv
Can't uninstall 'build'. No files were found to uninstall.
~/c/p/build (main|✔) $ echo $status
0
```

Ugh.

gaborbernat:
> This is broken on main too - pypy + pytest-rerunfailures + Windows blows up.

Shall we mark it then xfail and report it upstream, when they fix it we can re-enable it 🤔

henryiii:
@mattip, any idea why this is INTERNALERROR'ing pytest on all versions of PyPy on Windows? It didn't used to before the last release as far as I can tell. I've been trying to set up to test this off and on for the last three weeks, but it's hard to recreate locally - at least a simple pytest + pypy + windows + pytest-rerunfailures loaded without the INTERNALERROR. (PyCon and a bunch of other things keep getting in the way, and I specifically have to be at home to access my Windows box).

mattip:
It seems on PyPy `report.longrepr` is a string, and not whatever it should be. Do we know anyone who understands pytest debugging?

mattip:
> This is broken on main too - pypy + pytest-rerunfailures + Windows blows up

My guess is that any failing test (maybe PyPy only?) triggers some bad path through the report parsing. I submitted #467 to fix the test failures, which were actually tests marked xfail that are passing.

tox.ini
layday: I think that only works for building packages (assuming you mean the pip flag) but we could call python with `-E`.

PR #462, opened 2022-04-19
chore: typing updates

henryiii:
Some minor cleanup, and updates to typing.

- chore: remove wheel
- chore: minor style issue with pytest config
- typing: update mypy typing, fix Iterable -> Collection

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -63,7 +63,7 @@ test =
     setuptools>=56.0.0;python_version >= "3.10"
 typing =
     importlib-metadata>=4.6.4
-    mypy==0.931
+    mypy==0.942
     typing-extensions>=3.7.4.3
 virtualenv =
     virtualenv>=20.0.35
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -13,7 +13,7 @@ import sysconfig
 import tempfile
 
 from types import TracebackType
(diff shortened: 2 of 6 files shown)


layday (approved):

src/build/env.py
```diff
@@ -183,7 +183,7 @@ def executable(self) -> str:
     def scripts_dir(self) -> str:
         return self._scripts_dir
 
-    def install(self, requirements: Iterable[str]) -> None:
+    def install(self, requirements: Collection[str]) -> None:
```
henryiii, line 186: FYI, this is checked just below (`if not requirements`), which is invalid on an `Iterable` - it just happens to be true for some common iterables like set, list, and tuple. Caught by `"truthy-bool"`. I've changed it to `Collection`.

setup.cfg
```diff
@@ -63,7 +63,7 @@ test =
     setuptools>=56.0.0;python_version >= "3.10"
 typing =
     importlib-metadata>=4.6.4
-    mypy==0.931
+    mypy==0.941
```
gaborbernat, line 66: Any reason why not use `0.942`?

henryiii: Oops, forgot! No.

```suggestion
    mypy==0.942
```

gaborbernat (approved):

PR #463, opened 2022-04-20
Handle Debian's posix_local scheme

stefanor:
The issue and solution is the same as OSX.
The venv scheme wasn't standardised yet, so it's not using it yet.

Fixes: https://bugs.debian.org/1009916

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -307,6 +307,13 @@ def _find_executable_and_scripts(path: str) -> Tuple[str, str, str]:
         # See https://bugs.python.org/issue45413
         # and https://github.com/pypa/virtualenv/issues/2208
         paths = sysconfig.get_paths(scheme='venv', vars=config_vars)
+    elif 'posix_local' in scheme_names:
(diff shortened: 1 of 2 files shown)


henryiii:
Could probably duplicate this test and modify to cover this? https://github.com/pypa/build/blob/0b40b40625d475aecf5daeb8330fe301d5fd5e56/tests/test_env.py#L50-L56

stefanor:
Yeah, can do. I saw that test.

It won't be deep enough to give you a failure, unless run directly on the system python on a Debian system, but better than nothing.

stefanor:
> It won't be deep enough to give you a failure, unless run directly on the system python on a Debian system, but better than nothing.

Actually, I think this (otherwise noop) mock can ensure a failure, there.

stefanor:
Coverage shows as decreasing, but if the python knows about `posix_local`, it'll run the test and increase coverage.

gaborbernat (approved):

FFY00 (approved):

PR #469, opened 2022-05-12
chore: update changelog, minor touchups

henryiii:
A few fixes and cleanups.

- docs: update changelog
- chore: minor cleanups from #361
- chore: minor style check additions
- ci: update actions

### DIFF
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -3,6 +3,7 @@
 """
 build - A simple, correct PEP 517 build frontend
 """
+
 __version__ = '0.7.0'
 
 import contextlib
@@ -24,6 +25,7 @@ from typing import (
     Callable,
     Dict,
     Iterator,
+    List,
     Mapping,
     MutableMapping,
     Optional,
@@ -71,18 +73,18 @@ _DEFAULT_BACKEND = {
 }
 
 
-_logger = logging.getLogger('build')
+_logger = logging.getLogger(__name__)
 
 
 class BuildException(Exception):
     """
-    Exception raised by ProjectBuilder
+    Exception raised by :class:`ProjectBuilder`
     """
 
 
 class BuildBackendException(Exception):
     """
-    Exception raised when the backend fails
+    Exception raised when a backend operation fails
     """
 
     def __init__(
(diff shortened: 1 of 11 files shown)


src/build/__main__.py
```diff
@@ -78,8 +78,8 @@ def _error(msg: str, code: int = 1) -> NoReturn:  # pragma: no cover
     :param msg: Error message
     :param code: Error code
     """
-    print('{red}ERROR{reset} {}'.format(msg, **_STYLES))
-    exit(code)
+    print('{red}ERROR{reset} {}'.format(msg, **_STYLES), file=sys.stderr)
+    raise SystemExit(code)
```
henryiii, line 82: `exit()` is not part of the language, it's injected by `site.py`. Better to use `sys.exit()`, or even better, just `raise SystemExit` directly.

src/build/__init__.py
```diff
@@ -514,4 +517,8 @@ def log(message: str) -> None:
     'RunnerType',
     'TypoWarning',
     'check_dependency',
-)
+]
+
+
+def __dir__() -> List[str]:
```
henryiii, line 523: Makes tab completion prettier on Python 3.7+.

src/build/__main__.py
FFY00: Yes, that is a bug.

layday (approved):

layday:
I assume the tests are failing because errors are printed to stdout now.

src/build/__main__.py
```diff
@@ -78,8 +78,8 @@ def _error(msg: str, code: int = 1) -> NoReturn:  # pragma: no cover
     :param msg: Error message
     :param code: Error code
     """
-    print('{red}ERROR{reset} {}'.format(msg, **_STYLES))
-    exit(code)
+    print('{red}ERROR{reset} {}'.format(msg, **_STYLES), file=sys.stderr)
```
henryiii, line 81:
```suggestion
    print('{red}ERROR{reset} {}'.format(msg, **_STYLES))
```

We can change it out for stderr later.

henryiii:
Added a test for `dir(build)`, so this is ready. I'm ready for 0.8 after this. :)

PR #471, opened 2022-05-12
infra: switch to flit

layday:


henryiii:
Let's get 0.8.0 out first, then I think this is should go in. This is also part of why I added that per-file SDist/wheel test. ;)

layday:
Removing all the `egg-info` entries from the list was oddly satisfying.

tests/test_self_packaging.py
```diff
@@ -18,17 +18,6 @@
     'PKG-INFO',
     'README.md',
     'pyproject.toml',
-    'setup.cfg',
-    'setup.py',
-    'src',
-    'src/build',
```
layday, line 24: flit doesn't create folder entries in the zip.

henryiii:
Why are folders no longer there?

tests/test_self_packaging.py
henryiii: Interesting, okay. Not complaining. :D

pyproject.toml
```diff
@@ -1,6 +1,75 @@
...
+build-backend = "flit_core.buildapi"
+
+[project]
+name = "build"
+version = "0.7.0"
+description = "A simple, correct PEP 517 build frontend"
+readme = "README.md"
+requires-python = ">=3.6"
+license.file = "LICENSE"
+authors = [
```
gaborbernat, line 12: We should fill out the maintainer's list too, no?

layday: Added - please let me know if you'd like a different name or email being used.

gaborbernat: Looks good, ming fixing the merge conflicts?

gaborbernat: Also I think Filipe is still maintainer too :+1:

layday: > Looks good, ming fixing the merge conflicts?

Will do after 0.8.0 is out, otherwise I'd have to do it again for the changelog.

> Also I think Filipe is still maintainer too 👍

If it's ok with Filipe I'd just rather list everyone under authors. Everyone's got different ideas about what the difference between author and maintainer is and CM advises that maintainer should only be used if the original author is no longer active.

nanonyme:
Any more blockers? This and packaging are the final blockers on getting setuptools out of reverse dependencies of pypa/build and pypa/installer so everything can be built with pep517 backends.

henryiii:
Let's get #527 in, a 0.8.1 out, and then go for this.

henryiii (approved):

gaborbernat (approved):

FFY00 (approved):
Thanks :tada:

FFY00:
I think I am gonna make a new release with this PR if everyone is alright with that.

nanonyme:
I don't think packaging has made release with flit yet either.

nanonyme:
Packaging 22.0 has now been released with flit. This is final item in bootstrap chain.

PR #472, opened 2022-05-13
Changed ambigous date format to ISO

venthur:
The changelog used am ambiguous date format (e.g. `dd-mm-yyyy`) which is understood only in a very [few countries](https://en.wikipedia.org/wiki/Date_format_by_country#Listing) in the world. The vast majority of software projects have standardized on the non-ambigious [ISO format](https://en.wikipedia.org/wiki/ISO_8601) (e.g. `yyyy-mm-dd`) which is not only used in changelogs but also for most other situations where dates or date times are needed. Please consider accepting this change as it will make reading the change log less confusing.

https://keepachangelog.com/en/1.0.0/

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -31,7 +31,7 @@ Unreleased
 .. _#393: https://github.com/pypa/build/pull/393
 
 
-0.7.0 (16-09-2021)
+0.7.0 (2021-09-16)
 ==================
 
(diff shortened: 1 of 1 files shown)


layday (approved):

gaborbernat (approved):

layday:
Thanks :)

henryiii:
Sweet, I had just noticed this and it’s fixed already. :)

PR #474, opened 2022-05-18
style: add prettier

henryiii:
This adds prettier to our pre-commit, which styles yaml files (and potentially markdown & html if we have any eventually, even including yaml code blocks in markdown).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,14 +4,14 @@ on:
     branches:
       - main
     paths-ignore:
-      - 'docs/**'
-      - '*.md'
+      - "docs/**"
+      - "*.md"
   pull_request:
     branches:
       - main
     paths-ignore:
-      - 'docs/**'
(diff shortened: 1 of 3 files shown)


layday (approved):

.github/workflows/test.yml
```diff
@@ -4,14 +4,14 @@ on:
     branches:
       - main
     paths-ignore:
-      - 'docs/**'
-      - '*.md'
+      - "docs/**"
+      - "*.md"
```
gaborbernat, line 8: I think we need to remove this as prettier formats the markdown document, no?

henryiii: But pre-commit runs separately (on pre-commit.ci), it's not part of this job?

henryiii: (Same issue would be true with end-of-file checks, etc)

gaborbernat (approved):

PR #489, opened 2022-06-24
Add dependabot

gaborbernat:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,8 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+    ignore:
+      # Official actions have moving tags like v1
+      - dependency-name: "actions/*"
+        update-types:
+          ["version-update:semver-minor", "version-update:semver-patch"]
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,8 +4,3 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
-    ignore:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
```
henryiii, line 6:
```suggestion
      interval: "daily"
    ignore:
      # Official actions have moving tags like v1
      - dependency-name: "actions/*"
        update-types:
          ["version-update:semver-minor", "version-update:semver-patch"]
```

gaborbernat: Would this not stop updates from v1 to v2 though?

henryiii: No, since `version-update:semver-major` is not excluded. I'm not sure how it keeps `v1->v2` instead of `v1->v2.0.0`, though, but it does.

I also am not completely sure this is still required, the logic may have become smarter about respecting the existing version format. Originally it just tried to update to the latest "released" tag regardless of the format. I'm noticing that in https://github.com/scikit-build/cmake-python-distributions/network/updates/402609618 it doesn't seem to be making this mistake anymore. @mayeut, you made that change - is it no longer required to specify skipping minor/patch updates to get dependabot to keep "moving tag" releases?

henryiii: Ahh, I think this was fixed in https://github.com/dependabot/dependabot-core/pull/4953 - released in `v0.180.4, 6 April 2022`. So it's no longer required to add this at all, what you had was best. I'm almost two months behind the times!

.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
...
+    ignore:
+      # Official actions have moving tags like v1
+      - dependency-name: "actions/*"
+        update-types:
+          ["version-update:semver-minor", "version-update:semver-patch"]
```
henryiii, line 11:
```suggestion
```

Apologies! This was true before April.

PR #494, opened 2022-06-30
main: disable `colorama` on Linux and flush output

### LINKED ISSUE #493: `colorama` disables colors in GitHub Actions
GitHub Actions provide no TTY, so `colorama` import wraps stdout and strips all colors https://github.com/tartley/colorama/issues/209#issuecomment-465119302 so even forcing colors doesn't work.

```
2022-06-30T17:00:45.4633499Z Successfully installed build-0.8.0 pep517-0.12.0 tomli-2.0.1
2022-06-30T17:00:46.1506025Z ##[group]Run python -m build
2022-06-30T17:00:46.1506287Z [36;1mpython -m build[0m
2022-06-30T17:00:46.1506499Z [36;1mgit status --ignored[0m
2022-06-30T17:00:46.1559508Z shell: /usr/bin/bash -e {0}
2022-06-30T17:00:46.1559700Z env:
2022-06-30T17:00:46.1559871Z   FORCE_COLOR: 1
2022-06-30T17:00:46.1560056Z ##[endgroup]
2022-06-30T17:00:46.6361795Z * Creating venv isolated environment...
2022-06-30T17:00:49.6232262Z * Installing packages in isolated environment... (setuptools >= 40.8.0, wheel)
2022-06-30T17:00:52.0028343Z * Getting dependencies for sdist...
```

Took me a while to debug this. Added some debug statements.

```diff
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -49,6 +49,7 @@ def _init_colors() -> Dict[str, str]:
 _STYLES = _init_colors()
 print(_STYLES)
 
+print('{red}ERROR{reset} {}'.format("msg", **_STYLES))
 
 def _showwarning(
     message: Union[Warning, str],
@@ -65,6 +64,7 @@ def _showwarning(
 
 
 def _setup_cli() -> None:
+    print('{red}ERROR{reset} {}'.format("precolorama", **_STYLES))
     warnings.showwarning = _showwarning
 
     try:
@@ -73,6 +73,7 @@ def _setup_cli() -> None:
         pass
     else:
         colorama.init()  # fix colors on windows
+    print('{red}ERROR{reset} {}'.format("postcolorama", **_STYLES))
 
 
 def _error(msg: str, code: int = 1) -> NoReturn:  # pragma: no cover
```

And got this.

image

The solution is to enable `colorama` only on Windows.

abitrolly:
Fixes #493

### DIFF
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -47,7 +47,7 @@ def _init_colors() -> Dict[str, str]:
 _STYLES = _init_colors()
 
 
-def cprint(msg):
+def _cprint(msg):
     print(msg.format(**_STYLES), flush=True)
 
 
@@ -59,7 +59,7 @@ def _showwarning(
     file: Optional[TextIO] = None,
     line: Optional[str] = None,
 ) -> None:  # pragma: no cover
-    cprint('{yellow}WARNING{reset} ' + message)
+    _cprint('{yellow}WARNING{reset} ' + message)
 
 
 def _setup_cli() -> None:
@@ -79,20 +79,20 @@ def _error(msg: str, code: int = 1) -> NoReturn:  # pragma: no cover
     :param msg: Error message
     :param code: Error code
     """
-    cprint('{red}ERROR{reset} ' + msg)
+    _cprint('{red}ERROR{reset} ' + msg)
     raise SystemExit(code)
 
 
 class _ProjectBuilder(ProjectBuilder):
     @staticmethod
     def log(message: str) -> None:
-        cprint('{bold}* ' + message + '{reset}')
+        _cprint('{bold}* ' + message + '{reset}')
 
 
 class _IsolatedEnvBuilder(IsolatedEnvBuilder):
     @staticmethod
     def log(message: str) -> None:
-        cprint('{bold}* ' + message + '{reset}')
+        _cprint('{bold}* ' + message + '{reset}')
 
 
 def _format_dep_chain(dep_chain: Sequence[str]) -> str:
@@ -123,7 +123,7 @@ def _build_in_current_env(
         missing = builder.check_dependencies(distribution)
         if missing:
             dependencies = ''.join('\n\t' + dep for deps in missing for dep in (deps[0], _format_dep_chain(deps[1:])) if dep)
-            cprint('')
+            _cprint('')
             _error(f'Missing dependencies:{dependencies}')
 
     return builder.build(distribution, outdir, config_settings or {})
@@ -151,7 +151,7 @@ def _handle_build_error() -> Iterator[None]:
         _error(str(e))
     except BuildBackendException as e:
         if isinstance(e.exception, subprocess.CalledProcessError):
-            cprint('')
+            _cprint('')
             _error(str(e))
 
         if e.exc_info:
@@ -164,7 +164,7 @@ def _handle_build_error() -> Iterator[None]:
             tb = ''.join(tb_lines)
         else:
             tb = traceback.format_exc(-1)
-        cprint('\n{dim}' + tb.strip('\n') + '{reset}\n')
+        _cprint('\n{dim}' + tb.strip('\n') + '{reset}\n')
         _error(str(e))
 
 
@@ -374,10 +374,10 @@ def main(cli_args: Sequence[str], prog: Optional[str] = None) -> None:  # noqa: 
             artifact_list = _natural_language_list(
                 ['{underline}{}{reset}{bold}{green}'.format(artifact, **_STYLES) for artifact in built]
             )
-            cprint('{bold}{green}Successfully built ' + artifact_list + '{reset}')
+            _cprint('{bold}{green}Successfully built ' + artifact_list + '{reset}')
     except Exception as e:  # pragma: no cover
         tb = traceback.format_exc().strip('\n')
-        cprint('\n{dim}' + tb + '{reset}\n')
+        _cprint('\n{dim}' + tb + '{reset}\n')
         _error(str(e))
 
 
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -4,6 +4,7 @@
 import argparse
 import contextlib
 import os
+import platform
 import shutil
 import subprocess
 import sys
@@ -60,12 +61,14 @@ def _showwarning(
 def _setup_cli() -> None:
     warnings.showwarning = _showwarning
 
-    try:
-        import colorama
-    except ModuleNotFoundError:
-        pass
-    else:
-        colorama.init()  # fix colors on windows
+    if platform.system() == 'Windows':
+        # try to enable colors
+        try:
+            import colorama
+
+            colorama.init()
+        except ModuleNotFoundError:
+            pass
 
 
 def _error(msg: str, code: int = 1) -> NoReturn:  # pragma: no cover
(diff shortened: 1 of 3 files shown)


abitrolly:
pre-commit.ci autofix

src/build/__main__.py
```diff
@@ -60,12 +61,14 @@ def _showwarning(
...
-    else:
-        colorama.init()  # fix colors on windows
+    if platform.system() == 'Windows':
+        # try to enable colors
+        try:
+            import colorama
+
+            colorama.init()
+        except ModuleNotFoundError:
+            pass
```
henryiii, line 71: This is technically a weakness in our tests - we never verify that colorama can actually be missing on Windows.

abitrolly: It can not. because it is a required dependency.

https://github.com/pypa/build/blob/4edd5effe9912eee825ba9532255166dadc08a25/setup.cfg#L28-L33

abitrolly:
Got another problem. `colored` messages are piling at the end of the output.

image

Need to add more commits to turn off output buffering.

src/build/__main__.py
abitrolly: Removed exception check,

src/build/__main__.py
```diff
@@ -63,12 +63,9 @@ def _setup_cli() -> None:
...
     if platform.system() == 'Windows':
         # try to enable colors
-        try:
-            import colorama
+        import colorama
 
-            colorama.init()
-        except ModuleNotFoundError:
-            pass
+        colorama.init()
```
layday, line 72: I believe this was done on purpose to allow boostrapping build without colorama if so desired.

src/build/__main__.py
henryiii: It’s this way for bootstrapping. Please restore it. We can check for it by uninstalling it after installing build, or by checking in the “non installed” check. One of us can look into adding that, doesn’t mean you had to.

src/build/__main__.py
abitrolly: Do you need it? Or who then?

abitrolly:
@henryiii without output buffering everything works as expected. But I need to properly test formatting.

src/build/__main__.py
layday: https://github.com/pypa/build/pull/300#issuecomment-846318593

abitrolly: @FFY00 is anybody using `build` without `colorama` on Windows? I guess that needs a separate package with `colorama` commented out from dependencies. I removed import check, because otherwise it needs to be covered by tests.

src/build/__main__.py
```diff
@@ -46,6 +47,10 @@ def _init_colors() -> Dict[str, str]:
 _STYLES = _init_colors()
 
 
+def cprint(msg):
```
FFY00, line 50: Can you make this private?

src/build/__main__.py
abitrolly: But they won't be able to use it without installing dependencies. So they need to checkout the source and make edits to `setup.cfg`. In the same way they could just comment `colorama`.

Maybe it is better to use `NO_COLOR` than rebuilding the package?

https://github.com/pypa/build/blob/4edd5effe9912eee825ba9532255166dadc08a25/src/build/__main__.py#L37

It looks like a better place for `colorama` import.

src/build/__main__.py
abitrolly: Made it private.

src/build/__main__.py
FFY00: They will, you can install the package without it's dependencies. The point is that it won't blow up at runtime if the dependency is missing.

The main use-case for this is bootstraping a Python environment, where `build` is one of the packages required to do so. See https://github.com/FFY00/python-bootstrap for an example.

> It looks like a better place for `colorama` import.

Not really :rofl: 

`_setup_cli` is meant to contain all code that changes external state, which you only want to happen when you actually run the CLI. This should be better documented, sorry.

src/build/__main__.py
FFY00: Thanks!

src/build/__main__.py
abitrolly: Ok. I get back that import check. Coverage will fail.

FFY00: No worries, Codecov is a bit annoying, even though the job will fail, it is not required for merging.

abitrolly: I see. Got the import catch back.

src/build/__main__.py
```diff
@@ -47,7 +47,7 @@ def _init_colors() -> Dict[str, str]:
 _STYLES = _init_colors()
 
 
-def _cprint(fmt='', msg=''):
+def _cprint(fmt: Optional[str] = '', msg: Optional[str] = ''):
```
layday, line 50: This isn't optional, it would type error if None.

abitrolly: Just copy/pasted from below. After reading https://docs.python.org/3/library/typing.html#typing.Optional about how to annotate default params I guess it is fixed.

abitrolly:
Everything except `codecov` passes. I can rebase or squash if needed.

henryiii:
I think it would be easier to review if you split it into two PRs (I’m okay with it combined, but we’ll need a couple of sign offs and the two changes are actually not related, so might be simpler for others). I’ll see if I can easily add a test for missing colorama when I’m at a computer.

src/build/__main__.py
henryiii: Yeah, Optional really means `str | None` and isn’t tried to optional parameters (unless they are defaulted to None, for example).  Some people who shall not be names hate the name Optional for it. ;)

henryiii:
@gaborbernat or anyone else, is there an easy way to run an "uninstall colorama" only on windows in tox? Would keep from having to list the test deps again, which is what I'm trying here to see if that fixes the coverage check.

layday:
Is there a probable downside to uninstalling colorama unconditionally?

henryiii:
Ahh, it's a warning not an error code? Then no.

layday:
Personally I would rather print with fush and format were separate function but I don't think it matters that much.

layday (approved):

abitrolly:
Is there anything else that needs to be done to push this? I'd like to propose some semantic changes to the `build` output, and the diff will conflict with this one.

henryiii:
We don’t need to make the coverage pass. I’m traveling till the 27th so can only try a little bit here and there. Currently it seems like the coverage info from the path job is not being reported properly.

henryiii (approved):
Sorry that took so long to fix!

henryiii:
@gaborbernat I'd like your sign-off on this, especially since I touched the tox config quite a bit. We now produce coverage info from all the runs, and use the `coverage` job to combine it (it was unused before).

src/build/__main__.py
```diff
@@ -46,6 +47,10 @@ def _init_colors() -> Dict[str, str]:
 _STYLES = _init_colors()
 
 
+def _cprint(fmt: str = '', msg: str = '') -> None:
+    print(fmt.format(msg, **_STYLES), flush=True)
```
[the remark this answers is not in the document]
abitrolly: Yes, because Python default output buffering is evil )

 https://github.com/pypa/build/pull/494#issuecomment-1171960475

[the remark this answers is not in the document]
henryiii: Absolutely no argument there!!! Dealing with this in cibuildwheel, scikit-build, etc.

PR #512, opened 2022-09-02
build: add circular dependency protection

layday:


FFY00 (approved):
Just some minor comments. Thanks!

PR #514, opened 2022-09-06
build: clarify that dependencies are build dependencies

abitrolly:
I would also change parameter `distribution` to `dist_type`, because it is not the actual package that is being meant here. Tell me if it is ok to do, and I'll add commits here.


layday:
Let's rename the paramater in a separate PR - I'd also like to change its type to `Literal['sdist', 'wheel']` wherever it's used.

You need to update the tests.

layday:
pre-commit.ci autofix

abitrolly:
Squashed.

FFY00 (approved):
Thanks!

PR #519, opened 2022-10-04
tests: better isolate test_venv_fail

tjni:
In [nixpkgs](https://github.com/NixOS/nixpkgs), we run the tests in parallel using `pytest-xdist`, and I can consistently see a failure on my M1 Mac machine in `test_main.py#test_venv_fail` related to the actual stdout containing color escape sequences when none were expected.

If I reload `build.__main__` as is done in other tests, the failure goes away for me.


layday (approved):

henryiii (approved):

layday:
Thanks!

PR #532, opened 2022-11-13
Drop Python 3.6

layday:
3.6 reached its EOL in Dec of last year and our main dependencies (pep517 and packaging) have already dropped support for it.


gaborbernat (approved):

henryiii:
What about upgrading pre-commit's pyupgrade flag?

layday:
Thanks, I've also updated black's targets and removed 3.6 from the tox matrix. I think that's all the 3.6 references 😅

layday:
No clue why the docs are failing.

layday:
I'm taking these logging line checks out, they are a major nuisance to have to update with every minor change that we make.

layday:
> No clue why the docs are failing.

https://github.com/sphinx-doc/sphinx/issues/9275#issuecomment-1312782811

I just silenced the error for now, it doesn't look like we can link to `PackageMetadata` from the stdlib.

layday:
@FFY00 Please take a look if you have time :)

FFY00 (approved):
Looks good. Thanks @layday!

Could you just fix the title of the first two commits before merging?

layday:
Yeah, I just wasn't sure what prefix would be appropriate. `meta:` and `multiple:`, or maybe `chore:`?

FFY00:
I would do `meta:` and `build:`, respectively.

layday:
You'll have to remove the 3.6 tests from the main branch protection rules, BTW.

henryiii (approved):

henryiii:
Done.

henryiii:
I’d like to add a “pass” job so we don’t have to keep changing this in the future.

PR #79, opened 2020-09-05
build: change directory to the project directory when running the hooks

FFY00:
Signed-off-by: Filipe Laíns <lains@archlinux.org>

Issue #68, opened 2020-08-08
Command does not change the current directory to project root


uranusjr:
PEP 517 mandates “all hooks are run with working directory set to the root of the source tree”. This is important for in-tree hooks since the cwd is relied on to find the PEP 517 modules.

This fails:

```
$ python -m build ./example-project -s -o .
ERROR Backend 'pep517' is not available
```

But this works:

```
$ cd example-project
$ python -m build -s -o .
$ ls *.tar.gz
my_package-2.tar.gz
```

FFY00:
Can you confirm this is working on `master`?

uranusjr:
The directory change works now, thanks! The change however causes a side effect on the `-o` option:

```
$ python -m build ./example-project -s -o .
$ ls
example-project
$ ls example-project
example_project-1.0.tar.gz
pyproject.toml
```

The output directory is resolved against the project root, but I’d expect it to be resolved against the initial working directory instead.

FFY00:
Okay, closing this but keeping the other issue open.

PR #535, opened 2022-11-20
build: protect _working_directory with lock to make build thread-safe

radoering:
In python-poetry/poetry#6205, we noticed that running multiple builds in parallel is not thread-safe because build changes the working directory at some places. Of course, downstream users can protect calls to `build` with locks by themselves but it might be nice to not have to worry about it.

### DIFF
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -14,6 +14,7 @@ import re
 import subprocess
 import sys
 import textwrap
+import threading
 import types
 import warnings
 import zipfile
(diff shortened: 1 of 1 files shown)


layday:
Hmm, do we actually need to chdir? Isn’t this handled by pep517?

Sent with [Proton Mail](https://proton.me/) secure email.

------- Original Message -------

layday:
I'm quite certain this is redundant. `_working_dir` is only called once in `ProjectBuilder._handle_backend`. `_handle_backend` is only ever called with a PEP 517 hook.  pyproject-hooks (formerly pep517) sets the cwd on the subprocess runner whenever it invokes a hook in https://github.com/pypa/pyproject-hooks/blob/5b5b66bc559bdf0d9c65979b784ab4d34f0662df/src/pyproject_hooks/_impl.py#L313.

radoering:
According to #79 and #68, I suppose it was at least necessary at that time.

layday:
I think the issue then was that the source directory wasn't an absolute path.

PR #536, opened 2022-11-20
build: remove redundant change of working directory

radoering:
As per https://github.com/pypa/build/pull/535#issuecomment-1321163124 changing the working directory seems to be redundant. So let's remove it. (This also makes build thread-safe.)

Closes #535

### DIFF
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -126,18 +126,6 @@ class TypoWarning(Warning):
     """
 
 
-@contextlib.contextmanager
-def _working_directory(path: str) -> Iterator[None]:
-    current = os.getcwd()
-
-    os.chdir(path)
-
-    try:
-        yield
-    finally:
-        os.chdir(current)
-
-
 def _validate_source_directory(srcdir: PathType) -> None:
     if not os.path.isdir(srcdir):
         raise BuildException(f'Source {srcdir} is not a directory')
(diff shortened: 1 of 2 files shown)


gaborbernat (approved):

henryiii:
Aren't build hooks inherently thread unsafe due to the requirement that they be called in the source directory? You do not pass the source directory into the build hook, it's not part of the interface defined in PEP 517. So the build hook must require the working directory be the source directory. You look up the pyproject.toml by cwd when implementing a hook, as well as whatever it is supposed to build.

layday:
> A redundant change is fine to remove, but there is always at least one, right?

Do you mean that there is at least one `os.chdir`? There isn't, build hooks are executed in a subprocess with the `cwd` set to the project root.

layday:
Thanks for this. Can you add a simple test to validate that a relative path is made absolute by the `ProjectBuilder` constructor?

layday (approved):

layday:
I'll rebase this on top of https://github.com/pypa/build/pull/532 as soon as it's unblocked, don't bother fixing the line no :)

layday:
Actually, can you rebase on top of main, @radoering? I don't want to rewrite your history.

layday:
Thanks!

PR #537, opened 2022-11-21
Refactor `IsolatedEnv`, take two

layday:
This is a more modest set of changes than I had in #361.  Changing the runner's signature to make removing variables from the environ possible can come in a follow-up PR.  Just as #361, this removes the env builder, redefines the isolated env interface, and adds an isolated env constructor to the project builder.

### DIFF
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -16,16 +16,23 @@ import os
 import re
 import subprocess
 import sys
-import textwrap
-import types
 import warnings
 import zipfile
 
 from collections import OrderedDict
-from collections.abc import Iterator, Set
-from typing import Any, Callable, Mapping, MutableMapping, Optional, Sequence, Tuple, Type, Union
+from collections.abc import Iterator
+from typing import Any, Callable, Mapping, MutableMapping, Optional, Sequence, Union
 
 import pyproject_hooks
+
+from ._exceptions import (
+    BuildBackendException,
+    BuildException,
+    BuildSystemTableValidationError,
+    FailedProcessError,
+    TypoWarning,
+)
+from ._util import check_dependency
 
 
 TOMLDecodeError: type[Exception]
@@ -46,7 +53,6 @@ else:
 RunnerType = Callable[[Sequence[str], Optional[str], Optional[Mapping[str, str]]], None]
 ConfigSettingsType = Mapping[str, Union[str, Sequence[str]]]
 PathType = Union[str, 'os.PathLike[str]']
-_ExcInfoType = Union[Tuple[Type[BaseException], BaseException, types.TracebackType], Tuple[None, None, None]]
 
 
 _WHEEL_NAME_REGEX = re.compile(
@@ -65,67 +71,6 @@ _DEFAULT_BACKEND = {
 _logger = logging.getLogger(__name__)
 
 
-class BuildException(Exception):
-    """
-    Exception raised by :class:`ProjectBuilder`
-    """
-
-
-class BuildBackendException(Exception):
-    """
-    Exception raised when a backend operation fails
-    """
-
-    def __init__(
-        self, exception: Exception, description: str | None = None, exc_info: _ExcInfoType = (None, None, None)
-    ) -> None:
-        super().__init__()
-        self.exception = exception
-        self.exc_info = exc_info
-        self._description = description
-
-    def __str__(self) -> str:
-        if self._description:
-            return self._description
-        return f'Backend operation failed: {self.exception!r}'
-
-
-class BuildSystemTableValidationError(BuildException):
-    """
-    Exception raised when the ``[build-system]`` table in pyproject.toml is invalid.
-    """
-
-    def __str__(self) -> str:
-        return f'Failed to validate `build-system` in pyproject.toml: {self.args[0]}'
-
-
-class FailedProcessError(Exception):
-    """
-    Exception raised when an setup or prepration operation fails.
-    """
-
-    def __init__(self, exception: subprocess.CalledProcessError, description: str) -> None:
-        super().__init__()
-        self.exception = exception
-        self._description = description
-
-    def __str__(self) -> str:
-        cmd = ' '.join(self.exception.cmd)
-        description = f"{self._description}\n  Command '{cmd}' failed with return code {self.exception.returncode}"
-        for stream_name in ('stdout', 'stderr'):
-            stream = getattr(self.exception, stream_name)
-            if stream:
-                description += f'\n  {stream_name}:\n'
-                description += textwrap.indent(stream.decode(), '    ')
-        return description
-
-
-class TypoWarning(Warning):
-    """
-    Warning raised when a possible typo is found
-    """
-
-
 def _validate_source_directory(srcdir: PathType) -> None:
     if not os.path.isdir(srcdir):
         raise BuildException(f'Source {srcdir} is not a directory')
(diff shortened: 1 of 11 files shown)


layday:
Is there something I can do to push this forward?

henryiii:
Just need to push a release, then this can be reviewed and merged. (Technically can be reviewed anytime, but at least I'm likely busy till after a release).

src/build/__init__.py
```diff
@@ -13,23 +13,29 @@
...
-import textwrap
-import types
 import warnings
 import zipfile
 
-from collections import OrderedDict
-from collections.abc import Iterator, Set
-from typing import Any, Callable, Mapping, MutableMapping, Optional, Sequence, Tuple, Type, Union
+from collections.abc import Iterator
+from typing import Any, Callable, Mapping, Optional, Sequence, TypeVar, Union
```
henryiii, line 22: Several of these can be imported from `collections.abc` - at least Callable, Mapping, Sequence, assumgin they are not used in a non annotation setting with parameters. Not a blocker.

src/build/__init__.py
```diff
@@ -13,23 +13,29 @@
...
 
 import pyproject_hooks
 
+from . import env
+from ._exceptions import (
```
henryiii, line 27: Does this need to be `_exceptions` instead of `exceptions`? Exceptions are not private, and putting them in a common place doesn't seem to be API we need to hide (the tracebacks will print this module name unless you set `__module__` manually). Reexporting them here is fine, but hiding the path to them doesn't seem necessary. But don't feel strongly.

src/build/__init__.py
```diff
@@ -13,23 +13,29 @@
...
+    BuildBackendException,
+    BuildException,
+    BuildSystemTableValidationError,
+    FailedProcessError,
+    TypoWarning,
+)
+from ._util import check_dependency, parse_wheel_filename
```
henryiii, line 34: We already have a `util`, also having a `_util` seems confusing? If it's important to hide these, they could be `_check_dependency` and `_parse_wheel_filename` in `util`? Or maybe they could be in a file with a better private name?

Having public and private utils separated in this way isn't terrible though, just checking.

src/build/__init__.py
```diff
@@ -194,7 +74,28 @@ def _find_typo(dictionary: Mapping[str, str], expected: str) -> None:
...
+
+
+def _read_pyproject_toml(path: PathType) -> Mapping[str, Any]:
+    try:
+        with open(path, 'rb') as f:
+            return toml_loads(f.read().decode())
+    except FileNotFoundError:
+        return {}
+    except PermissionError as e:
+        raise BuildException(f"{e.strerror}: '{e.filename}' ")  # noqa: B904 # use raise from
```
henryiii, line 93: Can't you add `from None` or `from e` here? That's less characters than the noqa and explicitly acknowledges the chaining (which is the point of B904).

(I think you want `from None` for both of these, as they contain the required details already, not `from e`, which is the implicit default)

(Edit: Looks like this was just a moved problem from another place, rather than an added problem)

src/build/__init__.py
```diff
@@ -233,102 +134,80 @@ def _parse_build_system_table(pyproject_toml: Mapping[str, Any]) -> dict[str, An
...
+
+    return _invoke_wrapped_runner
+
+
 class ProjectBuilder:
     """
     The PEP 517 consumer API.
     """
 
+    _DEFAULT_RUNNER = pyproject_hooks.default_subprocess_runner
```
henryiii, line 149: Is this class property really needed? A subclass can just inherit & overload `__init__`, changing the default of `runner`. This just simplifies this slightly at the expense of having two ways to do the same thing, with a private/harder to discover way being the added way.

src/build/__init__.py
```diff
@@ -233,102 +134,80 @@ def _parse_build_system_table(pyproject_toml: Mapping[str, Any]) -> dict[str, An
...
+
     def __init__(
-        self,
-        srcdir: PathType,
-        python_executable: str = sys.executable,
-        scripts_dir: str | None = None,
```
henryiii, line 245: Was scripts_dir support not needed, I guess?

henryiii (commented):
A few minor questions and thoughts.

src/build/__init__.py
henryiii: Okay.

src/build/__init__.py
layday: I didn't wanna add another place where they can be publicly imported from in this PR, but I'm supportive of making `_exceptions` public.

src/build/__init__.py
layday: I just factored it out from `__init__` - it was that way before :) But happy to raise `from None`.

src/build/__init__.py
layday: It was only because it is referenced both in `__init__` and `from_env` - the intention wasn't to simplify overwriting the runner (which you cannot do anyway in Python since function defaults are early-bound).  I could move it to the top level or double underscore it.

src/build/__init__.py
layday: Responsibility for extending the `$PATH` with `scripts_dir` has shifted to the isolated env protocol.  This opens up the possibility of having multiple `script_dir`s and simplifies the relationship between isolated env and builder, with the builder no longer having to manipulate the previously-manipulated environ.

src/build/__init__.py
henryiii: Yes they can, we just need `__all__` (which we should be adding anyway, along with `__dir__`).

src/build/__init__.py
henryiii: Fine to make that touchup in a followup, keeping moves simple is probably best. Though fine either way.

src/build/__init__.py
henryiii: Ahh, okay, that's fine then. I'd slightly prefer this be on the top level, as I could see someone wanting to use `class NewThing: _DEFAULT_RUNNER=other` to override this, which is what I'd usually see this pattern for. Does it really need to be aliased, couldn't it be left like before and just referenced twice?

layday: I've put it back.

src/build/__init__.py
layday: But that would make them public assuming they are contained in a public (i.e. ununderscored) module.

src/build/__init__.py
henryiii: Agree.

src/build/__init__.py
henryiii: Okay, fine for a followup then, renaming this & can be discussed there.

src/build/__init__.py
henryiii: I'd say that an underscored name in a `__all__` is still hidden, it just means it is used from other modules. Using it downstream is just as risky as if there was no `__all__`. `__all__` describes what is used by other modules - things not in __all__ should only be used in the file they reside in.

You could even filter things that start with an underscore in `__dir__`, but I'd not bother, things like IDEs and IPython still filter them by default.

layday: Are they also hidden from Sphinx?

layday: I think having a private module for internal use is cleaner FWIW, but I don't feel strongly about this.

henryiii: I'm fine with it either way. Just a note.

Sphinx hides things starting with an underscore by default, but it's configurable, IIRC.

henryiii (approved):
@gaborbernat or @FFY00, could we get one more pair of eyes? If we don't hear anything, let's merge Jan 25.

PR #538, opened 2022-11-21
Replace setup-cfg-fmt pre-commit hook with validate-pyproject

layday:


FFY00 (approved):
This makes sense. Thanks @layday!

PR #543, opened 2022-12-01
Introduce a centralized GHA check/gate job

webknjaz:
This patch is intended to lower the maintenance burden of having to manually go through each matrix subjob name in the branch protection repository settings. It allows to only include the `check` job in the branch protection and it will robustly determine if the dependencies have succeeded or not.

It is currently mostly serves the Python ecosystem in projects like aiohttp, attrs, cryptography, pydantic, open edX, pip etc. But I've also seen other communities picking it up lately, like the AWS Rust SDK and even the engine powering https://dev.to, to my surprise. Strictly speaking, it is agnostic.

Ref: https://github.com/marketplace/actions/alls-green#why

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,6 +45,16 @@ jobs:
           - "tox"
           - "min"
 
+    continue-on-error: >- # jobs not required in branch protection
+      ${{
+        (
+          startsWith(matrix.py, 'pypy-')
+          && (!endsWith(matrix.py, '-3.7') || matrix.os == 'windows')
+        )
+        && true
+        || false
+      }}
+
     steps:
       - uses: actions/checkout@v3
         with:
@@ -128,3 +138,19 @@ jobs:
 
       - name: Run check for type
         run: tox -e type --skip-pkg-install
+
+  # https://github.com/marketplace/actions/alls-green#why
+  required-checks-pass: # This job does nothing and is only used for the branch protection
+    if: always()
+
+    needs:
+      - pytest
+      - type
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Decide whether the needed jobs succeeded or failed
+        uses: re-actors/alls-green@release/v1
+        with:
+          jobs: ${{ toJSON(needs) }}


.github/workflows/test.yml
```diff
@@ -128,3 +141,19 @@ jobs:
 
       - name: Run check for type
         run: tox -e type --skip-pkg-install
+
+  # https://github.com/marketplace/actions/alls-green#why
+  check: # This job does nothing and is only used for the branch protection
```
henryiii, line 146: I have always been calling this "pass", not "check". That is a name I'd associate with a linting job, not for if all jobs passed.

.github/workflows/test.yml
```diff
@@ -45,6 +45,19 @@ jobs:
...
           - "min"
 
+    continue-on-error: >- # jobs not required in branch protection
+      ${{
+        (
+          (
+            startsWith(matrix.py, 'pypy-')
+            && (!endsWith(matrix.py, '-3.7') || matrix.os == 'windows')
+          )
+          || matrix.py == '3.11'
```
layday, line 55: You don't have to exclude 3.11 here - we just forgot to add it to the protection rules.

webknjaz: Ah, I'll drop the line, then. Hopefully, once this is merged, it'll be harder to forget such things since they won't be hidden somewhere in the repo settings only a few folks get to see.

webknjaz: Done!

.github/workflows/test.yml
webknjaz: @henryiii yeah, naming of the ducks is hard. The linting association is weird, though. The statuses that show up at the bottom of PRs are coming (mostly) from the Checks API, for example — all of them are checks, not just linting ones.
I originally wanted to call this [`gate`](https://gating.dev) but a colleague convinced me to go for something more generic, which I now regret. So I've gone for using `check` in the examples, and now it's spread everywhere. And I keep using it because it's easier to update things across multiple projects if they look the same :man_shrugging:

Some people rename the job id, others add a custom `name` value (I saw `all-green` in PyCA, `required-checks-pass` by hynek, `success` by nedbat, I think somebody used `all-successful` and probably some others that I didn't care to remember). FWIW, I'll change to whatever @layday will want. Most folks just stick with `check` and don't care too much.

layday: I like "checks-pass" or "required-checks-pass" -- they get the point across. "check" I too would associate with linting and "pass" I'd have trouble deciphering.

layday: Let's go with "required-checks-pass" unless Henry disagrees.

webknjaz: Done! Resolving.

henryiii: That works! I've been using "pass" in other places, and it works well in repos where there are multiple workflows - meson-python has three, for example, so there's things like tests-pass and docs-pass. tests-checks and docs-checks doesn't work nearly as well.

henryiii (approved):

.github/workflows/test.yml
webknjaz: Yes, many people don't immediately realize that this is per workflow. OTOH, with the current state of the reusable workflows, I'm leaning toward merging all the bits in a single one for having one gate for everything.

webknjaz: > meson-python has three, for example, so there's things like tests-pass and docs-pass

So I've gone to check those workflows and saw that the aggregation jobs are now commented out and the previous implementation was faulty: [mesonbuild/meson-python@`1ef1721` (#221)](https://github.com/mesonbuild/meson-python/pull/221/commits/1ef1721772e505b869b7ef924a38556d8ed0b2bf#r1038060140).

layday (approved):
Thanks!

webknjaz:
@layday anything else needed for the merge? Note that there's an action item to swap out the branch protection settings that is post-merge. Don't do this before merging, since it'll block other PRs until something is able to report a check with this name.

layday:
I don't have permission to change the branch protection settings, so I'll leave this to @henryiii or @FFY00 to merge.

webknjaz:
@layday FWIW merging this is harmless, it's the branch protection bit that requires extra privileges. But merging will just add an extra check reported but not used, which is fine. I'd even recommend doing so earlier than later because all the PRs with the old workflow definitions will have to be restarted to report this check. The earlier this happens, the less friction switching the branch protection setting will cause.

henryiii:
Thanks!

PR #544, opened 2022-12-01
Wrap CLI options in description with backticks

webknjaz:
This is necessary to make Sphinx display them as inline code and so it stops converting double dashes (`--`) to M-dashes (`—`).

### DIFF
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -262,15 +262,15 @@ def main_parser() -> argparse.ArgumentParser:
                 '''
                 A simple, correct PEP 517 build frontend.
 
-                By default, a source distribution (sdist) is built from {srcdir}
+                By default, a source distribution (sdist) is built from ``{srcdir}``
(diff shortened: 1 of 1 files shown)


webknjaz:
I've also noticed weird section titles like https://pypa-build.readthedocs.io/en/stable/#python--m-build-options and posted https://github.com/tox-dev/sphinx-argparse-cli/issues/68 upstream assuming it's their bug.

layday:
This doesn't seem to render correctly: https://pypa-build--544.org.readthedocs.build/en/544/

webknjaz:
Ah, that's a bummer! I suppose that it's up to @gaborbernat to check this upstream, then?

Should we keep the PR open for now?

layday:
Sure.

gaborbernat:
> Ah, that's a bummer! I suppose that it's up to @gaborbernat to check this upstream, then?

PR to fix that is welcome 😊

gaborbernat:
Seem stalled.

PR #545, opened 2022-12-01
types: fix mypy check

henryiii:
Currently CI is broken. Also mypy was really old.

Personaly, I'm a fan of putting the mypy check in pre-commit. It's fast and auto-updated.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,8 +60,9 @@ test = [
   'setuptools >= 56.0.0; python_version >= "3.10"',
 ]
 typing = [
-  "importlib-metadata >= 4.6.4",
-  "mypy == 0.942",
+  "importlib-metadata >= 5.1",
+  "mypy == 0.991",
+  "tomli",
   "typing-extensions >= 3.7.4.3",
 ]
 virtualenv = [
--- a/src/build/util.py
+++ b/src/build/util.py
@@ -22,8 +22,7 @@ else:
 def _project_wheel_metadata(builder: build.ProjectBuilder) -> importlib_metadata.PackageMetadata:
     with tempfile.TemporaryDirectory() as tmpdir:
         path = pathlib.Path(builder.metadata_path(tmpdir))
-        # https://github.com/python/importlib_metadata/pull/343
-        return importlib_metadata.PathDistribution(path).metadata  # type: ignore[arg-type]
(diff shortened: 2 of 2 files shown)


layday:
What's the reason we don't have it in pre-commit?

pyproject.toml
```diff
@@ -60,8 +60,9 @@ test = [
   'setuptools >= 56.0.0; python_version >= "3.10"',
 ]
 typing = [
-  "importlib-metadata >= 4.6.4",
-  "mypy == 0.942",
+  "importlib-metadata >= 5.1",
+  "mypy == 0.991",
+  "tomli",
```
layday, line 65: Why do wee need tomli here?

henryiii: Because if you run the type check from 3.11, it won't have tomli and the type checking fails due to a missing library. Same reason we have importlib-metadata.

layday: I was confused because mypy should not type check code which is unreachable.  But then I remembered we are enforcing 3.7 in our `pyproject.toml`:

```toml
python_version = "3.7"
```

Not the first time it tripped me up -- this should go in tox, not in our config.

layday (approved):

layday:
It's just that I vaguely recall there was a reason we kept it out of pre-commit but I don't remember what it was.

FFY00:
Yes, it was kept out because it takes too long. It's *mostly* fine if you have cache, but if you don't, it takes a while. I am -1 on this change.

henryiii:
"this change" meaning adding it to pre-commit? This PR just fixes the current failure.

FFY00:
Yes, adding mypy to pre-commit. Sorry for the confusion.

pyproject.toml
henryiii: This would mean mypy would likely be broken out of the box unless run though tox. The "current Python" default means we'd have to support type checking on all versions of Python, which is hard(er) or impossible (due to type ignores becoming invalid). We have designed our types (and type ignores) targeting 3.7, and if someone wants to target some other version, they can explicitly do it on the command line.

layday: I don't think that it's particularly hard or impossible (we are talking about two or three type ignores, all relating to importlib-metadata) and best practice would be to type check across all versions we support.  But we can tackle this in a separate PR.

PR #550, opened 2022-12-10
tests: fix for PEP 685 in packaging 22

henryiii:
Support Packaging 22.

### DIFF
--- a/tests/test_projectbuilder.py
+++ b/tests/test_projectbuilder.py
@@ -60,13 +60,13 @@ class ExtraMockDistribution(MockDistribution):
                 Metadata-Version: 2.2
                 Name: extras_dep
                 Version: 1.0.0
-                Provides-Extra: extra_without_associated_deps
-                Provides-Extra: extra_with_unmet_deps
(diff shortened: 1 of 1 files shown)


layday:
If packaging doesn't normalise dependency names when stringifying dependencies, why does it normalise their extras?  Anyway, the easiest thing to do here is to replace all underscores in extra names in the tests with hyphens.

layday:
This only affects extra names in markers, BTW, not square bracket extra selectors.

nanonyme:
That sounds inconsistent enough to definitely warrant an upstream issue.

henryiii:
But those have to be consistent to make older packaging pass, since older packaging does not consider `a-b` to match `a_b`.

henryiii:
RtD is misbehaving today, also in pypa/packaging.

layday (approved):

henryiii:
> RtD supposed to work from forks at all?

I think so. Never had a problem before.

henryiii:
Passes now, so yes. :)

PR #563, opened 2023-01-16
Don't require toml in [test] on Python 3.11+

hroncok:
The test_toml_instead_of_tomli test is skipped on Python 3.11 anyway.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,10 +54,10 @@ test = [
   "pytest-mock >= 2",
   "pytest-rerunfailures >= 9.1",
   "pytest-xdist >= 1.34",
-  "toml >= 0.10.0",
   "wheel >= 0.36.0",
   'setuptools >= 42.0.0; python_version < "3.10"',
   'setuptools >= 56.0.0; python_version >= "3.10"',
+  'toml >= 0.10.0; python_version < "3.11"'
 ]
 typing = [
   "importlib-metadata >= 5.1",


hroncok:
If you wish to drop the fallback to toml, that also works for me.

frenzymadness:
Now there is `'tomli >= 1.1.0; python_version < "3.11"'` in dependencies, `"toml >= 0.10.0"` in test deps and `"tomli"` in typing deps.

layday:
We have toml in test deps because we are testing against it.  We have tomli in the typing deps because mypy simulates Python 3.7 and on 3.11, tomli is not required.

hroncok:
> Now there is `'tomli >= 1.1.0; python_version < "3.11"'` in dependencies, `"toml >= 0.10.0"` in test deps and `"tomli"` in typing deps.

After this PR, the toml dependency is also conditional.

henryiii (approved):

henryiii:
I think we should drop toml, but this is a (small) step forward.

layday:
I've got a PR to drop toml ready to go but I wanted to get #537 in first to avoid having to deal with merge conflicts.  If you wanna get in this in the interim, we can do so, but I don't think there's much point unless we are gonna cut a release without `toml` as a test dependency.

layday:
https://github.com/pypa/build/pull/567

layday:
Merged #567.

PR #565, opened 2023-01-20
infra: replace flake8 with ruff

layday:
As a first step, this ports all of our existing flake8 rules to Ruff. I have omitted pyupgrade for which Ruff has partial support.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -44,7 +44,7 @@ repos:
     hooks:
       - id: isort
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: v0.0.227
+    rev: v0.0.228
     hooks:
       - id: ruff
         args: [--fix, --format, grouped]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -43,12 +43,11 @@ repos:
     rev: 5.11.4
     hooks:
       - id: isort
-  - repo: https://github.com/PyCQA/flake8
-    rev: "6.0.0"
+  - repo: https://github.com/charliermarsh/ruff-pre-commit
+    rev: v0.0.226
     hooks:
-      - id: flake8
-        additional_dependencies: ["flake8-bugbear==22.7.1"]
-        language_version: python3.9
+      - id: ruff
+        args: [--force-exclude, --format, grouped, --show-source]
   - repo: https://github.com/codespell-project/codespell
     rev: "v2.2.2"
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -47,7 +47,7 @@ repos:
     rev: v0.0.226
     hooks:
       - id: ruff
-        args: [--force-exclude, --format, grouped]
+        args: [--fix, --format, grouped]
   - repo: https://github.com/codespell-project/codespell
     rev: "v2.2.2"
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -44,7 +44,7 @@ repos:
     hooks:
       - id: isort
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: v0.0.226
+    rev: v0.0.227
     hooks:
       - id: ruff
         args: [--fix, --format, grouped]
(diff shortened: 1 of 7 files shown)


henryiii:
You beat me (by about a day). :)

tests/test_integration.py
```diff
@@ -124,7 +124,7 @@ def test_build(monkeypatch, project, args, call, tmp_path):
 
 def test_isolation(tmp_dir, package_test_flit, mocker):
     try:
-        import flit_core  # noqa: F401
+        import flit_core  # noqa: F401  # imported but unused
```
henryiii, line 127: Just curious, would `import flit_core as _` work?

.pre-commit-config.yaml
```diff
@@ -43,12 +43,11 @@ repos:
...
-  - repo: https://github.com/PyCQA/flake8
-    rev: "6.0.0"
+  - repo: https://github.com/charliermarsh/ruff-pre-commit
+    rev: v0.0.226
     hooks:
-      - id: flake8
-        additional_dependencies: ["flake8-bugbear==22.7.1"]
-        language_version: python3.9
+      - id: ruff
+        args: [--force-exclude, --format, grouped, --show-source]
```
henryiii, line 50: Do we need `show-source`? IMO, the default looks just like a normal pre-commit check, and the paths are in the correct format for quick linking to source (in iterm, vscode, etc).

henryiii: I also think we can use `--fix`; depending on what checks end up getting added, we can always disable a fix for a check if we don't like it.

henryiii:
Ruff can also replace pyupgrade (the support is nearly there, only 1-2 fixes not supported - I think it's fine for a continuous check), isort, and the Python parts of the pygrep hooks. No need to do that right now, but would be a good followup.

tests/test_integration.py
henryiii: I guess the correct way to do this would be checking `importlib.util.find_spec("flit_core")` against `None`.

.pre-commit-config.yaml
layday: > Do we need show-source?

eh, I just like having the error highlighted. I'll take it out.

> I also think we can use --fix; depending on what checks end up getting added, we can always disable a fix for a check if we don't like it.

How does that work in CI with pre-commit?  Wil pre-commit automatically commit fixes?

tests/test_integration.py
layday: > Just curious, would `import flit_core as _` work?

It does not.

.pre-commit-config.yaml
Czaki: > How does that work in CI with pre-commit? Will pre-commit automatically commit fixes?

Yes. If you need I could look for an example commit.

layday: Thanks, I'll add `--fix` then.

henryiii: In the next version of the pre-commit action (AKA next version of ruff), you can drop the `--force-exclude` too. https://github.com/charliermarsh/ruff-pre-commit/pull/20

layday: Added `--fix` and dropped `--force-exclude`.

henryiii: Wow, that was fast. 👍

henryiii: If you wanted an example: https://github.com/pypa/cibuildwheel/pull/1405/commits/6857bd5a3bcb019a4b0a2881bd08b0800d81c0de

henryiii (approved):

tests/test_integration.py
webknjaz: Could try adding `del flit_core` to the else-block, then.

henryiii: `importlib.util.find_spec("flit_core")` is the correct solution imo. It’s also faster if the module is large, since you don’t have to load it. Not that important for tests, but nice to do things the right way as a good example.

henryiii: And you wouldn’t need an else block for the del. It’s never used, so it could be in the try.

layday: You'd just need to reference it; `flit_core` on a new line would be sufficient.  I don't think we need to do any of those things however, silencing the warning where it does not apply is perfectly fine.

webknjaz: @henryiii yep, using importlib is certainly better, but I was suggesting to use the fact that the else-block is already present, w/o much refactoring.
Also, referencing it in the try-section is a code smell because it should always have exactly one instruction.

PR #573, opened 2023-01-31
chore: fix ruff configuration and add a few checks

henryiii:
I was looking at adding a few checks (trying to check the validity of replacing isort), and discovered fixes/checks weren't being applied. It turns out `build` is being ignored by default! For the pre-commit check, I don't think the default excludes really make much sense, and I've got other projects with a submodule named build, too. I see why it might be ignored by default if you aren't running in pre-commit. CC @charliermarsh, would it make sense to pass --exclude="" or something for the pre-commit hook?


charliermarsh:
@henryiii - I've gotta think on this -- the confluence of all the different settings is making my head hurt 😅 I don't _think_ we can blindly set `--exclude ""` because users could have valid, custom exclusions set there, that could include Python files that are checked in to Git, and so we'd be linting them inadvertently.

layday:
Did you mean to replace isort with ruff? You haven't opted into it.

henryiii:
Not yet. I hit this and wanted to fix it first. And it might not be configurable enough yet.

layday:
But you removed isort's config.

henryiii:
Oops, thought I reverted that. 😅

henryiii:
That's better. Since we have custom spacing set for isort (not sure why?), I'm pretty sure we can't switch isort to Ruff (without changing it, that is).

I'm not sure what the correct solution to the exclude problem is, but it does seem that the current situation is not ideal. But I might just be more susceptible to it, due to working on a lot of building stuff.  :) I do think it's better to err on the side of not silently ignoring directors, as a linter can't tell if something is missed, it just never reports anything for that directory. I wouldn't mind adding manually adding extend-excludes for directories that are pretty common but not hidden.

layday:
I think you accidentally reverted the ruff config changes too 😅

layday (approved):

PR #578, opened 2023-02-09
Specify encoding.

### LINKED ISSUE #577: EncodingWarning when PYTHONWARNDEFAULTENCODING
```
  /Users/jaraco/code/jaraco/jaraco.net/.tox/python/lib/python3.11/site-packages/build/env.py:201: EncodingWarning: 'encoding' argument not specified
    with tempfile.NamedTemporaryFile('w+', prefix='build-reqs-', suffix='.txt', delete=False) as req_file:
```

jaraco:
Fixes failures when PYTHONWARNDEFAULTENCODING=1. Fixes #577.

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -130,7 +130,7 @@ class DefaultIsolatedEnv(IsolatedEnv):
 
         # pip does not honour environment markers in command line arguments
         # but it does for requirements from a file
-        with tempfile.NamedTemporaryFile('w', prefix='build-reqs-', suffix='.txt', delete=False) as req_file:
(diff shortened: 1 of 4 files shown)


layday (approved):
Thanks!

henryiii (approved):
I though we had a static check for this - thanks!


Hmm, not seeing one in Ruff. Not sure to find one, then. I guess we can set the PYTHONWARNDEFAULTENCODING variable in the tests.

jaraco:
I recommend setting that var. that’s what I’ve done in other projects and how i found this one. I’ll update the PR.

layday:
I would prefer if the env var was injected with `pytest-dev` so that the tests behave identically when run directly with pytest.

Issue #587, opened 2023-03-14
Strange test failure with pip 23.0.1


frenzymadness:
I know it's not your problem at all but you might know enough to help me with the following strange problem causing `test_output_env_subprocess_error[color]` to fail.

I maintain python-build RPM package in Fedora and it now fails to build since we updated pip from 22.3.1 to 23.0.1. The problem is this failing test:
```
=================================== FAILURES ===================================
___________________ test_output_env_subprocess_error[color] ____________________

mocker = <pytest_mock.plugin.MockerFixture object at 0x7f76f42d7f10>
monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x7f76f42bf5d0>
main_reload_styles = None
package_test_invalid_requirements = '/builddir/build/BUILD/build-0.10.0/tests/packages/test-invalid-requirements'
tmp_dir = '/tmp/python-build-test-0cztn5fn'
capsys = <_pytest.capture.CaptureFixture object at 0x7f76f43265d0>, color = True
stdout_body = ['\x1b[1m* Creating venv isolated environment...\x1b[0m', '\x1b[1m* Installing packages in isolated environment... (setuptools >= 42.0.0, this is invalid, wheel >= 0.36.0)\x1b[0m', '', '\x1b[2mTraceback (most recent call last):']
stdout_error = '\x1b[91mERROR\x1b[0m '

    @pytest.mark.pypy3323bug
    @pytest.mark.parametrize(
        ('color', 'stdout_error', 'stdout_body'),
        [
            (
                False,
                'ERROR ',
                [
                    '* Creating venv isolated environment...',
                    '* Installing packages in isolated environment... (setuptools >= 42.0.0, this is invalid, wheel >= 0.36.0)',
                    '',
                    'Traceback (most recent call last):',
                ],
            ),
            (
                True,
                '\33[91mERROR\33[0m ',
                [
                    '\33[1m* Creating venv isolated environment...\33[0m',
                    '\33[1m* Installing packages in isolated environment... '
                    '(setuptools >= 42.0.0, this is invalid, wheel >= 0.36.0)\33[0m',
                    '',
                    '\33[2mTraceback (most recent call last):',
                ],
            ),
        ],
        ids=['no-color', 'color'],
    )
    def test_output_env_subprocess_error(
        mocker,
        monkeypatch,
        main_reload_styles,
        package_test_invalid_requirements,
        tmp_dir,
        capsys,
        color,
        stdout_body,
        stdout_error,
    ):
[...]

PR #588, opened 2023-03-14
tests: report installed versions of common packages

henryiii:
In https://github.com/pypa/build/issues/587, the exact version of pip would help debug why our CI isn't seeing this issue (at least I think it would). Let's report the versions of things we care about in our test header to make such debugging easier.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,6 +48,7 @@ docs = [
 ]
 test = [
   "filelock >= 3",
+  'importlib-metadata; python_version<"3.8"',
   "pytest >= 6.2.4",
   "pytest-cov >= 2.12",
   "pytest-mock >= 2",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,6 @@ docs = [
 ]
 test = [
   "filelock >= 3",
-  'importlib-metadata; python_version<"3.8"',
   "pytest >= 6.2.4",
   "pytest-cov >= 2.12",
   "pytest-mock >= 2",
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -132,9 +132,8 @@ def pytest_report_header() -> str:
     ]
     valid = []
     for package in interesting_packages:
+        # Old versions of importlib_metadata made this FileNotFoundError
         with contextlib.suppress(ModuleNotFoundError, FileNotFoundError):
             valid.append(f'{package}=={metadata.version(package)}')
     reqs = ' '.join(valid)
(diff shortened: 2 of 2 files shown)


tests/conftest.py
```diff
@@ -109,3 +115,26 @@ def tmp_dir():
...
+        'wheel',
+    ]
+    valid = []
+    for package in interesting_packages:
+        with contextlib.suppress(ModuleNotFoundError):
+            valid.append(f'{package}=={metadata.version(package)}')
+    reqs = ' '.join(valid)
+    pkg_line = f'installed packages of interest: {reqs}'
+
+    return '\n'.join([pkg_line])
```
layday, line 140:
```suggestion
    return pkg_line
```

henryiii: This was originally designed to build up several new lines. Though even where I first wrote it, it's now down to this line only. This is certainly the only one that currently makes sense here. But that's why it was weirdly written. :)

tests/conftest.py
```diff
@@ -109,3 +115,26 @@ def tmp_dir():
...
+        'pip',
+        'pyproject_hooks',
+        'setuptools',
+        'tomli',
+        'virtualenv',
+        'wheel',
+    ]
+    valid = []
+    for package in interesting_packages:
+        with contextlib.suppress(ModuleNotFoundError):
```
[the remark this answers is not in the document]
henryiii: Oh, it's FileNotFoundError???

tests/conftest.py
```diff
@@ -109,3 +115,26 @@ def tmp_dir():
...
+        'pip',
+        'pyproject_hooks',
+        'setuptools',
+        'tomli',
+        'virtualenv',
+        'wheel',
+    ]
+    valid = []
+    for package in interesting_packages:
+        with contextlib.suppress(ImportError):
```
henryiii, line 135:
```suggestion
        with contextlib.suppress(ModuleNotFoundError, FileNotFoundError):
```

layday (approved):

pyproject.toml
```diff
@@ -48,6 +48,7 @@ docs = [
 ]
 test = [
   "filelock >= 3",
+  'importlib-metadata; python_version<"3.8"',
```
layday, line 51: This shouldn't be necessary, we have `importlib-metadata` on line 38.

henryiii: Oops, I saw it in typing and didn't look to see if it was in dependencies! Thanks!

FFY00 (approved):

PR #593, opened 2023-03-21
build: add circular dependency checker for build requirements

jameshilliard:
Implement a basic build requirement cycle detector per PEP-517:

- Project build requirements will define a directed graph of requirements (project A needs B to build, B needs C and D, etc.) This graph MUST NOT contain cycles. If (due to lack of co-ordination between projects, for example) a cycle is present, front ends MAY refuse to build the project.

- Front ends SHOULD check explicitly for requirement cycles, and terminate the build with an informative message if one is found.

See:
https://www.python.org/dev/peps/pep-0517/#build-requirements

### DIFF
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -21,6 +21,7 @@ import zipfile
 from collections.abc import Iterator
 from typing import Any, Callable, Mapping, Optional, Sequence, TypeVar, Union
 
+import packaging.utils
 import pyproject_hooks
 
 from . import env
@@ -28,11 +29,13 @@ from ._exceptions import (
     BuildBackendException,
     BuildException,
     BuildSystemTableValidationError,
+    CircularBuildDependencyError,
     FailedProcessError,
+    ProjectTableValidationError,
     TypoWarning,
+    ProjectNameValidationError,
 )
-from ._util import check_dependency, parse_wheel_filename
-
+from ._util import check_dependency, parse_wheel_filename, project_name_from_path
 
 if sys.version_info >= (3, 11):
     import tomllib
@@ -126,6 +129,23 @@ def _parse_build_system_table(pyproject_toml: Mapping[str, Any]) -> Mapping[str,
     return build_system_table
 
 
+def _parse_project_name(pyproject_toml: Mapping[str, Any]) -> str | None:
+    if 'project' not in pyproject_toml:
+        return None
+
+    project_table = dict(pyproject_toml['project'])
+
+    # If [project] is present, it must have a ``name`` field (per PEP 621)
+    if 'name' not in project_table:
+        raise ProjectTableValidationError('`project` must have a `name` field')
+
+    project_name = project_table['name']
+    if not isinstance(project_name, str):
+        raise ProjectTableValidationError('`name` field in `project` must be a string')
+
+    return project_name
+
+
 def _wrap_subprocess_runner(runner: RunnerType, env: env.IsolatedEnv) -> RunnerType:
     def _invoke_wrapped_runner(cmd: Sequence[str], cwd: str | None, extra_environ: Mapping[str, str] | None) -> None:
         runner(cmd, cwd, {**(env.make_extra_environ() or {}), **(extra_environ or {})})
@@ -168,9 +188,17 @@ class ProjectBuilder:
         self._runner = runner
 
         pyproject_toml_path = os.path.join(source_dir, 'pyproject.toml')
-        self._build_system = _parse_build_system_table(_read_pyproject_toml(pyproject_toml_path))
+        pyproject_toml = _read_pyproject_toml(pyproject_toml_path)
+        self._build_system = _parse_build_system_table(pyproject_toml)
 
+        self._project_name: str | None = None
+        self._project_name_source: str | None = None
+        project_name = _parse_project_name(pyproject_toml)
+        if project_name:
+            self._update_project_name(project_name, 'pyproject.toml [project] table')
         self._backend = self._build_system['build-backend']
+
+        self._check_dependencies_incomplete: dict[str, bool] = {'wheel': False, 'sdist': False}
 
         self._hook = pyproject_hooks.BuildBackendHookCaller(
             self._source_dir,
@@ -199,6 +227,15 @@ class ProjectBuilder:
         return self._source_dir
 
     @property
+    def project_name(self) -> str | None:
+        """
+        The canonicalized project name.
+        """
+        if self._project_name is not None:
+            return packaging.utils.canonicalize_name(self._project_name)
+        return None
+
+    @property
     def python_executable(self) -> str:
         """
         The Python executable used to invoke the backend.
@@ -214,7 +251,9 @@ class ProjectBuilder:
         """
         return set(self._build_system['requires'])
 
-    def get_requires_for_build(self, distribution: str, config_settings: ConfigSettingsType | None = None) -> set[str]:
+    def get_requires_for_build(
+        self, distribution: str, config_settings: ConfigSettingsType | None = None, finalize: bool = False
+    ) -> set[str]:
         """
         Return the dependencies defined by the backend in addition to
         :attr:`build_system_requires` for a given distribution.
@@ -223,14 +262,26 @@ class ProjectBuilder:
             (``sdist`` or ``wheel``)
         :param config_settings: Config settings for the build backend
         """
-        self.log(f'Getting build dependencies for {distribution}...')
+        if not finalize:
+            self.log(f'Getting build dependencies for {distribution}...')
         hook_name = f'get_requires_for_build_{distribution}'
         get_requires = getattr(self._hook, hook_name)
 
         with self._handle_backend(hook_name):
             return set(get_requires(config_settings))
 
-    def check_dependencies(self, distribution: str, config_settings: ConfigSettingsType | None = None) -> set[tuple[str, ...]]:
+    def check_build_system_dependencies(self) -> set[tuple[str, ...]]:
+        """
+        Return the dependencies which are not satisfied from
+        :attr:`build_system_requires`
+
+        :returns: Set of variable-length unmet dependency tuples
+        """
+        return {u for d in self.build_system_requires for u in check_dependency(d, project_name=self._project_name)}
+
+    def check_dependencies(
+        self, distribution: str, config_settings: ConfigSettingsType | None = None, finalize: bool = False
+    ) -> set[tuple[str, ...]]:
         """
         Return the dependencies which are not satisfied from the combined set of
         :attr:`build_system_requires` and :meth:`get_requires_for_build` for a given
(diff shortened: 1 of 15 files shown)


src/build/__init__.py
```diff
@@ -28,6 +29,7 @@
     BuildBackendException,
     BuildException,
     BuildSystemTableValidationError,
+    CircularBuildSystemDependencyError,  # noqa: F401
```
layday, line 32: You want to re-export this by adding it to `__all__` - drop the noqa.

src/build/__init__.py
```diff
@@ -47,6 +49,16 @@
 _TProjectBuilder = TypeVar('_TProjectBuilder', bound='ProjectBuilder')
 
 
+_SDIST_NAME_REGEX = re.compile(r'(?P<distribution>.+)-(?P<version>.+)\.tar.gz')
```
layday, line 52: Move to `_utils`.

src/build/__init__.py
```diff
@@ -47,6 +49,16 @@
...
+
+
+_WHEEL_NAME_REGEX = re.compile(
```
layday, line 55: Already exists in `_utils`.

src/build/__init__.py
```diff
@@ -133,6 +145,23 @@ def _invoke_wrapped_runner(cmd: Sequence[str], cwd: str | None, extra_environ: M
     return _invoke_wrapped_runner
 
 
+def _parse_project_name(pyproject_toml: Mapping[str, Any]) -> str | None:
```
layday, line 148: Move this directly under `_parse_build_system_table`.

src/build/__init__.py
```diff
@@ -133,6 +145,23 @@ def _invoke_wrapped_runner(cmd: Sequence[str], cwd: str | None, extra_environ: M
...
+    if 'project' not in pyproject_toml:
+        return None
+
+    project_table = dict(pyproject_toml['project'])
+
+    # If [project] is present, it must have a ``name`` field (per PEP 621)
+    if 'name' not in project_table:
+        return None
```
layday, line 156: This is an error condition.

src/build/__init__.py
```diff
@@ -133,6 +145,23 @@ def _invoke_wrapped_runner(cmd: Sequence[str], cwd: str | None, extra_environ: M
...
+
+    project_name = project_table['name']
+    if not isinstance(project_name, str):
+        return None
```
layday, line 160: Likewise.

src/build/__init__.py
```diff
@@ -170,8 +199,10 @@ def __init__(
         pyproject_toml_path = os.path.join(source_dir, 'pyproject.toml')
         self._build_system = _parse_build_system_table(_read_pyproject_toml(pyproject_toml_path))
 
+        self.project_name: str | None = _parse_project_name(_read_pyproject_toml(pyproject_toml_path))
         self._backend = self._build_system['build-backend']
 
+        self._requires_for_build_cache: dict[str, set[str] | None] = {'wheel': None, 'sdist': None}
```
layday, line 205: I really don't want us having to deal with caching issues - if you wanna "cache" the result you can save it to a variable.  If it is imperative that the output of `requires_for_build_x` is cached, you should have to opt into reading the requires from the cache.

src/build/__init__.py
```diff
@@ -286,7 +356,15 @@ def build(
...
         kwargs = {} if metadata_directory is None else {'metadata_directory': metadata_directory}
-        return self._call_backend(f'build_{distribution}', output_directory, config_settings, **kwargs)
+        basename = self._call_backend(f'build_{distribution}', output_directory, config_settings, **kwargs)
+        match = None
+        if distribution == 'wheel':
+            match = _WHEEL_NAME_REGEX.match(os.path.basename(basename))
+        elif distribution == 'sdist':
+            match = _SDIST_NAME_REGEX.match(os.path.basename(basename))
+        if match:
+            self.project_name = match['distribution']
```
layday, line 366: This should be a utility function.

src/build/_exceptions.py
```diff
@@ -64,6 +64,30 @@ def __str__(self) -> str:
         return description
 
 
+class CircularBuildSystemDependencyError(BuildException):
```
layday, line 67: This is any build dependency, not just build system dependencies.

```suggestion
class CircularBuildDependencyError(BuildException):
```

src/build/_util.py
```diff
@@ -13,8 +14,17 @@
 )
 
 
-def check_dependency(
-    req_string: str, ancestral_req_strings: tuple[str, ...] = (), parent_extras: Set[str] = frozenset()
+# https://www.python.org/dev/peps/pep-0503/#normalized-names
+def _normalize(name: str) -> str:
```
layday, line 18: `packaging.utils` provides this as `canonicalize_name`.

src/build/_util.py
```diff
@@ -13,8 +14,17 @@
...
+    return re.sub(r'[-_.]+', '-', name).lower()
+
+
+def check_dependency(  # noqa: C901
+    req_string: str,
+    ancestral_req_strings: tuple[str, ...] = (),
+    parent_extras: Set[str] = frozenset(),
+    project_name: str | None = None,
```
layday, line 26: You don't need a separate `project_name` - you just need to pass the project name as an `ancestral_req_strings` 1-tuple.

src/build/_util.py
```diff
@@ -48,6 +58,12 @@ def check_dependency(
             # dependency is satisfied.
             return
 
+    # Front ends SHOULD check explicitly for requirement cycles, and
+    # terminate the build with an informative message if one is found.
+    # https://www.python.org/dev/peps/pep-0517/#build-requirements
+    if project_name is not None and _normalize(req.name) == _normalize(project_name):
```
layday, line 64: You should perform this check on line 48; the current check is deficient.  It should be:

```py
    req = packaging.requirements.Requirement(req_string)
    canonicalized_name = packaging.utils.canonicalize_name(req.name)

[...]

    if any(
        packaging.utils.canonicalize_name(packaging.requirements.Requirement(a).name) == canonicalized_name
        for a in ancestral_req_strings
    ):
        raise CircularBuildDependencyError(...)

```

src/build/__init__.py
```diff
@@ -301,13 +379,17 @@ def metadata_path(self, output_directory: PathType) -> str:
...
+            if match:
+                self.project_name = match['distribution']
             return metadata
 
         # fallback to build_wheel hook
         wheel = self.build('wheel', output_directory)
         match = parse_wheel_filename(os.path.basename(wheel))
         if not match:
             raise ValueError('Invalid wheel')
+        self.project_name = match['distribution']
```
layday, line 392: I'm not very comfortable with mutating the builder in a bunch of different places. If we wanna revalidate the build deps I'd rather we reinitialised the builder by passing the dist name as an explicit parameter.

src/build/__init__.py
jameshilliard: Do we want this to trigger a fatal error? Returning `None` here means we fallback to pulling the project name from the filename instead as if there was no `project` entry at all.

layday: Yeah, I think we should error if the project table is invalid - if it isn't rejected by us it'll be (or should be) rejected by the backend.

jameshilliard: Changed this to an exception, looks like the `test_build_with_dep_on_console_script` test `project` had no `name` field so I added one there as well.

src/build/__init__.py
jameshilliard: I recall I had implemented this caching to try and avoid calling `get_requires_for_build_wheel` or `get_requires_for_build_sdist` more than once for a build. I'm thinking this caching behavior probably isn't actually necessary if I were to just re-query `get_requires_for_build_{distribution}` a second time. I'll try and rework this.

src/build/__init__.py
jameshilliard: > I'm not very comfortable with mutating the builder in a bunch of different places.

Yeah, this is def a bit messy, unfortunately we don't have a pep517 hook to get our own project name so I'm attempting to pull the name whenever possible.

> If we wanna revalidate the build deps I'd rather we reinitialised the builder by passing the dist name as an explicit parameter.

Well re-validation can often be skipped, if we have a `name` in out `pyproject.yaml` `project` table we can grab it from there to do self cycle checks from the start. It's only when we can't determine our own project name before the build that we need to revalidate it AFAIU.

src/build/__init__.py
```diff
@@ -170,8 +188,10 @@ def __init__(
         pyproject_toml_path = os.path.join(source_dir, 'pyproject.toml')
         self._build_system = _parse_build_system_table(_read_pyproject_toml(pyproject_toml_path))
 
+        self.project_name: str | None = _parse_project_name(_read_pyproject_toml(pyproject_toml_path))
```
layday, line 191: Only read `pyproject.toml` once.

src/build/_util.py
jameshilliard: This approach seems like it might be problematic due to the `CircularBuildDependencyError` exception class having special handling for `self.backend` which is used for tracing circular build dependencies via the `get_requires_for_build_{distribution}` hooks.

src/build/_util.py
jameshilliard: > the current check is deficient

I'm having trouble getting this working, can you clarify what condition this is intended to catch.

I added a bunch of additional tests to ensure we are catching the various different dependency cycles.

src/build/__init__.py
jameshilliard: I removed the caching which it appears I had implemented to avoid log message changes breaking some other tests, I tweaked the tests and used a different approach which seems a bit simpler.

src/build/__init__.py
jameshilliard: > If we wanna revalidate the build deps I'd rather we reinitialised the builder by passing the dist name as an explicit parameter.

I'm thinking we should probably be validating the normalized project `name` in the various areas where it can potentially be retrieved are all the same and throw an exception if any are different. Right now there doesn't seem to be anything ensuring that a project name set via say the `pyproject.toml` `project` table matches the `wheel`, `sdist` and `dist-info` project names pulled via the regexes.

src/build/_util.py
layday: You'll have to explain why it's problematic.

layday: Ignore this actually.  Circular runtime deps are no factor - we can't repurpose `ancestral_req_strings`.

src/build/_util.py
layday: Apologies, this is fine as it is.

src/build/__init__.py
jameshilliard: I've implemented project name validation which will throw a `ProjectNameValidationError` exception if at any point a name update is inconsistent with the previous name update. This should ensure that builder project name state changes are done in a way that effectively locks in the normalized project name once one is available.

gaborbernat:
Seems it stalled, @jameshilliard let me know if you want to pick it up again and we can reopen.

jameshilliard:
> Seems it stalled, @jameshilliard let me know if you want to pick it up again and we can reopen.

I think this was ready but just waiting on review, or were there unresolved issues?

gaborbernat:
@jameshilliard can you rebase it against main; seem has merge conflicts now...

jameshilliard:
> @jameshilliard can you rebase it against main

Sure, I'll try and take a look later today.

PR #601, opened 2023-04-08
Compat with typed pyproject-hooks

layday:


src/build/__init__.py
```diff
@@ -40,7 +40,18 @@
...
+class RunnerType(Protocol):
+    """A protocol for the subprocess runner."""
+
+    def __call__(
+        self,
+        cmd: Sequence[str],
+        cwd: str | None = None,
+        extra_environ: Mapping[str, str] | None = None,
+    ) -> None:
+        ...
```
henryiii, line 52: We can set `...` as uncoverable in coverage's config.

src/build/_util.py
```diff
@@ -1,10 +1,19 @@
...
+import abc
 import re
 import sys
+import typing
 
 from collections.abc import Iterator, Set
 
+if sys.version_info >= (3, 8):
+    from typing import Protocol as Protocol
+elif typing.TYPE_CHECKING:
```
henryiii, line 12: This too

pradyunsg:
Consider using `from pyproject_hooks import SubprocessRunner` behind a `TYPE_CHECKING`  conditional instead of redefining the protocol.

layday:
> Consider using `from pyproject_hooks import SubprocessRunner` behind a `TYPE_CHECKING` conditional instead of redefining the protocol.

`SubprocessRunner` is not public :)

layday:
All of our other types are also available at runtime, so it'd be odd for this one not to be.

pradyunsg:
Look at the PR again. ;)

gaborbernat:
Seems it stalled, @layday let me know if you want to pick it up again and we can reopen.

layday:
This was in draft until a version of pyproject-hooks is released with type hints, which hasn't happened yet.

PR #606, opened 2023-04-23
fix: mypy update

henryiii:
This fixes the issue with two type ignores no longer being needed, updates mypy, moves the configuration out of extras (it was outdated and not helpful to force pin mypy in extras, and not required to use mypy to type anyway, and removes the requirement that build be installed to type check (which is not needed).

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -54,7 +54,15 @@ commands_pre =
 
 [testenv:type]
 description = run type check on code base
-extras = typing
+skip_install = true
+setenv =
+    PYTHONWARNDEFAULTENCODING =
+deps =
+    importlib-metadata>=5.1
+    mypy==1.2.0
+    tomli
+    typing-extensions>=3.7.4.3
+    packaging
 commands =
     mypy
 
--- a/tox.ini
+++ b/tox.ini
@@ -54,15 +54,9 @@ commands_pre =
 
 [testenv:type]
 description = run type check on code base
-skip_install = true
+extras = typing
 setenv =
     PYTHONWARNDEFAULTENCODING =
-deps =
-    importlib-metadata>=5.1
-    mypy==1.2.0
-    tomli
-    typing-extensions>=3.7.4.3
-    packaging
 commands =
     mypy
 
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -54,7 +54,15 @@ commands_pre =
...
 description = run type check on code base
-extras = typing
+skip_install = true
+setenv =
+    PYTHONWARNDEFAULTENCODING =
+deps =
+    importlib-metadata>=5.1
+    mypy==1.2.0
+    tomli
+    typing-extensions>=3.7.4.3
```
layday, line 64: Not needed unless it's a runtime dependency. Mypy pulls typing-extensions in automatically.

tox.ini
```diff
@@ -54,7 +54,15 @@ commands_pre =
...
+    packaging
```
layday, line 65: Duplicate.

layday:
I'm not a fan of specifying dependencies in `tox.ini`, personally.  I wanna be able to `pip install -e .[all]` and be in a position to test, type check and lint the code.

henryiii:
I've reverted the env changes for now, so the CI fix can go though I don't like the current setup:

* mypy should not be required in a public extra - I'd expect `build[typing]` to install build + whatever it needs to be used in static typing, not a type checker.
* You shouldn't have to install the package when type checking it.

The example you gave doesn't work today, we don't provide an `all`? And we can't without a lot of duplication or dynamic metadata for extras...

I'd normally put the type check in pre-commit for any smaller package, but the drawbacks are the same (though bumping mypy would be easier). Honestly, the best way to do this today that would support all possible ways to run it, like in editors, would probably be a requirements.txt. And I still don't know if you'd put mypy in it - several editors have their own type checkers.

layday (approved):

layday:
> You shouldn't have to install the package when type checking it.

I'm not sure I agree.  For development, probably not.  But before publishing, I feel you should type check the built package, just as you would test against the built package.

> And we can't without a lot of duplication or dynamic metadata for extras...

We can do with self-referencing extras.  But `[all]` was just an example.

> I'd normally put the type check in pre-commit [...]

Maybe I should write a tool to parse pre-commit dependencies and install them in the current env with pip, for troglodytes comme moi to use.

PR #609, opened 2023-04-28
main: filter out malicious files when extracting tar archives

layday:
See https://peps.python.org/pep-0706/.

### DIFF
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -9,7 +9,6 @@ import platform
 import shutil
 import subprocess
 import sys
-import tarfile
 import tempfile
 import textwrap
 import traceback
@@ -230,6 +229,8 @@ def build_package_via_sdist(
     :param isolation: Isolate the build in a separate environment
     :param skip_dependency_check: Do not perform the dependency check
     """
+    from ._util import TarFile
+
     if 'sdist' in distributions:
         raise ValueError('Only binary distributions are allowed but sdist was specified')
 
@@ -240,7 +241,7 @@ def build_package_via_sdist(
     built: list[str] = []
     if distributions:
         # extract sdist
-        with tarfile.open(sdist) as t:
+        with TarFile.open(sdist) as t:
             t.extractall(sdist_out)
             try:
                 _ProjectBuilder.log(f'Building {_natural_language_list(distributions)} from sdist')
--- a/src/build/_util.py
+++ b/src/build/_util.py
@@ -2,6 +2,8 @@ from __future__ import annotations
 
 import re
 import sys
+import tarfile
+import typing
 
 from collections.abc import Iterator, Set
 
@@ -65,3 +67,18 @@ def check_dependency(
 
 def parse_wheel_filename(filename: str) -> re.Match[str] | None:
     return _WHEEL_FILENAME_REGEX.match(filename)
+
+
+if typing.TYPE_CHECKING:
+    TarFile = tarfile.TarFile
+
+else:
+    # Per https://peps.python.org/pep-0706/, the "data" filter will become
+    # the default in Python 3.14.
+    if sys.version_info < (3, 14) and hasattr(tarfile, 'data_filter'):
+
+        class TarFile(tarfile.TarFile):
+            extraction_filter = staticmethod(tarfile.data_filter)
+
+    else:
+        TarFile = tarfile.TarFile
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -87,6 +87,7 @@ source = [
 exclude_lines = [
   '\#\s*pragma: no cover',
   '^\s*raise NotImplementedError\b',
+  "if typing.TYPE_CHECKING:",
 ]
 
 [tool.coverage.paths]


src/build/_util.py
```diff
@@ -65,3 +66,14 @@ def check_dependency(
 
 def parse_wheel_filename(filename: str) -> re.Match[str] | None:
     return _WHEEL_FILENAME_REGEX.match(filename)
+
+
+# Per https://peps.python.org/pep-0706/, the "data" filter will become
+# the default in Python 3.14.
+if sys.version_info >= (3, 12) and sys.version_info < (3, 14):
```
s-t-e-v-e-n-k, line 73: From https://docs.python.org/3/library/tarfile.html:

Extraction filters were added to Python 3.12, and are backported to older versions as security updates. To check whether the feature is available, use e.g. hasattr(tarfile, 'data_filter') rather than checking the Python version.

src/build/__main__.py
```diff
@@ -239,7 +240,7 @@ def build_package_via_sdist(
     built: list[str] = []
     if distributions:
         # extract sdist
-        with tarfile.open(sdist) as t:
+        with TarFile.open(sdist) as t:
```
s-t-e-v-e-n-k, line 243: You don't need this new class.From https://docs.python.org/3/library/tarfile.html:

```
t.extraction_filter = getattr(tarfile, 'data_filter', (lambda member, path: member))
```

on the line before the `extractall()` call.

layday: Doing it that way at call site is a little unsightly and it's not reusable. I'd prefer to keep the subclass.

src/build/_util.py
layday: Thank you, updated.

src/build/_util.py
```diff
@@ -65,3 +67,18 @@ def check_dependency(
...
+if typing.TYPE_CHECKING:
+    TarFile = tarfile.TarFile
+
+else:
+    # Per https://peps.python.org/pep-0706/, the "data" filter will become
+    # the default in Python 3.14.
+    if sys.version_info < (3, 14) and hasattr(tarfile, 'data_filter'):
+
+        class TarFile(tarfile.TarFile):
+            extraction_filter = tarfile.data_filter
```
hroncok, line 81: https://peps.python.org/pep-0706/

> Applications and system integrators may wish to change extraction_filter of the TarFile class itself to set a global default. When using a function, they will generally want to wrap it in staticmethod() to prevent injection of a self argument.

```

  File "/builddir/build/BUILDROOT/python-build-0.10.0-5.fc39.x86_64/usr/lib/python3.12/site-packages/build/__main__.py", line 376, in main
    built = build_call(
            ^^^^^^^^^^^
  File "/builddir/build/BUILDROOT/python-build-0.10.0-5.fc39.x86_64/usr/lib/python3.12/site-packages/build/__main__.py", line 243, in build_package_via_sdist
    t.extractall(sdist_out)
  File "/usr/lib64/python3.12/tarfile.py", line 2253, in extractall
    tarinfo = self._get_extract_tarinfo(member, filter_function, path)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/tarfile.py", line 2305, in _get_extract_tarinfo
    tarinfo = filter_function(tarinfo, path)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: data_filter() takes 2 positional arguments but 3 were given
```

layday: Thanks, fixed.

layday: Thanks for testing it!

henryiii (approved):

src/build/_util.py
```diff
@@ -65,3 +67,18 @@ def check_dependency(
 
 def parse_wheel_filename(filename: str) -> re.Match[str] | None:
     return _WHEEL_FILENAME_REGEX.match(filename)
+
+
+if typing.TYPE_CHECKING:
```
layday, line 72:
```suggestion
if typing.TYPE_CHECKING:  # pragma: no cover
```

encukou:
Note that this is a behaviour change -- though I'd argue it's a minor one. See discussion on the `pip` issue: https://github.com/pypa/pip/issues/12111

src/build/_util.py
henryiii: Shouldn't we just disable coverage for all `if typing.TYPE_CHECKING`?

layday: Yeah, we should.

layday:
Thank you - I think build being a development tool is better positioned than pip to trial the `data` filter. If any issues arise around permission bits, we can consider switching to the `tar` filter.   The community's moving towards a direction where packages are defined statically, so the argument that "sdists involve arbitrary code execution" might hold less water now than it did a couple of years ago, and we should begin to see fewer "exotic" setups.

encukou:
FWIW, I'm proposing a PEP on this: https://discuss.python.org/t/28928

Issue #264, opened 2021-03-23
_build_in_isolated_env fails to pass in config_settings to builder.get_dependencies


e2thenegpii:
https://github.com/pypa/build/blob/2c1da830f214580e8e4b94e4c88345963fa32d2e/src/build/__main__.py#L58

When using python -m build.  The config_settings dictionary is always None in the get_requires_for_build_* callbacks regardless of the options passed on the command line.

The fix is to pass in config_settings dictionary to get_dependencies on the call on the referenced line.

I will likely have the time to add a pull request in a few days.

layday:
This is an oversight but a fortunatee one, because fixing it would break passing config settings to setuptools' sdists and wheels :(

e2thenegpii:
Would you mind elaborating how it would break setuptools? Did you run a test suite to come to that conclusion? Looking at the setuptools PEP517 backend at https://github.com/pypa/setuptools/blob/88b8b7807ec3064fec713da83fc0c7860b982536/setuptools/build_meta.py#L152 it seems setuptools is going to ignore any entry that doesn't have the key of "--global-option"

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

layday:
setuptools only accepts `--global-option`.  This is passed to every hook invoked from build, so that an option that is valid for e.g. `build_wheel` (`python setup.py bdist_wheel`) is invalid for `get_requires_for_build_wheel` (`python setup.py egg_info`).

animalize:
setuptools v64.0.0 (11 Aug 2022) added `--build-option`.
Release note: https://setuptools.pypa.io/en/latest/history.html#v64-0-0
The PR added `--build-option`:https://github.com/pypa/setuptools/pull/3380

henryiii:
That's great to hear! For other build backends (scikit-build-core, meson-python, etc), this bug will likely be problematic, and would be good to fix. For example, in scikit-build-core, every config option is settable via config settings, and some of them do affect the dependency step, like setting the min version of CMake.

Think it's time to fix this bug and note how to use it with modern setuptools?

layday:
I don't think it makes a lot of difference that setuptool accepts `--build-option` - global option just means the option comes before the command and build option comes after it, e.g. `python setup.py --global-opts-here build_sdist --build-opts-here`.  The issue where setuptools will pass all global/build opts to call commands remains.

henryiii:
[...]

PR #627, opened 2023-06-26
main: ensure config_settings are passed to get_requires_for_build

jameshilliard:
Fixes builds for packages like psycopg2(for example if the `--pg-config` config setting needs to be set) that can hard error in some cases without `config_settings` being passed to their `get_requires_for_build` hooks.

### DIFF
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -112,7 +112,7 @@ def _build_in_isolated_env(
         # first install the build dependencies
         env.install(builder.build_system_requires)
         # then get the extra required dependencies from the backend (which was installed in the call above :P)
-        env.install(builder.get_requires_for_build(distribution))
+        env.install(builder.get_requires_for_build(distribution, config_settings or {}))
         return builder.build(distribution, outdir, config_settings or {})
 
 
@@ -126,7 +126,7 @@ def _build_in_current_env(
     builder = _ProjectBuilder(srcdir)
 
     if not skip_dependency_check:
-        missing = builder.check_dependencies(distribution)
+        missing = builder.check_dependencies(distribution, config_settings or {})
         if missing:
             dependencies = ''.join('\n\t' + dep for deps in missing for dep in (deps[0], _format_dep_chain(deps[1:])) if dep)
             _cprint()
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -135,7 +135,7 @@ def test_build_isolated(mocker, package_test_flit):
 
     install.assert_any_call({'flit_core >=2,<3'})
 
-    required_cmd.assert_called_with('sdist')
+    required_cmd.assert_called_with('sdist', {})
     install.assert_any_call(['dep1', 'dep2'])
 
     build_cmd.assert_called_with('sdist', '.', {})


layday:
Please see https://github.com/pypa/build/issues/264.

jameshilliard:
> Please see #264.

I'm a bit confused what's going on there, does the issue there apply to just `--global-option` usage or to `--build-option` as well? I'm using the `--build-option` [here](https://patchwork.ozlabs.org/project/buildroot/patch/20230626181531.2312002-5-james.hilliard1@gmail.com/) in my setuptools pep517 migration patch, does that look like the right approach or are there better options available?

henryiii:
The problem is that if we fix this bug some usages of setuptools break.

IMO, I'd be happy to have this with some sort of flag that would allow it to be done correctly but have an opt-in to the old behavior to keep setuptools working, and maybe setuptools would finally fix it on their side. (CC @abravalheri)

jameshilliard:
> I'd be happy to have this with some sort of flag that would allow it to be done correctly but have an opt-in to the old behavior to keep setuptools working

Any preference for what I should name the flag?

henryiii:
`--no-config-settings-for-requires`, maybe? @layday, does that sound like a good path forward?

layday:
Given that passing config settings to setuptools' `get_requires` is a practical impossibility the path of least resistance would be for setuptools to actually exclude config settings from `get_requires` until such time that they are namespaced. This can be done by e.g. providing an alternative `get_requires` config setting-less backend which can serve as a stopgap, much like `__legacy__`.

jameshilliard:
> passing config settings to setuptools' `get_requires` is a practical impossibility the path of least resistance would be for setuptools to actually exclude config settings from `get_requires` until such time that they are namespaced

What's a good way to reproduce the issue with passing config settings to `get_requires`? It's unclear to me exactly what the issue is right now.

Did setuptools [adding the `--build-option` config](https://github.com/pypa/setuptools/pull/3380) change the situation?

layday:
Passing any option that egg_info and sdist/bdist_wheel do not share will result in an error.

Sent with [Proton Mail](https://proton.me/) secure email.

------- Original Message -------

FFY00:
I commented on https://github.com/pypa/build/issues/264#issuecomment-1609513063, but essentially I think we should proceed with this change special-casing setuptools to default to the old behavior.

FFY00:
> `--no-config-settings-for-requires`, maybe? @layday, does that sound like a good path forward?

I'd prefer to not add more config options, it makes maintenance harder (remember this is something that will end up hardcoded in build scripts for decades!) and makes it more reasonable for build backends to not have a universal `config_settings` interface across all hooks. This is something that I'd definitely be open to considering if there's a good argument for it, but so far it's just "setuptools does not support (certain?) arguments in the `config_settings` for non `build_*` hooks".

We do need a way to temporarily switch to the old behavior, but I think an environment variable would be better suited here, as it won't cause breakage when we remove support for it (after which we would issue a warning if the envvar is set for a while at least).

> Given that passing config settings to setuptools' `get_requires` is a practical impossibility the path of least resistance would be for setuptools to actually exclude config settings from `get_requires` until such time that they are namespaced.

I agree.

> This can be done by e.g. providing an alternative `get_requires` config setting-less backend which can serve as a stopgap, much like `__legacy__`.

That will still require new versions of setuptools 🫤

henryiii:
Environment variable sounds fine.

> That will still require new versions of setuptools 🫤

I think the idea was that a package could provide a custom build backend?

```toml
[build-system]
requires = ["setuptools"]
build-backend = "backend"
backend-path = ["_custom_build"]
```

```python
# _custom_build.py
import setuptools.build_meta
from setuptools.build_meta import *

def get_requires_for_build_sdist(config_settings=None):
    return setuptools.build_meta.get_requires_for_build_sdist()
# etc.
```

Though I don't think that every package that's affected should be adding a workaround, or even chaining to some new "setuptools.build_meta.__modern__" or something backend either; setuptools should be able to handle config-settings without errors for these other commands.

henryiii:
Also, since this might take a while, what about having something like `BUILD_PASS_CONFIG_SETTINGS_TO_REQUIRES`, were setting it to 0 will be the current behavior, and 1 will be the new behavior. If unset, the next version of build will default to 0, while a future version of build will change the default to 1. That way, we can ship a workaround for this now without breaking lots of setuptools projects, and in the future, once there's a released version of setuptools that won't break when this is set, we can make the change so then projects with old setuptools have to opt-into the old workaround.

Otherwise, I don't see a good way to allow non-setuptools backends to work correctly quickly.

layday:
I don't think an env var is much better than a flag.  Making something less discoverable doesn't mean we can make breaking changes to it more easily.  A `BUILD_PASS_CONFIG_SETTINGS_TO_REQUIRES` flag is still something we'll probably want to deprecate before removal; worse, we'll need to implement custom error handling post-removal instead of having argparse error on being passed an unknown option.  I don't like either option personally.  I think we should:

* Delay merging this PR and work with the setuptools folks to provide an alternative (be it a new backend or a new way to pass config settings)
* *Or* allow passing different config settings per hook from build - we can discuss how this is gonna look

henryiii (approved):

henryiii:
I’d like two more approvals on this before merging of possible. Ideally @layday and @FFY00.

layday (approved):

layday:
I'm convinced by the findings in #264 that this is the best way forward.

PR #631, opened 2023-06-27
Bump importlib metadata dependency

### LINKED ISSUE #630: InvalidRequirement on Python 3.8
In pypa/setuptools#3965, I discovered that `build` fails to build a dependency with a url req that's part of extras because it stumbles on python/importlib_metadata#357, which was only backported to Python 3.9, but [build uses stdlib importlib.metadata for Python 3.8](https://github.com/pypa/build/blob/cd06da25481b9a610f846fa60cb67b5a5fa9a051/src/build/__init__.py#L150-L153). Build should probably use `importlib_metadata` for Python 3.8 and maybe for Python 3.9 ([quite a few improvements and fixes](https://importlib-metadata.readthedocs.io/en/latest/history.html#v4-3-0) landed in CPython 3.10).

jaraco:
- Consolidate importlib metadata import
- Rely on importlib_metadata for Python 3.9 and earlier. Fixes #630.


layday:
You'll also need to update the min version in https://github.com/pypa/build/blob/main/tests/constraints.txt.  Technically, the PR looks solid - we should've consolidated the metadata import in a separate module long ago :)  I don't know if we wanna add a fallback for packagers however.

henryiii:
FYI, personally, I have started putting all backports into `_compat`, with the same folder structure as the stdlib. So this would be `_compat/importlib.py` or `_compat/importlib/metadata.py`. Don't have to follow that here, I just find it works well, scales to multiple backports, and is easy to remove when the backport period is over.

> I don't know if we wanna add a fallback for packagers however.

Guessing you mean to make bootstrapping, etc easier? So we'd have a try/except around the <3.10 branch?

src/build/_importlib.py
```diff
@@ -0,0 +1,10 @@
+import sys
+
+
+if sys.version_info >= (3, 10):
+    from importlib import metadata
+else:
+    import importlib_metadata as metadata
```
henryiii, line 7:
```suggestion
    try:
        import importlib_metadata as metadata
    except ModuleNotFoundError:
        from importlib import metadata
```

I think this is what was suggested? (Should we have a warning for this fallback branch?)

jaraco:
> FYI, personally, I have started putting all backports into `_compat`, with the same folder structure as the stdlib. So this would be `_compat/importlib.py` or `_compat/importlib/metadata.py`. Don't have to follow that here, I just find it works well, scales to multiple backports, and is easy to remove when the backport period is over.

I often do something similar, and I go a step further and group the compat modules by Python version, so `py38compat` supplies compatibility interfaces until Python 3.8 support is dropped ([example](https://github.com/python/importlib_metadata/blob/main/importlib_metadata/_py39compat.py)). That makes it really easy to locate and remove the relevant pieces when they're no longer needed.

jaraco:
> You'll also need to update the min version in https://github.com/pypa/build/blob/main/tests/constraints.txt.

Done in 72537a7.

src/build/_importlib.py
jaraco: I'm not sure that's what's suggested. Agreed there should be a warning if that branch is reached (as it could mask failures in the `-min` tests). I'd argue there should not be a fallback. That does mean that downstream packagers will be forced to install `importlib_metadata` on later Pythons where they did not need to before, but that's what this change is about. If downstream packagers want to bypass the fix, they should do so in a patch (IMHO).

henryiii: Couldn't just add a workaround for this issue, instead of changing the requirements? Looks to me like you could very simply process this? That would keep us from adding just a massive dependency stack for Python 3.8 & 3.9. We could still keep the attempt to use `importlib_metadata` if Python < 3.10, but not change our dependencies.

jaraco: > Couldn't we just add a workaround for this issue, instead of changing the requirements? Looks to me like you could very simply process this requirement before handing it off to packaging? That would keep us from adding just a massive dependency stack for Python 3.8 & 3.9. We could still keep the attempt to use `importlib_metadata` if Python < 3.10, but not change our (required) dependencies.

Yes, that seems plausible. It does mean that other bugs may later be discovered and requiring revisiting only to discover that they've already been fixed, just not incorporated. The risk is _probably_ small, and I would have said negligible prior to stumbling onto this issue.

In general, my philosophy is, where possible, to drive toward a single implementation at HEAD (i.e. latest, best version of code is available as `importlib_metadata`), rather than maintaining lots of forks each of which need their own intervention (i.e. Python 3.8 importlib.metadata + bug wrappers). I know that's not always possible and the packaging challenges above illustrate one such consequence.

It's a shame the bootstrapping issue is still such an impediment. I'd hoped the PEP 517+518 system would have unblocked the ability for packagers to reliably and mechanically build with any number of build and install dependencies without substantial maintenance cost. It's clear that the platform packagers haven't realized this goal. (and to be sure, I don't have a better solution)

I'd be okay if someone wanted to devise an alternate workaround wrapper for the reported issue. I'd be somewhat reluctant to own that effort, as I've already fixed the bug twice (once in importlib_metadata and again in CPython versions accepting the fix).

henryiii: > It's a shame the bootstrapping issue is still such an impediment. I'd hoped the PEP 517+518 system would have unblocked the ability for packagers to reliably and mechanically build with any number of build and install dependencies without substantial maintenance cost.

It absolutely has, but the key enablers of this was not "PEP 517/518" in general, but instead a set of tools built on it - flit-core (which has 0 dependencies), build, and installer. These packages have a clear bootstrapping story, and you can then use them to support everything else, including setuptools, poetry, or whatever. So being a dependency of one of these three packages means you need to part of that story. This has worked well - for example, `wheel` ([in a PR](https://github.com/conda-forge/wheel-feedstock/pull/54)) can now build on conda-forge without `setup.py` (current method) or `pip` (which depends on it in conda-forge!), but instead using flit-core and installer.

But it does mean these packages are special, as they need to have minimal dependencies, and if there are any, dependencies ideally need to be built with flit-core (especially for 3.12+, but it would be nice to have a modern, simple story for all supported Pythons, instead of reverting to a pre-included distutils/setputools and setup.py.)

layday: I'd be happy to merge this with the fallback personally.

pyproject.toml
```diff
@@ -36,7 +36,7 @@ dependencies = [
   "pyproject_hooks",
   # not actually a runtime dependency, only supplied as there is not "recommended dependency" support
   'colorama; os_name == "nt"',
-  'importlib-metadata >= 0.22; python_version < "3.8"',
+  'importlib-metadata >= 4.6; python_version < "3.10"',
```
henryiii, line 39:
```suggestion
  'importlib-metadata >= 4.6; python_version < "3.10"',  # Not required for 3.8+, but fixes a bug
```

src/build/_importlib.py
```diff
@@ -4,7 +4,10 @@
 if sys.version_info >= (3, 10):
     from importlib import metadata
 else:
-    import importlib_metadata as metadata
+    try:
```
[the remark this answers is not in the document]
jaraco: Your call. I try to keep it simple by assuming the requirements have been met. It feels overly defensive to be validating the packaging on every import. The implication is that every package should be validating every package on every import, which feels a little like madness.

layday (approved):

PR #632, opened 2023-07-01
Allow passing config settings on individual hook basis

layday:
Alternative to the flag/env var being discussed to accommodate setuptools.  This would enable uses such as
`-Cget_requires_for_build_wheel:--tag-build=123 -Cbuild_wheel:--compression=stored` possible.

Whether per-hook configuration is desirable remains to be decided, but I thought it was an idea worth exploring.

### DIFF
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -17,13 +17,23 @@ import warnings
 
 from collections.abc import Iterator, Sequence
 from functools import partial
-from typing import NoReturn, TextIO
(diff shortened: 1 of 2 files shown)


layday:
Not able to reliably catch the invalid hook warning because of: https://github.com/pypa/build/blob/7e0e79884888b7c1cdaf719915d31d9597412bf8/src/build/__main__.py#L70.

PR #636, opened 2023-07-03
main: avoid cost of importing virtualenv if not using it

### LINKED ISSUE #510: Don't import `virtualenv` if it is not actually needed.
Importing `virtualenv` causes `setuptools` plugins being executed. So it causes them being initialized and executed multiple times, which is slow. I know that that flaw is not in the `build`, but importing the unneeded (I always use `-n` flag) `virtualenv` and so wasting time and memory on its loading is a flaw by itself.

It is proposed to move importing of `virtualenv` into the functions actually using it.

henryiii:
Fix #510. If a user is using `-n`, skip the cost of importing virtualenv.

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import abc
 import functools
+import importlib.util
 import logging
 import os
 import platform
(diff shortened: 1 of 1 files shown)


layday (approved):

KOLANICH:
Thanks.

PR #657, opened 2023-08-21
fix: include all files in SDist

### LINKED ISSUE #656: the PyPi package does not contain tests directory
As title says, the [package released on PyPi](https://files.pythonhosted.org/packages/de/1c/fb62f81952f0e74c3fbf411261d1adbdd2d615c89a24b42d0fe44eb4bcf3/build-0.10.0.tar.gz) does not contain the `tests` dir, which could be useful to test the package when distributing it.

henryiii:
Building with build isn't the same as building with flit, due to flit-core not behaving the same when using standardized tools. :(

This adds all the files in the git repo, just like building with flit would. I've used check-sdist to do so, but haven't added it as a pre-commit check (yet).

Closes #656.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["flit-core >= 3.4"]
+requires = ["flit-core >= 3.8"]
 build-backend = "flit_core.buildapi"
 
 [project]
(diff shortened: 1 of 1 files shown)


henryiii:
As I expected, the readthedocs build is broken by https://github.com/pradyunsg/furo/discussions/708.

pyproject.toml
```diff
@@ -77,6 +77,11 @@ pyproject-build = "build.__main__:entrypoint"
 [project.entry-points."pipx.run"]
 build = "build.__main__:entrypoint"
 
+[tool.flit.sdist]
+include = ["tests/", ".gitignore", "CHANGELOG.rst", "docs/", ".dockerignore", "tox.ini"]
+exclude = ["tests/__pycache__", "docs/_build"]
```
layday, line 82: Should be able to use `"**/__pycache__"` now.

henryiii: Okay, did require a bump to flit-core 3.8+.

gaborbernat (approved):

PR #662, opened 2023-08-22
chore: add some exception handling minor cleanup checks

henryiii:
Some cleanup based on Ruff checks.

### DIFF
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -68,11 +68,13 @@ def _find_typo(dictionary: Mapping[str, str], expected: str) -> None:
 
 def _validate_source_directory(source_dir: PathType) -> None:
     if not os.path.isdir(source_dir):
-        raise BuildException(f'Source {source_dir} is not a directory')
+        msg = f'Source {source_dir} is not a directory'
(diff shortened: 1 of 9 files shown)


src/build/__init__.py
```diff
@@ -340,17 +351,15 @@ def _handle_backend(self, hook: str) -> Iterator[None]:
...
-            )
+            ) from None
         except subprocess.CalledProcessError as exception:
-            raise BuildBackendException(  # noqa: B904 # use raise from
-                exception, f'Backend subprocess exited when trying to invoke {hook}'
-            )
+            raise BuildBackendException(exception, f'Backend subprocess exited when trying to invoke {hook}') from None
         except Exception as exception:
-            raise BuildBackendException(exception, exc_info=sys.exc_info())  # noqa: B904 # use raise from
+            raise BuildBackendException(exception, exc_info=sys.exc_info()) from None
```
henryiii, line 362: I think these were holdovers from Python 2 days. We are propagating the info anyway, so I think `from None` is correct here.

gaborbernat (approved):

PR #663, opened 2023-08-22
chore: add RET to Ruff checks

henryiii:
This adds the RET checks to Ruff.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -38,7 +38,7 @@ repos:
     rev: v0.0.285
     hooks:
       - id: ruff
-        args: [--fix, --format, grouped, --show-fixes]
+        args: [--fix, --show-fixes]
   - repo: https://github.com/codespell-project/codespell
     rev: "v2.2.5"
     hooks:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -159,6 +159,7 @@ select = [
   "UP",   # pyupgrade
   "W",    # pycodestyle
   "YTT",  # flake8-2020
+  "RET",  # flake8-return
 ]
 src = ["src"]
 
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -100,7 +100,7 @@ def _parse_build_system_table(pyproject_toml: Mapping[str, Any]) -> Mapping[str,
     if 'requires' not in build_system_table:
         _find_typo(build_system_table, 'requires')
         raise BuildSystemTableValidationError('`requires` is a required property')
-    elif not isinstance(build_system_table['requires'], list) or not all(
+    if not isinstance(build_system_table['requires'], list) or not all(
         isinstance(i, str) for i in build_system_table['requires']
     ):
         raise BuildSystemTableValidationError('`requires` must be an array of strings')
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -42,7 +42,7 @@ def _init_colors() -> dict[str, str]:
         if 'FORCE_COLOR' in os.environ:
             warnings.warn('Both NO_COLOR and FORCE_COLOR environment variables are set, disabling color', stacklevel=2)
         return _NO_COLORS
-    elif 'FORCE_COLOR' in os.environ or sys.stdout.isatty():
+    if 'FORCE_COLOR' in os.environ or sys.stdout.isatty():
         return _COLORS
     return _NO_COLORS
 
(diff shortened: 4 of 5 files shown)


.pre-commit-config.yaml
```diff
@@ -38,7 +38,7 @@ repos:
     rev: v0.0.285
     hooks:
       - id: ruff
-        args: [--fix, --format, grouped, --show-fixes]
+        args: [--fix, --show-fixes]
```
henryiii, line 41: Oops, didn't mean to combine this into this PR. Though I'm a fan of removing this, the "grouped" form doesn't quick-link to the exact location in the source in terminals (like iterm) or IDEs (like VSCode and PyCharm). It's also not possible to process it into annotations (though GitHub currently has a bug keeping annotations and color output from both working together, see https://github.com/actions/runner/pull/2430).

layday:
Personally, I like using "else" where it might indicate that the "else" condition is part of a set.  You can either build with isolation, or without isolation. Here, the "else" is not for fallback; it's simply how we chose to map out the set given the grammatical constraints of the language.

henryiii:
I know some people don't like the Ruff/flake8-return/PyLint checks for this (the Rubocop one doesn't ever seem to be debated, though), so if everyone disagrees, I don't have to add it. But let me try to convince you (and everyone else) that this is a good idea:

The problem is that the `elif`/`else` after a control statement (`raise`, `return`, `break`, or `continue`) is completely unnecessary both at runtime and in type checking, and has no effect. That means adding it:

1. Is completely arbitrary. You might try to communicate intent via this, but from looking at lots of these, I've never consistently seen it applied or used in a way that added anything to the code. Instead, it provides two ways to write the same thing, which is the point of Black, isort, and other checks - the same code should be written the same way. Several of the ones above are a mix of elses and guards.
1. It is trying to convey intent only to the reader, and has no effect on the program. Just like commented out code, or excessive comments. Only it doesn't have a comment symbol to immediately tell the reader it's meaningless.
1. It adds useless indentation and structure, and it doesn't nest well at all. Even just two levels read _much_ better without the else's. I think one of the main issues people used to have with Python's whitespace is due to using too many else's.
1. Else blocks are already ugly - they require a reader to keep context of where the if started. The "guard pattern" lets you read top-to-bottom without having to track beginnings of long blocks (for some reason this pattern is more commonly talked about in Ruby, maybe because Rubocop can rewrite code using it and Ruby has a language feature that lets you write a guard in one line).
1. Everyone who reads Python already has to be able to read it the else-less way (and in fact it's less complicated due to less structure).
1. No one would ever ask someone to _add_ meaningless else's in a code review if it had been written without them.
1. Just like Black, you can always refactor code to write it the way you want, in this case included else's by pulling it out and assigning it to variable or similar. Or you can disable the check for that line.
1. Match statements are often a better choice if you really want an else chain.
1. The final "else" is the "nothing above matched" case, and deserves to be special and unintended, to help the reader understand it's the "fallback" and something will always be hit in the chain. Especially noticeable when nested!

For example, compare this code:

```python
if a:
    if b:
       return 1
    else:
       return 2
else:
   return 3
```

To see that it always returns, you have to match up every else to every if, and make sure there are no passes that miss. If you removed the inner `else: return 2`, there would be a way to fall through and return None, and it's not obvious unless you look at the entire structure. Now compare this:

```python
if a:
    if b:
       return 1
    return 2
return 3
```

It's immediate obvious that every pass returns. If you remove the `return 2` line, it still doesn't `accidentally` fall through, since the `return 3` line is still hit.  A reader only has to see that final line and its _level of indentation_ to immediately be able to tell there's always a return; they don't need to inspect the if's above at all.


For an example in this PR, look at `src/build/__main__.py`. That had a guard for `element==0`, and then an if/else chain for `element==1`. It reads much better and more consistent with less indentation by making both guards.

henryiii:
Ahh, missed that you pointed one out. I think both read fine - `if isolation` do this. Otherwise do that. It's three lines instead of four, and I don't think it would confuse anyone reading it. Ideally, within the constructs of the language, the best way to write it would be this:

```python
return _build_in_isolated_env(srcdir, outdir, distribution, config_settings) if isolation else _build_in_current_env(srcdir, outdir, distribution, config_settings, skip_dependency_check)
```

Or, when formatted,

```python
return (
    _build_in_isolated_env(srcdir, outdir, distribution, config_settings)
    if isolation
    else _build_in_current_env(
        srcdir, outdir, distribution, config_settings, skip_dependency_check
    )
)
```

But this syntax doesn't handle long expressions very well and the Black formatting for it is not great when the expressions are so long.

gaborbernat:
Seems it stalled, @henryiii let me know if you want to pick it up again and we can reopen.

PR #664, opened 2023-08-22
docs: update changelog for 0.11.0

henryiii:
Starting to get ready.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,15 +11,42 @@ Unreleased
   (PR :pr:`567`)
 - Added ``runner`` parameter to ``util.project_wheel_metadata``
   (PR :pr:`566`, Fixes :issue:`553`)
-- Modified ``ProjectBuilder`` constructor signature,
-  added alternative ``ProjectBuilder.from_env`` constructor,
-  redefined ``env.IsolatedEnv`` interface, and exposed ``env.DefaultIsolatedEnv``,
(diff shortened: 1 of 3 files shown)


CHANGELOG.rst
```diff
@@ -11,15 +11,42 @@ Unreleased
   (PR :pr:`567`)
 - Added ``runner`` parameter to ``util.project_wheel_metadata``
   (PR :pr:`566`, Fixes :issue:`553`)
```
layday, line 13: For consistency:

```suggestion
  (PR :pr:`566`, fixes issue :issue:`553`)
```

layday (approved):

pyproject.toml
```diff
@@ -122,7 +122,6 @@ filterwarnings = [
   "error",
   "ignore:path is deprecated.:DeprecationWarning",
   "ignore:The --rsyncdir command line argument and rsyncdirs config variable are deprecated.:DeprecationWarning",
-  "ignore:Python 3.14 will, by default, filter extracted tar archives:DeprecationWarning",
```
henryiii, line 125: I think flit is triggering this? Pretty sure it's not us. I might try only ignoring it for Flit, if that doesn't work, I'll just restore it.

henryiii: Ahh, it's actually in our tests. Okay, will put it back for now. We maybe can silence this in our tests in the future.

gaborbernat (approved):
We should make it 1.0.0 instead of 0.11.0.

Issue #651, opened 2023-08-09
--extra-index-url


SerGeRybakov:
Is it possible to build source distributions that depend on at least one package, which is not in public PyPI? 

For example, I have a requirements.txt

```
numpy
requests

--extra-index-url https://test.pypi.org/simple/
some-package
```

When I'm doing `python -m build --sdist` I get
```
setuptools.extern.packaging.requirements.InvalidRequirement: Expected package name at the start of dependency specifier
    --extra-index-url https://test.pypi.org/simple/
    ^
```
I tried to set pip-args in envs and pass it directly as `--extra-index-url=https://test.pypi.org/simple/ --trusted-host=https://test.pypi.org/simple/` but no result

Is there any way to do it?

gl-yziquel:
You do not need such a flag. I build python code with locally deployed dependencies. The trick is to having correct configurations globally. For instance, in the `~/.pypirc` file, you can have declarations like:

```console
[distutils]
index-servers =
    local

[local]
repository = http://127.0.0.1:3141/root/local/
```

This is kind of the default setup you'd get with a locally installed `devpi` server.

Adapt to your specific personal needs.

agriyakhetarpal:
As a workaround, it should be possible to install (into a virtual environment) the private-indexed dependency first, and then run `python -m build` (inside the virtual environment) with the `--no-isolation` command-line flag (while pre-installing any other dependencies required at build-time).

i.e., this might work:

```bash
python3.X -m virtualenv venv
source venv/bin/activate
pip install cmake dependencyA dependencyB  # and so on
pip install private_dependency --extra-index-url "https://path.to.your/pvt/index/"
pip install build
python -m build --sdist --no-isolation --outdir PATH
```

you might need to make adjustments if you have a custom `build_sdist` command in `setup.py` (assuming that you're using `setuptools` as a build-backend). 

P.S. It looks like PDM and Poetry directly support adding extra index URLs in `pyproject.toml`, according to [this discussion](https://discuss.python.org/t/how-to-specify-extra-index-in-a-pyproject-toml-for-pip-and-pip-tools/23592/7)), but [`setuptools` has removed it](https://github.com/pypa/setuptools/discussions/3860) in the past.

<hr>

P.P.S. I tested this locally with the [nightly wheels for SciPy](https://anaconda.org/scientific-python-nightly-wheels/scipy):

This is what I modified my build-system table to look like:

```toml
[build-system]
requires = [
  "setuptools>=64",
  "pooch",
[...]

PR #665, opened 2023-08-24
env: validate PEP 508 specifiers before passing them as arguments to pip

eli-schwartz:
Per the specification for pyproject.toml build-system.requires, these MUST be valid PEP 508 strings, and when checking to see if they are satisfied during --no-isolation builds, they will be parsed using `packaging.requirements` which checks this for free as a side effect of converting each string into a good format for checking against importlib.metadata dists.

But when building by default, they are just pip installed and pip supports many things that aren't just PEP 508 strings. Fix this by first parsing them with `packaging.requirements` just to get the error message.


eli-schwartz:
What should the preferred UX of this be? Previously, this emitted:
```
ERROR Command '['/tmp/build-env-3atlhb3x/bin/python', '-Im', 'pip', 'install', '--use-pep517', '--no-warn-script-location', '-r', '/tmp/build-reqs-a3ytyjeu.txt']' returned non-zero exit status 1.
```

with a preceding traceback and a pip error:
```
ERROR: Invalid requirement: 'xxx' (from line 1 of /tmp/build-reqs-a3ytyjeu.txt)
```

With my change, it errors out with a packaging.requirements.Requirement exception and logs:
```
ERROR Expected package name at the start of dependency specifier
    _vendored/meson
    ^
```

or
```
ERROR Expected end or semicolon (after name and no valid version specifier)
    this is invalid
         ^
```

The tests are doing some finicky position-based output matching, trying to figure out what it even expects...

henryiii:
The position based matching is annoying, you can just update it to whatever the new line numbers in the source are.

> pip supports many things that aren't just PEP 508 strings.

Will this break any current workarounds? (I'm thinking of #651, which apparently doesn't work now, but just wondering if there's anything like that to worry about). If so, maybe a warning first then an error later would be better? (Even though we say non-PEP 508 strings are unsupported, someone might not know what they are doing is unsupported, so a phase out period is friendlier in that case). But if not, then no need.

> What should the preferred UX of this be? 

The above seems fine to me? Though maybe printing out where this occurs (`build-system.requires` in `pyproject.toml` would be helpful.

henryiii:
> Will this break any current workarounds?

From the linked issue:

> But if you build with isolation disabled then it does check that all dependencies are installed

So this is expanding the "with `-n` but without `-x`" behavior to all runs.

I think this is fine, but what about only running this if `-x` is not passed? That way, someone could disable this check by adding `-x` if they do need the undefined behavior. That would mean the default is now safe, but someone could work around this if they know what they are doing and need to build a package with an invalid spec here.

layday:
`-x` is for (not) checking whether all declared dependencies are installed.  The fact that it parses requirement strings to do that is incidental.  IMO if we want to check that requirement strings are well-formed we should do that universally, in the builder and not the env, so that the same guarantee will extend to `-nx` builds as well as custom envs.

henryiii:
> IMO if we want to check that requirement strings are well-formed we should do that universally

IMO, long term, validating requirement strings is probably fine, but I'm thinking in a backward compatible, if someone today puts "subdirectory/project" as a requirement string, they can build their project out of the box today, as long as they don't pass `-n`. I think the best way forward is to enable requirement checking by default, but only if `-x` is not passed. That way people using "subdirectory/project" will be alerted that what they are doing is not supported, but they have an emergency workaround (pass `-x`) while they redesign what they are doing.

So the suggestion is:

| `-n` | `-x` | Today    | Suggested | Someday |
|------|------|----------|-----------|---------|
| ❌   | ❌  | Unchecked | Checked   | Checked |
| ❌   | ✅  | Unchecked | Unchecked | Checked |
| ✅   | ❌  | Checked   | Checked   | Checked |
| ✅   | ✅  | Unchecked | Unchecked | Checked |

gaborbernat:
Seems it stalled, @eli-schwartz  let me know if you want to pick it up again and we can reopen.

PR #670, opened 2023-09-02
[feat]: Added Support for --overwrite argument

### LINKED ISSUE #441: [feature] Add sub command or flag to overwrite directory dist
For now, we can could python package by running the command `python -m build`. and both `tar.gz` and `whl` files would be created and location it to directory `dist`. but when I change versions of the package and re-run the command `python -m build`, the old package still exists.

could we add a flag `--overwirte, -o` to build, if user add this and we would clean the distribute dir first before we output our build result

Mr-Sunglasses:
Added Support for --overwrite argument which cleans the old build files.

fix #441 

**Note**  - This PR is currently a Draft because there are some doubts related to this issue when the discussion will be resolved then, it'll remove it from draft.

### DIFF
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -319,6 +319,11 @@ def main_parser() -> argparse.ArgumentParser:
         metavar='PATH',
     )
     parser.add_argument(
+        '--overwrite',
+        '-ow',
+        help=f'clear the dir'
+    )
+    parser.add_argument(
         '--skip-dependency-check',
         '-x',
         action='store_true',
@@ -377,6 +382,14 @@ def main(cli_args: Sequence[str], prog: str | None = None) -> None:
 
     # outdir is relative to srcdir only if omitted.
     outdir = os.path.join(args.srcdir, 'dist') if args.outdir is None else args.outdir
+
+    if args.overwrite:
+        directory_path = "dist"
+        for filename in os.listdir(directory_path):
+            file_path = os.path.join(directory_path, filename)
+            if os.path.isfile(file_path):
+                os.remove(file_path)
+
 
     if distributions:
         build_call = build_package
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -40,7 +40,10 @@ _NO_COLORS = {color: '' for color in _COLORS}
 def _init_colors() -> dict[str, str]:
     if 'NO_COLOR' in os.environ:
         if 'FORCE_COLOR' in os.environ:
-            warnings.warn('Both NO_COLOR and FORCE_COLOR environment variables are set, disabling color', stacklevel=2)
+            warnings.warn(
+                'Both NO_COLOR and FORCE_COLOR environment variables are set, disabling color',
+                stacklevel=2,
+            )
         return _NO_COLORS
     elif 'FORCE_COLOR' in os.environ or sys.stdout.isatty():
         return _COLORS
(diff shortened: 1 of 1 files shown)


henryiii:
This isn't "overwrite" - build already does that. This is "delete everything in the directly you are targeting before building", which is dangerous - if you targetted your home dir and added `--overwrite`, then it wipes everything.

If someone wants to delete things in the output folder, you can use a second command. You don't need to make one command do everything. Especially when that command is `rm *` (or `rm *.tar.gz *.whl`). You can also make a new build dir by pointing at a new output dir.

(That's my opinion, anyway, unless I see a really good argument for why `rm dist/* && pipx run build` isn't a better, more explicit option)

gaborbernat:
I agree with @henryiii here, so we'll not move ahead with this.

gaborbernat:
I'd be alright with this, `rm` is not cross platform, so might not be the same. However, I want to keep the flag as explicit as possible. Instead of `--overwrite` it should be `--delete-output-dir-before-build` 👍

src/build/__main__.py
```diff
@@ -378,6 +383,14 @@ def main(cli_args: Sequence[str], prog: str | None = None) -> None:
     # outdir is relative to srcdir only if omitted.
     outdir = os.path.join(args.srcdir, 'dist') if args.outdir is None else args.outdir
 
+    if args.overwrite:
+        directory_path = "dist"
```
q0w, line 387: should it not be args.outdir?

henryiii:
Build is not usually used cross-platform in this way. SDists only need to be made once. If wheels depend on the platform, it's pretty common you'd need cibuildwheel to handle making redistributable wheels (which itself can use build). So a very minor cross-platform benefit to adding a "rm *" replacement I don't think it worth the potential issues.

If someone had a clear use case where having a cross platform delete everything option would really simplify something (without risking bad side effects), that might sway me, but I don't see one. At the moment it seems like it's fixing something that could also be fixed by `rm`'ing (or `del`'ing) files first.

gaborbernat:
> someone had a clear use case where having a cross platform delete everything option would really simplify something (without risking bad side effects), that might sway me, but I don't see one. At the moment it seems like it's fixing something that could also be fixed by `rm`'ing (or `del`'ing) files first.

Consider https://github.com/tox-dev/pyproject-api/blob/main/.github/workflows/release.yml#L1 . Currently in that config there's nothing that requires me to run on ubuntu, I could change the `runs-on: ubuntu-latest` to `runs-on: windows-latest` and would work just the same. If I start adding there rm that's no longer the case. I could use `python -c 'import shutil; shutil.rmtree` to keep it cross platform but this flag could save that headache. Furthermore would make tutorials much simpler, because the author would not have to change the cleanup command based on the platform to ensure consistent behavior.

henryiii:
> there's nothing that requires me to run on ubuntu

pypa/gh-action-pypi-publish is linux only. So it's already not cross-platform. And changing rm to del isn't really an issue, the issue is if you put this in a matrix, and didn't want to use if statements or set the shell (GHA supports powershell and bash on all platforms).

Also `rm` does work in powershell.

> Furthermore would make tutorials much simpler

CI builds fresh every time. You don't need this flag in CI. It would only be a local issue.

> If I start adding there rm that's no longer the case.

I thought about that, but I think I would not put a long, dangerous option like this in a tutorial. If you really had to remove things, it's much easier to understand what "rm -r dist" does without looking up some package's docs.

layday:
We don't have consensus for this change so I'm gonna close it. Thanks for taking the time to work on it.

Mr-Sunglasses:
Understood, and thanks for letting me know. I appreciate the opportunity to have contributed and am looking forward to any future chances to collaborate.

> We don't have consensus for this change so I'm gonna close it. Thanks for taking the time to work on it.

PR #676, opened 2023-09-06
chore: add an easy way to bump the version

henryiii:
This seems to be the replacement for the unmaintained bump2version, which is a fork of the unmaintained bumpversion...

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -111,7 +111,7 @@ depends =
 description = bump versions, pass major/minor/patch
 skip_install = true
 deps =
-    bump-my-version
+    bump-my-version >=0.10
 set_env =
 commands =
     bump-my-version bump {posargs}
--- a/tox.ini
+++ b/tox.ini
@@ -106,3 +106,12 @@ commands =
 depends =
     path
     {py312, py311, py310, py39, py38, py37, pypy39, pypy38, pypy37}{, -min}
+
+[testenv:bump]
+description = bump versions, pass major/minor/patch
+skip_install = true
+deps =
+    bump-my-version
+set_env =
+commands =
+    bump-my-version bump {posargs}
--- a/tox.ini
+++ b/tox.ini
@@ -111,7 +111,7 @@ depends =
 description = bump versions, pass major/minor/patch
 skip_install = true
 deps =
-    bump-my-version >=0.10
+    bump-my-version>=0.10
 set_env =
 commands =
     bump-my-version bump {posargs}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -175,3 +175,14 @@ lines-after-imports = 2
 ignore = [
   "W005",  # We _are_ build
 ]
+
+[tool.bumpversion]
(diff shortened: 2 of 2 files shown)


tox.ini
```diff
@@ -104,3 +104,11 @@ commands =
     coverage html -d {toxworkdir}/htmlcov -i
     python -m diff_cover.diff_cover_tool --compare-branch {env:DIFF_AGAINST:origin/main} {toxworkdir}/coverage.xml
 depends = {py312, py311, py310, py39, py38, py37, pypy37, pypy38, pypy39}{,-min}, path
+
+[testenv:bump]
+description = bump versions, pass major/minor/patch
+skip_install = true
+setenv =
```
gaborbernat, line 111: set_env

tox.ini
```diff
@@ -104,3 +104,11 @@ commands =
...
+deps = bump-my-version
```
gaborbernat, line 112: Add lower bound version spec

gaborbernat (changes requested):

tox.ini
henryiii: :facepalm: Ahh, I should have noticed that.

tox.ini
henryiii: Okay with something recent? Given it's internal, I'd assume yes. (0.10 came out yesterday I think)

tox.ini
henryiii: Ahh, I was copying the examples above, which also use the older format. Maybe we should move to the newer more readable forms all at once?

gaborbernat: Use https://github.com/tox-dev/tox-ini-fmt that auto upgrades

henryiii: That didn't add the underscores, all fields are the same after running it. It just moved things around and added some indentation. And broke this:

```diff
-depends = {py312, py311, py310, py39, py38, py37, pypy37, pypy38, pypy39}{,-min}, path
+depends =
+    -min}
+    path
+    py310
+    py311
+    py37
+    py38
+    py39
+    pypy37
+    pypy38
+    pypy39}{
+    {py312
```

gaborbernat: image

Are you using the latest?

gaborbernat: Fixed https://github.com/tox-dev/tox-ini-fmt/commit/69ba5f57bfbd234f3de4d15275e8122526b4d80f 👍

tox.ini
```diff
@@ -106,3 +106,12 @@ commands =
 depends =
     path
     {py312, py311, py310, py39, py38, py37, pypy39, pypy38, pypy37}{, -min}
+
+[testenv:bump]
+description = bump versions, pass major/minor/patch
+skip_install = true
+deps =
+    bump-my-version
```
henryiii, line 114:
```suggestion
    bump-my-version >=0.10
```

henryiii:
pre-commit.ci autofix

gaborbernat (approved):

PR #687, opened 2023-10-01
Build in pip-less environments

sbidoul:
Hi! Here is a little idea to improve performance and simplify a couple of things.

Since 22.3, pip has as `--python` option which lets it manage a different environment than the one it is installed in.

So if `build` depends on `pip` it can control the minimum pip version that is installed with it.
It can then create build environments without pip for better performance, and use the pip version from its own environment to manage them.

What do you think?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,6 +34,7 @@ urls.changelog = "https://pypa-build.readthedocs.io/en/stable/changelog.html"
 dependencies = [
   "packaging >= 19.0",
   "pyproject_hooks",
+  "pip>=22.3",
   # not actually a runtime dependency, only supplied as there is not "recommended dependency" support
   'colorama; os_name == "nt"',
   'importlib-metadata >= 4.6; python_version < "3.10"',  # Not required for 3.8+, but fixes a stdlib bug
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -5,7 +5,6 @@ import functools
 import importlib.util
 import logging
 import os
-import platform
 import shutil
 import subprocess
 import sys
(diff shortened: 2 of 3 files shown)


layday:
I agree that this is a much cleaner approach, but because build is used to bootstrap packaging in Python, we can only take on a small number of dependencies which are built directly with flit, which are then used to build build, which is used to build everything else, including pip.

henryiii:
Yes, but this could be an option, perhaps?

pyproject.toml
```diff
@@ -34,6 +34,7 @@ urls.changelog = "https://pypa-build.readthedocs.io/en/stable/changelog.html"
 dependencies = [
   "packaging >= 19.0",
   "pyproject_hooks",
+  "pip>=22.3",
```
uranusjr, line 37:
```suggestion
  "pip>=22.3",
  "pyproject_hooks",
```

Just keeping things roughtly alphabetically sorted

sbidoul:
Would it make sense for a bootstrap process to download the pip sdist and add add `pip`'s `src` directory to `build`'s `PYTHONPATH`?

sbidoul:
> Why not use the pyz distribution?

Also, but is that acceptable for folks bootstrapping from source?

layday:
I don't imagine that initiating out-of-band network requests during the bootstrap/build process would be acceptable to redistributors.

sbidoul:
@layday sorry, I was unclear, let me rephrase that. Would it make sense for a bootstrap process to add `pip`'s src directory to  `build`'s PYTHONPATH? I'm looking for a way to bootstrap `build` without building pip in response to https://github.com/pypa/build/pull/687#issuecomment-1773847997

layday:
But pip is not required for bootstrapping; build would not be built with isolation and pip would not even need to be imported.  If pip is declared as a runtime dependency it needs to actually be installed before build. That can work in one of two ways: if pip is built using flit, or if it's vendored inside build (uh-oh :).

gaborbernat:
Seems it stalled, @sbidoul  let me know if you want to pick it up again and we can reopen.

Issue #692, opened 2023-10-17
project_wheel_metadata returns email.message.Message instead of importlib.metadata._adapters.Message in python 3.9


GianlucaFicarelli:
When using python 3.9.18 (but it happens also with other 3.9.x), `project_wheel_metadata` returns an instance of `email.message.Message` instead of `importlib.metadata._adapters.Message`.

The problem is that the returned object doesn't have the property `json` as defined by [PackageMetadata](https://importlib-metadata.readthedocs.io/en/latest/api.html#importlib_metadata.PackageMetadata.json)

```python
import sys
from build.util import project_wheel_metadata
metadata = project_wheel_metadata(".", isolated=False)
```
```python
>>> print(sys.version)
3.10.8 (main, Nov 16 2022, 15:31:10) [Clang 14.0.0 (clang-1400.0.29.202)]
>>> print(repr(metadata))
<importlib.metadata._adapters.Message object at 0x10e8c3fa0>
>>> hasattr(metadata, "json")
True
```
```python
>>> print(sys.version)
3.9.18 (main, Oct 17 2023, 16:56:25) 
[Clang 14.0.0 (clang-1400.0.29.202)]
>>> print(repr(metadata))
<email.message.Message object at 0x1015f66d0>
>>> hasattr(metadata, "json")
False
```
```python
>>> print(sys.version)
3.8.15 (default, Nov 16 2022, 15:37:40) 
[Clang 14.0.0 (clang-1400.0.29.202)]
>>> print(repr(metadata))
<importlib_metadata._adapters.Message object at 0x10bb06100>
>>> hasattr(metadata, "json")
True
```
As it's shown above, the `json` property works correctly with python 3.10.8 and 3.8.15, but not with python 3.9.18.

Installed packages in python 3.10.8:
```
build==1.0.3
packaging==23.2
pip==23.3
pyproject_hooks==1.0.0
setuptools==68.2.2
tomli==2.0.1
wheel==0.41.2
```
Installed packages in python 3.9.18:
```
build==1.0.3
importlib-metadata==6.8.0
packaging==23.2
pip==23.3
pyproject_hooks==1.0.0
setuptools==68.2.2
tomli==2.0.1
wheel==0.41.2
zipp==3.17.0
```
Installed packages in python 3.8.15:
```
build==1.0.3
importlib-metadata==6.8.0
packaging==23.2
pip==23.3
pyproject_hooks==1.0.0
setuptools==68.2.2
tomli==2.0.1
[...]

PR #693, opened 2023-10-17
Use importlib_metadata for Python < 3.10.2

GianlucaFicarelli:
Closes https://github.com/pypa/build/issues/692.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ dependencies = [
   "pyproject_hooks",
   # not actually a runtime dependency, only supplied as there is not "recommended dependency" support
   'colorama; os_name == "nt"',
-  'importlib-metadata >= 4.6; python_version < "3.10.2"',  # Not required for 3.8+, but fixes a stdlib bug
+  'importlib-metadata >= 4.6; python_full_version < "3.10.2"',  # Not required for 3.8+, but fixes a stdlib bug
   'tomli >= 1.1.0; python_version < "3.11"',
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ dependencies = [
   "pyproject_hooks",
   # not actually a runtime dependency, only supplied as there is not "recommended dependency" support
   'colorama; os_name == "nt"',
-  'importlib-metadata >= 4.6; python_version < "3.10"',  # Not required for 3.8+, but fixes a stdlib bug
+  'importlib-metadata >= 4.6; python_version < "3.10.2"',  # Not required for 3.8+, but fixes a stdlib bug
   'tomli >= 1.1.0; python_version < "3.11"',
 ]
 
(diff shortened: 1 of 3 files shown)


src/build/_importlib.py
```diff
@@ -1,13 +1,8 @@
 import sys
 
 
-if sys.version_info < (3, 8):
+if sys.version_info < (3, 10, 2):
     import importlib_metadata as metadata
-elif sys.version_info < (3, 9, 10) or (3, 10, 0) <= sys.version_info < (3, 10, 2):
```
henryiii, line 6: Please don't remove the try/except part. That helps bootstrapping (where the `.json` property isn't required).

pyproject.toml
```diff
@@ -36,7 +36,7 @@ dependencies = [
   "pyproject_hooks",
   # not actually a runtime dependency, only supplied as there is not "recommended dependency" support
   'colorama; os_name == "nt"',
-  'importlib-metadata >= 4.6; python_version < "3.10"',  # Not required for 3.8+, but fixes a stdlib bug
+  'importlib-metadata >= 4.6; python_version < "3.10.2"',  # Not required for 3.8+, but fixes a stdlib bug
```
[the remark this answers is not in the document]
GianlucaFicarelli: Thanks 👍  Just for the record the correct name is `python_full_version` according to https://packaging.python.org/en/latest/specifications/dependency-specifiers/

src/build/_importlib.py
GianlucaFicarelli: I added the try/except and a comment with the bootstring. Do you think it's correct and understandable?

pyproject.toml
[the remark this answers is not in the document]
henryiii: Oops, yes, I don't use it very often. :)

henryiii (approved):

src/build/_importlib.py
```diff
@@ -1,14 +1,13 @@
 import sys
 
 
-if sys.version_info < (3, 8):
-    import importlib_metadata as metadata
-elif sys.version_info < (3, 9, 10) or (3, 10, 0) <= sys.version_info < (3, 10, 2):
+if sys.version_info >= (3, 10, 2):
+    from importlib import metadata  # type: ignore[attr-defined]
```
GianlucaFicarelli, line 5: The `type: ignore` workaround is needed to pass the "type" job https://github.com/pypa/build/actions/runs/6558894339/job/17813388696?pr=693
I didn't find a better way 🤔

henryiii: Something like this?

```python
if sys.version_info < (3, 8):
    import importlib_metadata as metadata
elif sys.version_info >= (3, 10, 2):
else:
    try:
        import importlib_metadata as metadata
    except ModuleNotFoundError:
        # helps bootstrapping when dependencies aren't installed
```

GianlucaFicarelli: Yes this worked, thanks (side note, I'm surprised that mypy still raises an error when the the order of the first two conditions is inverted)

henryiii (approved):

Issue #626, opened 2023-06-25
pydantic dist.version is None causes TypeError: expected string or bytes-like object, got 'NoneType'


Torxed:
This might be an issue in my environment, and it might belong in `packaging` but throwing out a question/bug report here just in case as I drew a judgement call that it's `build` supplying `packaging` with information in a potentially wrong way. And just to note if it matters, I'm managing packages via my operating system.

When I manually build `pymilter` I get the following error:
```
Traceback (most recent call last):
  File "/usr/lib/python3.11/site-packages/build/__main__.py", line 375, in main
    built = build_call(
            ^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/build/__main__.py", line 208, in build_package
    out = _build(isolation, builder, outdir, distribution, config_settings, skip_dependency_check)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/build/__main__.py", line 147, in _build
    return _build_in_current_env(builder, outdir, distribution, config_settings, skip_dependency_check)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/build/__main__.py", line 127, in _build_in_current_env
    missing = builder.check_dependencies(distribution)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/build/__init__.py", line 370, in check_dependencies
    return {u for d in dependencies for u in check_dependency(d)}
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/build/__init__.py", line 370, in <setcomp>
    return {u for d in dependencies for u in check_dependency(d)}
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/build/__init__.py", line 186, in check_dependency
[...]

Issue #558, opened 2022-12-29
Build fails to build the voila project: TypeError: expected string or bytes-like object


yurivict:
[voila](https://github.com/voila-dashboards/voila)

```
===>  Building for py39-voila-0.4.0
* Getting build dependencies for wheel...

Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/build/__main__.py", line 373, in main
    built = build_call(
  File "/usr/local/lib/python3.9/site-packages/build/__main__.py", line 206, in build_package
    out = _build(isolation, builder, outdir, distribution, config_settings, skip_dependency_check)
  File "/usr/local/lib/python3.9/site-packages/build/__main__.py", line 145, in _build
    return _build_in_current_env(builder, outdir, distribution, config_settings, skip_dependency_check)
  File "/usr/local/lib/python3.9/site-packages/build/__main__.py", line 125, in _build_in_current_env
    missing = builder.check_dependencies(distribution)
  File "/usr/local/lib/python3.9/site-packages/build/__init__.py", line 396, in check_dependencies
    return {u for d in dependencies for u in check_dependency(d)}
  File "/usr/local/lib/python3.9/site-packages/build/__init__.py", line 396, in <setcomp>
    return {u for d in dependencies for u in check_dependency(d)}
  File "/usr/local/lib/python3.9/site-packages/build/__init__.py", line 202, in check_dependency
    if req.specifier and not req.specifier.contains(dist.version, prereleases=True):
  File "/usr/local/lib/python3.9/site-packages/packaging/specifiers.py", line 728, in contains
    item = parse(item)
  File "/usr/local/lib/python3.9/site-packages/packaging/version.py", line 49, in parse
    return Version(version)
  File "/usr/local/lib/python3.9/site-packages/packaging/version.py", line 264, in __init__
    match = self._regex.search(version)
TypeError: expected string or bytes-like object

ERROR expected string or bytes-like object
*** Error code 1
```

Version: 0.9.0, 0.8.0
Python-3.9
FreeBSD 13.1

layday:
I assume that the version returned by `importlib.metadata.distribution.version` for one of your installed packages is not a string.  That would be the case if e.g. the version was missing from the metadata, then it would be `None`.  The `Version` is [required](https://packaging.python.org/en/latest/specifications/core-metadata/) by Core Metadata; metadata without a `Version` field are invalid.

yurivict:
The version returned for one of the packages (hackling) in None.
[...]

PR #695, opened 2023-10-25
Add more helpful error message when metadata is malformed

timkpaine:
xref #558 #626


tests/test_projectbuilder.py
```diff
@@ -21,7 +21,7 @@
 
 DEFAULT_BACKEND = {
     'build-backend': 'setuptools.build_meta:__legacy__',
-    'requires': ['setuptools >= 40.8.0'],
+    'requires': ['setuptools >= 40.8.0', 'wheel'],
```
FFY00, line 24: This seems like a regression. Why is it needed? `wheel` was required on older `setuptools` versions, but shouldn't be anymore. If our minimum `setuptools` version does not yet include the fix to return `wheel` in `get_requires_for_build_wheel`, we should update it.

src/build/_util.py
```diff
@@ -53,6 +53,12 @@ def check_dependency(
         # dependency is not installed in the environment.
         yield (*ancestral_req_strings, normalised_req_string)
     else:
+        if dist.version is None:
+            # Malformed dist-info or some other form of metadata corruption
+            # req.specifier.contains will raise TypeError, so let's do the same
+            # with a more helpful error message
+            msg = f'Package {req.name} has malformed metadata and no version information could be found'
```
FFY00, line 60: nitpick: so that the name is quoted:

```suggestion
            msg = f'Package {req.name!r} has malformed metadata and no version information could be found'
```

src/build/_util.py
```diff
@@ -53,6 +53,12 @@ def check_dependency(
...
+            raise TypeError(msg)
```
FFY00, line 61: I wonder if we shouldn't make this a warning instead and return this package as an unmet dependency. That should give more information to the user. As it stands, they would only see an error message about a specific bad dependency, instead of getting a list of all bad ones with the extra context of why this one was considered bad.

FFY00 (changes requested):

src/build/_util.py
timkpaine: Seems fine to me, sounds reasonable to want all issues at the same time instead of potentially having to address one-by-one

PR #696, opened 2023-10-26
fix: switch to ruff-format

henryiii:
This is 30x faster than black, drops a dependency, and supports single quotes as a setting.

See https://astral.sh/blog/the-ruff-formatter & https://github.com/scientific-python/cookie/pull/299.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,7 +30,7 @@ repos:
     hooks:
       - id: prettier
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.1.2
+    rev: v0.1.3
     hooks:
       - id: ruff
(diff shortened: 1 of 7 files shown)


henryiii:
pre-commit.ci autofix

henryiii:
We have a mix of single and double triple quoted strings in the codebase. This just normalizes on double quoted triple quote strings, following [PEP 257](https://peps.python.org/pep-0257/#what-is-a-docstring).

layday:
You can also remove black's config from pyproject.toml.

henryiii:
Ahh, thanks! I could have sworn I did that, but I think I just started to and didn't finish.

layday (approved):

henryiii:
pre-commit.ci autofix

Issue #617, opened 2023-05-14
[UX] Quote individual dependencies in the list logged to console


webknjaz:
See
```console
* Installing packages in isolated environment... (Cython, expandvars, setuptools-scm-git-archive >= 1.1; python_version < '3.7', setuptools-scm[toml] >= 3.5, < 7.0; python_version < '3.7', setuptools-scm[toml] >= 7.0; python_version >= '3.7', setuptools>=45, toml)
```

There's a few dependencies that are conditional on environment and/or have complex version specifiers. In it, a comma is use in two contexts — as a package request separator and as a separator of parts of a version specifier. This is something an inexperienced individual could easily get confused about when reading the logs.

I think that these deps should be `repr()`ed or smth similar for clarity.

layday:
PR improving the CLI output welcome :)

webknjaz:
It seems like here's a PR now: https://github.com/pypa/build/pull/711

layday:
Addressed in https://github.com/pypa/build/pull/749.

PR #711, opened 2023-12-18
feat: quote installed dependencies

mazzus:
Implements quoted dependencies as requested in #617. This reduces the confusion around commas used to separate the dependencies and commas used in version specifiers.

**Changes**
* Quote the installed dependencies
* Update tests to reflect the new output

**Old output**
```
* Installing packages in isolated environment... (setuptools >= 42.0.0, wheel >= 0.36.0)
```
**New output**
```
* Installing packages in isolated environment... ("setuptools >= 42.0.0", "wheel >= 0.36.0")
```

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -123,7 +123,7 @@ class DefaultIsolatedEnv(IsolatedEnv):
         if not requirements:
             return
 
-        self.log(f'Installing packages in isolated environment... ({", ".join(sorted(requirements))})')
+        self.log(f'Installing packages in isolated environment... ({", ".join(_quote_text(req) for req in sorted(requirements))})')
 
         # pip does not honour environment markers in command line arguments
         # but it does for requirements from a file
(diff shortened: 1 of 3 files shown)


mazzus:
I would appreciate some help with the failing tests. There seems to be an issue with the build process in the tests. I am not sure how that is related to the changes I made.

src/build/env.py
```diff
@@ -123,7 +123,9 @@ def install(self, requirements: Collection[str]) -> None:
         if not requirements:
             return
 
-        self.log(f'Installing packages in isolated environment... ({", ".join(sorted(requirements))})')
+        self.log(
+            f'Installing packages in isolated environment... ({", ".join(_quote_text(req) for req in sorted(requirements))})'
```
webknjaz, line 127: I think there's an `shlex.quote()` helper doing this already. Try using it.

Also, nested things within f-strings are hard to perceive sanely. This calls for a helper variable, I believe.

layday: Let's just have a sub-list, it'll be easier to scan:

```
  - setuptools
  - wheel
```

layday:
> I would appreciate some help with the failing tests. There seems to be an issue with the build process in the tests. I am not sure how that is related to the changes I made.

It looks like 3.7 on Windows with minimum version dependencies started failing nearly two months ago in https://github.com/pypa/build/actions/runs/6742803732, and 3.8 and 3.9 a few days later in https://github.com/pypa/build/actions/runs/6795527020.  Both runs were triggered from the same commit, were executed on the same version runner with the same Python release, and pip installed all of the same versions of build's dependencies.

gaborbernat:
Seems it stalled, @jameshilliard let me know if you want to pick it up again and we can reopen.

PR #716, opened 2023-12-23
Stop injecting `wheel` as a build dep fallback

webknjaz:
PEP 517 doesn't mandate depending on `wheel` when a `__legacy__` setuptools fallback is used. Historically, it used to be assumed as necessary, but later it turned out to be wrong. The reason is that `setuptools`' `get_requires_for_build_wheel()` hook already injects this dependency when building wheels is requested [[1]]. It also used to have this hint in the docs, but it was corrected earlier [[2]].

This patch removes `wheel` from said `requires` list fallback.

[1]: https://github.com/pypa/setuptools/blob/v40.8.0/setuptools/build_meta.py#L130
[2]: https://github.com/pypa/setuptools/pull/3056

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,15 @@
 Changelog
 +++++++++
 
+
+next release
+============
+
+- Stopped injecting ``wheel`` as a build dependency automatically, in the
+  case of missing ``pyproject.toml`` -- by :user:`webknjaz`.
+  (:pr:`716`)
+
+
 1.0.3 (2023-09-06)
 ==================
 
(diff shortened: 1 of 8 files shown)


webknjaz:
_Motivation/backstory:_ I was reading https://gregoryszorc.com/blog/2023/10/30/my-user-experience-porting-off-setup.py/ (it was shared @ https://github.com/pypa/packaging.python.org/issues/1468) and it got me puzzled — I thought I'd gotten rid of all the prominent places newbies may copy-and-paste the suggestion to depend on `wheel` unconditionally...

webknjaz:
Similar PR in pip: https://github.com/pypa/pip/pull/12449.

layday (approved):

henryiii:
Thanks!

Issue #733, opened 2024-02-17
PERF: optional pip install?


henryiii:
I've been looking at UV, and wondered if we could get a speed up with virtualenv/venv by checking for pip in build's environment. If pip is present and of reasonable version, we could skip installing it into the isolated environments and use `--python` to run the outer pip command against the inner environment. That saves at least 200ms per environment, and maybe a lot more if pip has to be downloaded. If pip is not present, then we do what we do now.

Thoughts?

layday:
Sounds reasonable.  We could also look into supporting `uv venv` as an alternative backend.

henryiii:
https://github.com/astral-sh/uv/issues/1510 It would be nice if we could get info about the created venv back as json. We want a couple of paths.

layday:
Didn't you need that same information for https://github.com/wntrblm/nox/pull/762 - how did you make it work?

henryiii:
I believe that hard codes the bin/Script search based on platform.

layday:
Might be good enough for us too :) Alternatively, we could use the same logic we do for the stdlib venv module.

henryiii:
How would we handle opting-in/out? Should we add a `--backend=uv` option (along with `venv` and `virtualenv`)? Or an environment variable? (Or both?). I don't think we want to automatically switch based on the presence of a `uv` command, like we do with virtualenv being an installed module.

layday:
I would be in favour of a CLI option, but I know we've been reluctant to add more options.  The default can be kept dynamic between venv and virtualenv and would be reflected in the CLI's help text, which I think is a plus.  Prolly a good idea to call the new option something other than "backend" though to avoid confusion with build backend - `--venv-impl(ementation)` perhaps?

henryiii:
I meant to mention that "backend" was just a placeholder because I was unimaginative, but forgot. What about `--venv=venv`, `--venv=virtualenv`, `--venv=uv`? With the default basically being `--venv=virtualenv|venv`.

layday:
I'm okay witht that.  How would the new option interact with `-n`? Would they be mutually exclusive? Should `--venv` accept `none` with `-n` becoming a synonym of `--venv=none`?

henryiii:
Why can't we just embed creating the virtualenv and not worry about `virtualenv` and instead just have controls over the installer? We'd need to write out:

```
.venv
├── bin
[...]

PR #736, opened 2024-02-19
feat: use external pip if available

henryiii:
Trying #733 out. Pip is used if it is in the same environment as build and new enough. Saves a bit of time setting up the venvs. Haven't worked on tests yet.

From quick testing building both an sdist and wheel for itself with virtualenv installed this goes from about 4.7 seconds to 3.2 seconds. With `venv` only, it goes from 10.5 seconds to 4.4 seconds.

FYI, with `uv` (hacked in) it took 2.36 secs.

(This is a very simple build, of course, with only a small handful of deps)

uv's experimental native build support took 571ms, but that's only building the wheel (above is SDist and wheel).

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -73,17 +73,20 @@ def _minimum_pip_version() -> str:
 def _has_valid_pip(**distargs: object) -> bool:
     """
     Given a path, see if Pip is present and return True if the version is
-    sufficient for build, False if it is not.
+    sufficient for build, False if it is not. ModuleNotFoundError is thrown if
+    pip is not present.
     """
 
     import packaging.version
 
-    if sys.version_info < (3, 8):
-        import importlib_metadata as metadata
-    else:
-        from importlib import metadata
+    from ._importlib import metadata
 
-    pip_distribution = next(iter(metadata.distributions(name='pip', **distargs)))
+    name = 'pip'
+
+    try:
+        pip_distribution = next(iter(metadata.distributions(name=name, **distargs)))
+    except StopIteration:
+        raise ModuleNotFoundError(name) from None
 
     current_pip_version = packaging.version.Version(pip_distribution.version)
 
@@ -93,16 +96,13 @@ def _has_valid_pip(**distargs: object) -> bool:
 @functools.lru_cache(maxsize=None)
 def _valid_global_pip() -> bool | None:
     """
-    This checks for a valid global pip. Returns None if the prerequisites are
-    not available (Python 3.7 only) or pip is missing, False if Pip is too old,
-    and True if it can be used.
+    This checks for a valid global pip. Returns None if pip is missing, False
+    if Pip is too old, and True if it can be used.
     """
 
     try:
         return _has_valid_pip()
-    except ModuleNotFoundError:  # Python 3.7 only
-        return None
-    except StopIteration:
+    except ModuleNotFoundError:
         return None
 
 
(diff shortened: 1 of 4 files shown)


src/build/env.py
```diff
@@ -83,13 +83,29 @@ def _has_valid_pip(purelib: str) -> bool:
...
+def _valid_global_pip() -> bool | None:
+    """
+    This checks for a valid global pip. Returns None if the prerequisites are
+    not available (Python 3.7 only) or pip is missing, False if Pip is too old,
+    and True if it can be used.
+    """
+
+    try:
+        return _has_valid_pip()
+    except ModuleNotFoundError:  # Python 3.7 only
```
layday, line 103: Why does this happen?

layday: I would prefer if we always threw this from `_has_valid_pip` instead of the cryptic `StopIteration` error.

src/build/env.py
```diff
@@ -83,13 +83,29 @@ def _has_valid_pip(purelib: str) -> bool:
     else:
         from importlib import metadata
```
layday, line 84: Import this from `_importlib`.

src/build/env.py
```diff
@@ -70,7 +70,7 @@ def _minimum_pip_version() -> str:
     return '19.1.0'
 
 
-def _has_valid_pip(purelib: str) -> bool:
+def _has_valid_pip(**distargs: object) -> bool:
```
layday, line 73: I'd rather we didn't lose the typing here, but Python doesn't make it easy:

```py
if typing.TYPE_CHECKING:
    from typing_extensions import NotRequired, TypedDict, Unpack

    class _Distargs(TypedDict):
        path: NotRequired[list[str]]

...

def _has_valid_pip(**distargs: Unpack[_Distargs]) -> bool:
```

src/build/env.py
```diff
@@ -139,6 +155,12 @@ def python_executable(self) -> str:
         """The python executable of the isolated build environment."""
         return self._python_executable
 
+    def _pip_args(self, *, isolate: bool = False) -> list[str]:
```
layday, line 158: This doesn't appear to be ever be called without `isolate` being true.

src/build/env.py
```diff
@@ -201,7 +221,11 @@ def _create_isolated_env_virtualenv(path: str) -> tuple[str, str]:
     """
     import virtualenv
 
-    cmd = [str(path), '--no-setuptools', '--no-wheel', '--activators', '']
+    if _valid_global_pip():
+        cmd = [str(path), '--no-seed', '--activators', '']
```
layday, line 225: The `path` is already a string.

src/build/env.py
```diff
@@ -240,18 +264,22 @@ def _create_isolated_env_venv(path: str) -> tuple[str, str]:
...
     executable, script_dir, purelib = _find_executable_and_scripts(path)
 
     # Get the version of pip in the environment
-    if not _has_valid_pip(purelib):
+    if not _valid_global_pip() and not _has_valid_pip(path=[purelib]):
         _subprocess([executable, '-m', 'pip', 'install', f'pip>={_minimum_pip_version()}'])
 
     # Avoid the setuptools from ensurepip to break the isolation
-    _subprocess([executable, '-m', 'pip', 'uninstall', 'setuptools', '-y'])
+    if _valid_global_pip():
```
layday, line 278: Redundant, setuptools is not installed if `with_pip` is false. (It is also not installed on Python 3.12+.)

src/build/env.py
henryiii: Okay. And the error was if importlib_metadata was missing, though we have it as a dep, so that would only be for bootstrapping. This makes it have the same signature as pip missing, though, so no need to special case.

src/build/env.py
henryiii: I left it along for now, would prefer a `_compat/typing` module if we add it.

src/build/env.py
henryiii: Okay, dropped the arg.

src/build/env.py
henryiii: Ah, okay, yes. Removed this branch.

src/build/env.py
layday: +1 to rename it to `_compat/importlib`.

src/build/env.py
henryiii: I can do that in a separate PR. I also usually have `_compat/tomllib` (https://github.com/scikit-build/scikit-build-core/tree/main/src/scikit_build_core/_compat for example).

layday (approved):
Thanks!

PR #738, opened 2024-02-22
ci: try using tox-uv

henryiii:
Speeds up CI a little bit. Closes #858.

### DIFF
--- a/.github/workflows/reusable-pytest.yml
+++ b/.github/workflows/reusable-pytest.yml
@@ -71,7 +71,7 @@ jobs:
           python-version: "3.12"
 
       - name: Install tox
-        run: python -m pip install tox
+        run: python -m pip install tox tox-uv uv
 
       - name: Run test suite via tox
         if: matrix.tox-target == 'tox'
--- a/.github/workflows/reusable-pytest.yml
+++ b/.github/workflows/reusable-pytest.yml
@@ -68,11 +68,6 @@ jobs:
               file.write(env)
         shell: python
 
-      - name: Setup python for tox
-        uses: actions/setup-python@v5
-        with:
-          python-version: "3.12"
-
       - name: Install tox (uv)
         if: matrix.py != 'pypy3.8'
         run: uv tool install tox --with tox-uv
--- a/.github/workflows/reusable-pytest.yml
+++ b/.github/workflows/reusable-pytest.yml
@@ -49,6 +49,9 @@ jobs:
           python-version: ${{ matrix.py }}
           allow-prereleases: true
 
+      - name: Setup uv
+        uses: astral-sh/setup-uv@v4
+
       - name: Pick environment to run
         run: |
           import platform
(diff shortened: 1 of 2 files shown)


FFY00:
Perhaps we could try using https://github.com/tox-dev/tox-gh?

henryiii:
The hope here was to make the tests a bit faster; tox-gh might shorten the config, but wouldn't be faster (AFAIK). I don't understand why one test now uninstalls setuptools when it didn't before when running with tox-uv.

layday (approved):

.github/workflows/reusable-pytest.yml
```diff
@@ -68,10 +71,15 @@ jobs:
       - name: Setup python for tox
         uses: actions/setup-python@v5
         with:
-          python-version: 3.9
+          python-version: 3.12
```
layday, line 74:
```suggestion
          python-version: "3.12"
```

henryiii: Actually, I think this isn't needed anymore, since I changed to to be installed via `uv tool`. Let's check that first.

henryiii:
Intersting, the three failures are: `AssertionError: Expected 'get_paths' to be called once. Called 4 times.` from `test_can_get_venv_paths_with_posix_local_default_scheme[False]`.

tests/test_env.py
```diff
@@ -46,7 +46,11 @@ def test_can_get_venv_paths_with_conflicting_default_scheme(
     assert get_scheme_names.call_count == 1
 
 
-@pytest.mark.skipif('posix_local' not in sysconfig.get_scheme_names(), reason='workaround for Debian/Ubuntu Python')
+SCHEME_NAMES = sysconfig.get_scheme_names()
+
+
+@pytest.mark.skipif('posix_local' not in SCHEME_NAMES, reason='workaround for Debian/Ubuntu Python')
+@pytest.mark.skipif('venv' in SCHEME_NAMES, reason='different call if venv is in scheme names')
```
layday, line 53: Isn't `venv` always in `SCHEME_NAMES`, since 3.11?

henryiii: How does this test pass, then? https://github.com/pypa/build/blob/e9adbf04f98745185ffb6552a7088a8ec12cfe2f/src/build/env.py#L346-L353 seems to trigger a path that would fail this test.

henryiii: (`get_paths.assert_called_once_with(scheme='posix_prefix', vars=mocker.ANY`, specifically)

gaborbernat (approved):

tests/test_env.py
layday: Sorry, I forgot to reply here. This test [isn't covered](https://app.codecov.io/gh/pypa/build/blob/henryiii%2Fbuild%3Ahenryiii%2Fci%2Fub/tests%2Ftest_env.py), so I assume it's always skipped.

PR #741, opened 2024-02-26
fix: support invalid folders in cwd

henryiii:
Noticed in https://github.com/astral-sh/uv/issues/1972. Fix pulled from Pip, but putting it in `__main__.py` doesn't seem to be working - runpy seems to load `__init__.py` before `__main__.py`. This affects `python -m venv` too, so not sure we can do much about it.

Placing this at the top of `__init__.py` does fix the problem, though. I think `pipx run build` would still be broken since that breaks before getting to build.

`pyproject-build` and `PYTHONSAFEPATH=1 python -m build` (3.11+) both work fine as is.

To reproduce, clone `mihalikv/improved-umbrella` and switch to the `nested_lib` dir and try to build.

### DIFF
--- a/src/build/__init__.py
+++ b/src/build/__init__.py
@@ -6,15 +6,20 @@ build - A simple, correct PEP 517 build frontend
 
 from __future__ import annotations
 
+import os
+import sys
 
-__version__ = '1.0.3'
+
+# Remove '' and current working directory from the first entry
+# of sys.path, if present to avoid using current directory
+if sys.path[0] in ('', os.getcwd()):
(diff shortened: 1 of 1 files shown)


henryiii:
Not sure this is worth fixing. It's easy to work around in Python 3.11+ with PYTHONSAFEPATH. Should we not worry about it?

layday:
This is fundamental to how `runpy` works - it applies to build just as much as any other tool.  Let's not worry about it, and also maybe consider advertising `pyproject-build` more prominently, which bypasses `runpy`.

PR #742, opened 2024-02-26
compat: create separate importlib, tarfile and tomllib shims

layday:


henryiii:
Looks good. Did we want to pull out tomllib here too, or in a follow-up, etc?

henryiii (approved):

PR #746, opened 2024-02-29
env: fix global pip

layday:


src/build/env.py
```diff
@@ -101,7 +101,8 @@ def _valid_global_pip() -> bool | None:
     """
 
     try:
-        return _has_valid_pip()
+        # Version to have added the `--python` option.
```
layday, line 104: https://github.com/pypa/pip/commit/0a76da3a94130fad58b086e331c3d3e1b02a89eb

henryiii: I thought this was a newer option but didn't see it from quick searches. Thanks!

PR #748, opened 2024-03-01
build: split `__init__` module and add `_types`, `Distribution`

layday:


src/build/__main__.py
```diff
@@ -117,10 +120,10 @@ def _build_in_isolated_env(
 
 
 def _build_in_current_env(
-    srcdir: PathType,
-    outdir: PathType,
-    distribution: str,
-    config_settings: ConfigSettingsType | None,
+    srcdir: _types.StrPath,
```
gaborbernat, line 123: I'd prefer to import these directly without needing to namespace access it like here...

layday: Agreed, done.

src/build/__init__.py
```diff
@@ -1,387 +1,31 @@
...
     'FailedProcessError',
     'ProjectBuilder',
     'RunnerType',
     'TypoWarning',
-    'check_dependency',
 ]
-
-
-def __dir__() -> list[str]:
-    return __all__
```
henryiii, line 387: We shouldn't lose this, it keeps things like `annotations` from showing up in tab completion, and is needed for mypy re-exports.

layday: Needed for mypy re-exports - what do you mean?

I removed it because the `__init__` module doesn't contain anything other than what's exported, barring `annotations`.

src/build/__init__.py
```diff
@@ -1,387 +1,31 @@
-# SPDX-License-Identifier: MIT
-
-"""
-build - A simple, correct PEP 517 build frontend
-"""
-
```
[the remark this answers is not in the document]
layday: Restored and made to match the readme ("Python build frontend" iso "PEP 517 build frontend").

src/build/__init__.py
layday: (Restored.)

layday: I’m not fussed either way - I’ve put it back 😄

PR #749, opened 2024-03-01
main: improve console output

layday:


layday:
Screenshot 2024-03-01 at 21 28 30

henryiii:
Would it make sense to remove the bold on the package names? I think that would look a little more distinct:

Screenshot 2024-03-01 at 2 37 30 PM

layday:
That makes sense, I've changed it.

henryiii:
Much better, I think:

Screenshot 2024-03-01 at 3 55 57 PM

layday:
Any thoughts on the verbosity option?

henryiii:
Basically just passes through to the underlying pip? Seems fine.

gaborbernat (approved):

PR #751, opened 2024-03-05
env: add support for uv

layday:


pyproject.toml
```diff
@@ -63,11 +63,15 @@ test = [
...
 ]
 typing = [
+  "build[uv]",
   "importlib-metadata >= 5.1",
   "mypy ~= 1.5.0",
   "tomli",
   "typing-extensions >= 3.7.4.3",
 ]
+uv = [
+  "uv >= 0.1.6",
```
gaborbernat, line 73: Why not 0.1.15?

layday: Can do, v0.1.6 was the version which exposed `find_uv_bin`: https://github.com/astral-sh/uv/commit/469ccf826f9e58d452877fb0c2888d189423e040 (familiar name :)

gaborbernat: Unless we run the test suite with 0.1.6 I'd prefer to use latest the CI actually runs again... Otherwise, how do you know it works?

layday: I would've added it to our `constraints.txt` but I don't think we need to be backward-compatible with uv versions prior to our supporting it - I'll bump the version number.

layday: I've done that [here](https://github.com/layday/instawow/blob/896dfcab3a757a8e3849096d8cab03fcb0dafe14/noxfile.py#L81-L83). The only trouble is that it doesn't have a second-level dependency resolution option so you've got to pluck the package's direct dependencies out of the wheel metadata (or `pyproject.toml`).

layday:
It looks like `uv venv` doesn't work with PyPy on Windows.

gaborbernat:
Yeah https://github.com/astral-sh/uv/issues/2096, there's no official support for PyPy today, and what works, it works by chance.

webknjaz:
@layday have you considered exposing an API for pluggable build env provisioners?

webknjaz:
> Can you go into more detail?

I mean, it'd be nice if third parties could stick their bits of build env provisioning logic through some standardized interface. This could be useful for various downstreams and probably other tools that act as frontends.

layday:
Isolated envs are already easily pluggable in code (see the `IsolatedEnv` protocol and `ProjectBuilder.from_isolated_env`), but I don't see us wanting to go down the path of exposing a plug-in interface for isolated env selection through the CLI.

layday:
So, this adds an `--env-impl` flag to the CLI, with the only accepted value being `venv+uv`. Old semantics apply if omitted.

I've modified the venv selection order slightly so that venv is preferred whenever a valid outer pip is present for a minor speed boost.  `virtualenv` continues to be preferred with an outdated pip.

The `DefaultIsolatedEnv` is now tested with all venv and installer combos (`virtualenv` used to be mocked). This revealed the `venv` module was preferred with `packaging` 19.0 even when a valid `virtualenv` was present.  The culprit was a marker evaluation bug in `packaging` which was fixed in 19.1.

Finally, I've added a setuptools existence check before attempting to uninstall the ensurepip copy of setuptools - it was dropped in Python 3.12.  Again, this should make using venv without an outer pip slightly faster.

layday:
Currently, we're only looking for a `uv` executable on `$PATH` (not attempting an import), but see the discussion happening at https://github.com/wntrblm/nox/pull/795#discussion_r1517325469.

src/build/env.py
```diff
@@ -177,35 +266,56 @@ def install(self, requirements: Collection[str]) -> None:
...
-        cmd = [path, '--no-seed', '--activators', '']
-    else:
-        cmd = [path, '--no-setuptools', '--no-wheel', '--activators', '']
+        self._env_path = path
+
+        # ``uv.find_uv_bin`` will look for uv in the user prefix if it can't
+        # find it under ``sys.prefix``, essentially potentially rearranging
+        # the user's $PATH.  We'll only look for uv under the prefix of
+        # the running interpreter for unactivated venvs then defer to $PATH.
+        uv_bin = shutil.which('uv', path=sysconfig.get_path('scripts'))
```
gaborbernat, line 288: uv comes with a find_uv_bin method we should use that.

gaborbernat (changes requested):

src/build/env.py
layday: The comment explains why it's not being used.

gaborbernat: Guess I agree here with uv's choice, I find it more expected, and disagree with the comment.

src/build/env.py
```diff
@@ -177,35 +266,56 @@ def install(self, requirements: Collection[str]) -> None:
...
+
+        if not uv_bin:
+            uv_bin = shutil.which('uv')
```
gaborbernat, line 291: I strongly disagree with such fallback. Either pick uv from sys.prefix or fail.

src/build/env.py
layday: How is it expected that a uv you've installed in a venv might load a uv from an outer env?

```
$ python -m venv --without-pip foo
$ VIRTUAL_ENV=foo uv pip install uv
Resolved 1 package in 3ms
Installed 1 package in 5ms
 + uv==0.1.16
$ foo/bin/python -c 'import uv; print(uv.find_uv_bin())'
/Users/dimitris/code/python-packaging/build/foo/bin/uv
$ rm foo/bin/uv
$ foo/bin/python -c 'import uv; print(uv.find_uv_bin())'
/Users/dimitris/.local/bin/uv
```

If uv should function as a Python package, it has no business looking outside its prefix.

I'd personally prefer to take a standards-based approach than rely on `uv`'s custom script finder.  Is this not the correct way to locate a script under the running interpreter's prefix?

gaborbernat: Open an issue in UV and we should fix it there then.

layday: Chaned to only use `uv.find_uv_bin`.

src/build/env.py
henryiii: I've generally been respecting the design of uv, which is to _not_ install itself in environments, but to be used globally (via https://astral.sh/uv/install.sh, cargo install, homebrew, pacman, etc) and only target environments. If a user does install uv in build's environment, that should take preference, but I think we should try to use an outer/system `uv` otherwise. I don't think someone passing `--installer=uv` would be unhappy if a `uv` executable on their path is used. (same with `cmake`, `gcc`, `ninja`, `patchelf`, etc., lots of executables get used from the path)

But willing to go with this if you really prefer.

src/build/env.py
```diff
@@ -21,6 +20,11 @@
 from ._util import check_dependency
 
 
+Installer = typing.Literal['pip', 'uv']
```
layday, line 23: Do we wanna call this `uv-pip` instead of simply `uv`?

src/build/env.py
gaborbernat: Then remove the UV extra because user should install it separately? Having an extra but then taking a global is unexpected to me.

henryiii: A user _can_ install it, say if the system `uv` is older than they like, or if they just use `pipx run build[uv]` regardless of if the system has uv or not. But, if they are on a system that does not support binary wheels, like BSDs, they might prefer or even need to install it globally and not use the extra.

(This is very common for binary deps with optional PyPI packages like CMake and ninja)

gaborbernat: We are not in the business of installing cmake, why is UV any different (other than it happens to be available on pypi, is the same a random binary).

henryiii: I’m saying it is similar (any binary CLI tool with a PyPA distribution), and this what experience has shown is the most useful way to handle it. PyPI package preferred, binary on path second is used in scikit-build and many hand written packages. Otherwise you can’t control the CMake version on binary supported platforms or you can’t build at all on non binary supported platforms.

Also, other distributions of uv might not include this Python package, the CMake Python wrapper is not included in conda-forge, for example.

gaborbernat: I don't see why we have to support that use case. From my point of view the only use case we need to target is installing from pypi. If the users install it through their operating system package manager they can ask for an os package variant that includes that package.

henryiii: Also, this is opt-in, so a user will never be surprised by uv on the path or installed when they weren’t trying to use uv. “Try to use local uv the try to use global uv when a user explicitly requests uv” seems quite reasonable.

If you want to just go for one of the two, though, feel free, this is just my opinion (based heavily on experience distributing CMake and ninja packages).

src/build/env.py
henryiii: I think the `--installer` implies uv pip. And if they added a “uv install”, we could transparently start using that instead.

src/build/env.py
henryiii: Why would non-PyPI distributions include the Python helpers? They aren’t required and require the package to depend on python(s). Homebrew doesn’t tie this to (each) version of Python, for example.

henryiii: And you can’t install a system package manager distribution’s Python helpers into an environment anyway.

gaborbernat: But then you're saying that this will only be installed into virtual environment, in which case there should be no case when the UV it's not pulled from Pypi and be the latest, and available alongside the build pythons environment. And if that is the case why do we need the ability to specify / use external versions of it? I just don't get it 😅

gaborbernat: Let's put it this way if the user submits a bug report, what information do we need to understand and replicate the issue? Until now we just require the installed packages into the build project python environment. If all of a sudden we allow external applications even by accident then how do we know what version of UV are they using? Wouldn't be surprising if they're the UV package inside the environment they are running and for whatever reason we are using some external variant available on to the path? The only say I would be okay with specifying external applications / dependencies would be if they would have to be explicitly specified during the invocation inside the CLI.

henryiii: >  it's not pulled from PyPI

The procedure would be, for example on FreeBSD:

```console
$ pkg install uv
$ pipx run build --installer=uv 
```

Ports has uv (https://cgit.freebsd.org/ports/tree/devel/uv), and it _cannot get binary packages from PyPI_; there is no wheel tag for BSD. You cannot use `pipx run build[uv] --installer=uv` on any of the BSDs (at least not without the Rust toolchain, etc). This is the only way to get use it without downloading the Rust toolchain and building from source (assuming it works without dependencies; Ruff has a hidden dependency on GNU Make for a C library that causes issues on minimal docker images).

This is exactly how you build scikit-build-core libraries with cmake and ninja, or meson-python libraries with ninja; you pre-install cmake/ninja of sufficient version from Ports, then you build, and the CLI versions are used instead of requiring them from PyPI - which won't work, since there are no binaries and CMake is required to build them.

I think it's important to view the `uv` PyPI package as simply a distribution of a CLI tool, and not the canonical way to get it. It has a small helper Python file too, but no other distribution includes that - and that's okay, it's just there to make the PyPI package a little friendlier. (And IMO, from cmake and ninja experience, I would have recommended not including a Python helper, I think, it causes lots of arguments about if they should be included in other distributions (which is a clear no, you don't want to force a (specific) Python dependency when there isn't one)).

> Let's put it this way if the user submits a bug report, what information do we need to understand and replicate the issue?

If a user specifies `--installer=uv`, then I think it's perfectly fair to ask them for the output of `pip list` and `uv --version`. It's only an opt-in "advanced" feature.

>  If all of a sudden we allow external applications

Binary builds already use external applications. PyPI doesn't ship compilers. Ruff for example, besides requiring a Rust toolchain, also requires GNU Make (on UNIX, though there's a way to turn of the component that depends on a C lib build). Etc. Installing `build[uv]` requires at least a Rust toolchain.

> The only say I would be okay with specifying external applications / dependencies would be if they would have to be explicitly specified during the invocation inside the CLI.

This is explicitly activated in the CLI. It never runs by default. You are literally passing `--installer=uv`. The only thing we don't do is allow a full path here, it has to be on the PATH if it's not co-installed with build. I think having `--installer=uv` look for `uv` on the path is reasonable.

henryiii: > Wouldn't be surprising if they're the UV package inside the environment they are running and for whatever reason we are using some external variant available on to the path? 

Oh, I do think the uv package should take precedence over the external search. That was a discussion in nox, too, and we've gone with the way that it's been done for years for cmake/ninja, local installs always override external tools.

gaborbernat: I guess I can accept falling back to the operating system version, but if we do that in my opinion we must print out of warning onto the standard output or error to notify the user that this is happening.

henryiii: Something like this would be perfect, I think:

```
* Creating venv isolated environment...
* Using uv 0.1.16 from /usr/local/bin/uv to install packages...
* Installing packages in isolated environment... (hatch-fancy-pypi-readme, hatch-vcs, hatchling)
```

Whereas if it was from the local environment, it could be simplified:

```
* Creating venv isolated environment...
* Using uv 0.1.16 to install packages...
* Installing packages in isolated environment... (hatch-fancy-pypi-readme, hatch-vcs, hatchling)
```

Would that work?

gaborbernat: Perfect.

layday: Reminder that we have a verbosity flag now that will print out the command invoked including the full uv executable path.  Printing out the version for uv executable from $PATH would require an additional subprocess call and parsing the output of `-V` which is a flimsy affair.  This same consideration also applies to an outer, although the surprise factor is slightly diminished since we’re only looking in the running interpreter’s Python path.

layday: Screenshot 2024-03-10 at 11 08 41

How's this?

gaborbernat (approved):

PR #752, opened 2024-03-06
perf: avoid compiling unused bytecode

henryiii:
Compiling bytecode is supposed to make usage more consistent for interaction. But it compiles everything, which isn't required for a single use non-interactive temporary environment. We can potentially save a bit of time this way.

This will need to be skipped for the uv backend in #751, since uv defaults to not compiling bytecode, and instead requires a flag to do so (most recent version). But flags aren't identical anyway.

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -261,6 +261,7 @@ class _PipBackend(_EnvBackend):
                 'install',
                 '--use-pep517',
                 '--no-warn-script-location',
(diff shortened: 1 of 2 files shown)


gaborbernat (approved):

henryiii:
I’ll leave merging this up to @layday so it doesn’t interfere with #751.

layday (approved):

PR #758, opened 2024-03-27
chore: prepare for 1.2.0

henryiii:
Would a 1.2 release make sense?

I also single-sourced the version, unless there's some reason we weren't doing that.

Also, https://github.com/pypa/pyproject-metadata/issues/68#issuecomment-2020350973 has instructions for signing GitHub release with Sigstore, we could set that up if desired. PyPI no longer allows GPG signing, which is still in our instructions (it just gets stripped out, I tried it on the last release).


pyproject.toml
```diff
@@ -31,7 +30,8 @@ urls.changelog = "https://build.pypa.io/en/stable/changelog.html"
 urls.homepage = "https://build.pypa.io"
 urls.issues = "https://github.com/pypa/build/issues"
 urls.source = "https://github.com/pypa/build"
-
+# Version read from __version__ field in __init__.py by Flit
+dynamic = ["version"]
```
layday, line 34: Is flit clever about this or will it require installing all of our runtime deps?

henryiii: I'm pretty sure it's clever about it. But good point, let me check.

gaborbernat (approved):

pyproject.toml
henryiii: https://github.com/pypa/flit/blob/8ead22d4d3028f2dd2c952c85e5bf1535cca3b40/flit_core/flit_core/common.py#L140

CHANGELOG.rst
```diff
@@ -6,7 +6,20 @@ Changelog
 Unreleased
```
layday, line 6: Rm?

henryiii: I left that there for our next release, but yes, I can drop it until we have something to put there.

layday: Only because it's gonna be visible in the published docs.

layday (approved):

PR #759, opened 2024-03-27
ci: sign the files with sigstore

henryiii:
Adds sigstore signing. Still not automatically released.


henryiii:
Ideally, PRs from within the repo are fine, just not forks, but not sure how to detect that in GHA.

henryiii:
(because I made the PR from a fork, won't really be able to check it until we merge it)

PR #761, opened 2024-03-28
fix: support min width not detectable

### LINKED ISSUE #760: Terminal size error with 1.2.0
Hi there,

It appears that the changes added in https://github.com/pypa/build/pull/749 have the potential to break usage of `build` when the terminal size cannot be determined.

In our CI pipeline on Buildkite, we call `python -m build -w` on a `python:3.8` Docker container. Since the release of `build==1.2.0` this is failing with the following error:

image

The call to `shutil.get_terminal_size().columns` is evidently returning zero in this case and, as per the [docs](https://docs.python.org/3/library/shutil.html#shutil.get_terminal_size), when using Python < 3.11 the default `fallback` value will not be used.

> Changed in version 3.11: The fallback values are also used if [os.get_terminal_size()](https://docs.python.org/3/library/os.html#os.get_terminal_size) returns zeroes.

While it is possible to work around the issue in our pipeline, I think this scenario should be better handled by `build`.

henryiii:
Fix #760.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,11 @@
 Changelog
 +++++++++
 
+1.2.1 (2024-03-28)
+==================
+
+- Avoid error when terminal width is undetectable on Python < 3.11
+  (PR :pr:`761`)
 
 1.2.0 (2024-03-27)
 ==================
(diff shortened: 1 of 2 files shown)


layday (approved):
Thank you.

`python -m build -h` would have failed too before 1.2.0 with sub-zero values, but it’s not something that would’ve been run in CI.

PR #764, opened 2024-03-29
build: add `editable` to dist type

### LINKED ISSUE #763: types in release 1.2.1 exclude "editable"
`builder.get_requires_for_build("editable")`

but the types insist that the value should be `Literal['wheel', 'sdist']`

layday:
Closes #763.

### DIFF
--- a/src/build/_builder.py
+++ b/src/build/_builder.py
@@ -11,7 +11,7 @@ import warnings
 import zipfile
 
 from collections.abc import Iterator
-from typing import Any, Mapping, Sequence, TypeVar
+from typing import Any, Literal, Mapping, Sequence, TypeVar
 
 import pyproject_hooks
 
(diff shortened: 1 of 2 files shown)


dimbleby:
thanks!  an alias for the three-way literal would be a convenience for callers, though of course we can write our own

layday:
We could, though I'm not sure what would be a good name for it. I could also see us adding `editable` to the `Distribution` set and creating a secondary, internal literal for the CLI. I'm inclined to leave it as is for now.

dimbleby:
> I'm not sure what would be a good name for it.

Haha, well I guess callers will just have to answer that ourselves.  

Honestly I think that declining to put a name on this part of the nterface is a minor failure, but it's one that users can paper over.

henryiii (approved):

PR #766, opened 2024-03-31
Solicit isolation setting from the environment.

### LINKED ISSUE #556: Downstream packager unable to indicate "no isolation" for offline builds
In jaraco/jaraco.packaging#7 and jaraco/jaraco.packaging#9 and jaraco/jaraco.packaging#10, downstream packagers are struggling with the isolated builds as created by `build.utils.project_wheel_metadata`. These packagers would like to perform offline builds (preventing access to the network for resolving dependencies). The `project_wheel_metadata` call does allow for a `isolated=False`, but in the case where it's being enacted programmatically, the downstream user doesn't have any control.

I could imagine that each consumer of `project_wheel_metadata` could expose an environment variable or other configuration mechanism to alter the behavior, but probably `build` should do that instead. If, for example, the definition was `project_wheel_metadata(isolated=get_bool_env('BUILD_ISOLATED', True))`, then the downstream consumers could set `BUILD_ISOLATED=false` to disable isolated builds (and put the onus on them to supply the dependencies).

jaraco:
Closes #556.

### DIFF
--- a/src/build/util.py
+++ b/src/build/util.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import os
 import pathlib
 import tempfile
 
@@ -21,7 +22,7 @@ def _project_wheel_metadata(builder: ProjectBuilder) -> importlib.metadata.Packa
 
 def project_wheel_metadata(
     source_dir: StrPath,
-    isolated: bool = True,
+    isolated: bool = os.environ.get('BUILD_ENVIRONMENT', 'isolated') == 'isolated',
     *,
     runner: SubprocessRunner = pyproject_hooks.quiet_subprocess_runner,
 ) -> importlib.metadata.PackageMetadata:
@@ -34,7 +35,9 @@ def project_wheel_metadata(
     :param source_dir: Project source directory
     :param isolated: Whether or not to run invoke the backend in the current
                      environment or to create an isolated one and invoke it
-                     there.
+                     there. Defaults to True (isolated), but with
+                     ``BUILD_ENVIRONMENT=current`` set, the default
+                     is False.
     :param runner: An alternative runner for backend subprocesses
     """
 
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -1,6 +1,8 @@
 # SPDX-License-Identifier: MIT
 
 import importlib.util
+import sys
+import subprocess
 
 import pytest
 
@@ -47,3 +49,11 @@ def test_with_get_requires(package_test_metadata):
     assert str(metadata['version']) == '1.0.0'
     assert metadata['summary'] == 'hello!'
     assert isinstance(metadata.json, dict)
+
+
+def test_current_environment_default():
+    check_script = "import build.util, inspect; print(inspect.signature(build.util.project_wheel_metadata).parameters['isolated'].default)"
+    cmd = [sys.executable, '-c', check_script]
+    env = {'BUILD_ENVIRONMENT': 'current'}
+    res = subprocess.check_output(cmd, env=env, text=True, encoding='utf-8')
+    assert res.strip() == 'False'


layday:
This has the appearance of 'spooky action at a distance' - I'm not in favour of modifying library functions through environment variables, which take effect globally.  An env var-controlled function can have multiple consumers stepping on each other's toes.

In the original ticket, you say:

> I could imagine that each consumer of project_wheel_metadata could expose an environment variable or other configuration mechanism to alter the behavior, but probably build should do that instead. 

I think this is something that should in fact be done by the consumer and not by build.

jaraco:
> I think this is something that should in fact be done by the consumer and not by build.

Since there is more than one consumer (at least pytest-checkdocs and jaraco.packaging), requiring the setting to be solicited by the consumer requires all of the consumers to coordinate or provide locally-unique settings. Do we really want `JARACO_PACKAGING_BUILD_ENVIRONMENT` and `PYTEST_CHECKDOCS_BUILD_ENVIRONMENT` (and one more for each consumer) and then ask downstream builders to configure each of these settings based on which tools that happen to be in use? Alternatively, each of these consumers could potentially solicit `BUILD_ENVIRONMENT` or `PYPROJECT_BUILD_ENVIRONMENT` and then honor than when calling `project_build_metadata`, which would have the same effect as this change, _except_ that other consumers would not get that behavior and there would be no good authority or forum for coordinating the use of this common setting (maybe `jaraco.packaging` implements one value but `pytest-checkdocs` likes the other, or maybe an issue is reported and one package changes the name but the other doesn't, or maybe both manage to coordinate and make the change at the same time but downstream they get different versions with alternate behaviors). The whole point of filing the issue was to point out that there is a common concern that's bound to the signature of this function.

This behavior doesn't feel that much more spooky than other environmental build flags like `PIP_NO_INDEX` or `PIP_FIND_LINKS`, which drastically change the behavior at a distance.

Even the consumers of this API don't have a great way to solicit this behavior, so they solicit it through environment variables. One is a sphinx plugin, the other a pytest plugin, so it's not obvious what direct method they could have to influence the behavior, and there's almost certainly no common solution across those two CLI surfaces (i.e. no argument you could pass to `sphinx build` and `pytest` that would signal to use non-isolated builds whenever metadata is needed).

To my knowledge, these are just two consumers of this API, and these are possibly the only consumers of this API, and I do maintain both, so maybe I shouldn't be concerned about the coordination concerns, but my hope was to solicit the behavior where it's needed but make it reachable by the builders without having to plumb it through every consumer.

At the very least, I'd like to consider adding a helper function that consumer could use thus:

```python
from build.util import project_wheel_metadata, infer_isolation

project_wheel_metadata(..., isolated=infer_isolation())
```

So that the logic behind the environment variable and inference could be consolidated. It still would require each consumer to opt into the inference, and it still would be spooky action at a distance, but I don't see a good alternative.

In light of those concerns, does that shift your opinion at all?

jaraco:
I'm pretty sure the test failures are unrelated to my changes.

layday:
> This behavior doesn't feel that much more spooky than other environmental build flags like `PIP_NO_INDEX` or `PIP_FIND_LINKS`, which drastically change the behavior at a distance.

These env vars (are only advertised to) work with the pip CLI and not with its programmatic interface. I've no expectation that a hypothetical pip API would be parameterisable via env vars.  It sounds like you're planning on invoking the build API in a subprocess (and are comparing it with pip as a result), but there's no reason why that should always be the case.  I don't think `project_wheel_metadata` is very useful in its current form, but opening it up to manipulation 'at a distance' probably isn't the solution.

jaraco:
It sounds like maybe there's no path forward with build. My next best solution is to provide this behavior through another library. As much trouble as it's been to contribute this functionality to pep517 and now to build, I'm conceding defeat. If build is refusing to address the concerns of downstream builders, we'll need yet another layer to address concerns. I guess I'll tackle the problem in jaraco.packaging.

PR #778, opened 2024-05-02
build: make compatible with typed `pyproject-hooks`

layday:


layday:
Right, `SubprocessRunner` doesn't exist at runtime because `pyproject-hooks` maintains support for Python 3.7.

henryiii:
I think this is the remaining bug:

https://github.com/pypy/pypy/issues/4958

I'm not sure how to go about patching it for now though. I think if tox normalized the path on Windows, that would fix it. And if PyPy releases a fix, 3.9 will likely remain broken, leaving only 3.8 and 3.10 working.

layday:
LGTM (can’t self approve)

henryiii (approved):

PR #782, opened 2024-05-28
ci: publish artifact attestation and PyPI packages

henryiii:
This adds Trusted Publisher based PyPI publishing and GitHub Artifact Attestations for the binaries, which can be queried from GitHub after publishing at https://github.com/pypa/build/attestations. Not supported on PyPI yet, [PEP 740](https://peps.python.org/pep-0740/) will add support.

Replaces https://github.com/pypa/build/pull/759. Needs setup on PyPI (which I can do if we think this is a good idea).

WDYT?

- https://github.blog/2024-05-02-introducing-artifact-attestations-now-in-public-beta/
- https://docs.github.com/en/actions/security-guides/using-artifact-attestations-to-establish-provenance-for-builds
- PEP 740

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -14,4 +14,30 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+
       - uses: hynek/build-and-inspect-python-package@v2
+
+  publish:
+    needs: [dist]
+    runs-on: ubuntu-latest
+    if: github.event_name == 'release' && github.event.action == 'published'
+    environment:
+      name: pypi
+      url: https://pypi.org/p/build
+    permissions:
+      id-token: write
+      attestations: write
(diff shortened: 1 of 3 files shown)


.github/workflows/cd.yml
```diff
@@ -14,4 +14,30 @@ jobs:
...
+    needs: [dist]
+    runs-on: ubuntu-latest
+    if: github.event_name == 'release' && github.event.action == 'published'
+    environment:
+      name: pypi
+      url: https://pypi.org/p/build
+    permissions:
+      id-token: write
+      attestations: write
+      contents: read
```
henryiii, line 30:
```suggestion
```

I think this is not needed.

FFY00 (approved):
LGTM

We should also update the [installation page in the docs](https://build.pypa.io/en/stable/installation.html), as that's now outdated.

PR #805, opened 2024-08-28
docs: Fix forward reference in docs check

edgarrmondragon:
https://github.com/pypa/build/pull/804#issuecomment-2315744235

### DIFF
--- a/src/build/util.py
+++ b/src/build/util.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import os  # noqa: F401
 import pathlib
 import tempfile
 


henryiii:
Looks like setuptools 74 broke our tests. I'm fixing other fallout from 74 in scikit-build (classic) currently. :)

henryiii:
Going to fix in #804 since the docs don't run if they aren't changed.

edgarrmondragon:
> Going to fix in #804 since the docs don't run if they aren't changed.

Gotcha 👍, I'll close this then.

PR #807, opened 2024-08-30
tests: optional uv

carlwgeorge:
Skip the tests that need uv when uv is not installed.  This makes it easier for Linux distros that package build to avoid a build-time dependency on uv.

### DIFF
--- a/tests/test_env.py
+++ b/tests/test_env.py
@@ -21,6 +21,7 @@ import build.env
 
 IS_PYPY = sys.implementation.name == 'pypy'
 IS_WINDOWS = sys.platform.startswith('win')
+MISSING_UV = not shutil.which('uv')
 
 
 @pytest.mark.isolated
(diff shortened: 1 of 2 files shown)


carlwgeorge:
It looks like the only test failures are the same ones that are currently already failing on HEAD.

- `tests/test_main.py::test_output[False-via-sdist-isolation]`
- `tests/test_main.py::test_output[False-wheel-direct-isolation]`

tests/test_env.py
```diff
@@ -257,9 +264,10 @@ def test_venv_creation(
         'pip',
         pytest.param(
             'uv',
-            marks=pytest.mark.xfail(
+            marks=[
                 IS_PYPY and IS_WINDOWS and sys.version_info < (3, 9), reason='uv cannot find PyPy 3.8 executable on Windows'
```
henryiii, line 268:
```suggestion
                pytest.mark.xfail(IS_PYPY and IS_WINDOWS and sys.version_info < (3, 9), reason='uv cannot find PyPy 3.8 executable on Windows'
```

henryiii (approved):

layday:
There’s two ways that we detect the presence of uv and this only checks one. The tests might be skipped even if uv is present (e.g. if it’s on the PYTHONPATH but its executable isn’t on PATH).

henryiii:
Yes, we should add the other way to the check.

PR #809, opened 2024-08-30
tests: handle no-wheel setuptools

henryiii:
Fixing the expectation of tests that wheel is requested (setuptools no longer requests it).


henryiii:
I think I've basically recreated #778 too.

henryiii:
Two remaining issues: The Windows PyPy builds are failing pretty much entirely (50+ failures). And a sampling of jobs have a single "file not found" error coming from setuptools (for files it is writing). In the last week, both PyPy and setuptools have updated, but PyPy 3.9 wasn't included in the update (or 3.8), so I strongly suspect this is due to the setuptools update.

PR #816, opened 2024-09-07
docs: mention conda-forge name in README

henryiii:
Also forgot this.


gaborbernat (approved):

.github/workflows/reusable-docs.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       - name: Setup Python 3.10
```
henryiii, line 14:
```suggestion
      - name: Setup Python 3.12
```

PR #820, opened 2024-09-29
chore: 1.2.2.post1

henryiii:
Just metadata updates, to get the 3.13 classifier on PyPI.

PR #823, opened 2024-10-06
tests: avoid pyproject-hooks 1.2 on tests for now

henryiii:
This is a very specific issue, and I think eliminating 1.2.0 from the full requirements will have more fallout than the bug does. We can add a note for users. Needed for #820 to pass. See https://github.com/pypa/pyproject-hooks/issues/206.


layday:
The failing test doesn't need to depend on setuptools. Can we roll a custom backend with just a `build_wheel` hook?

layday:
#824

PR #824, opened 2024-10-06
tests: fix  under pyproject-hooks 1.2

layday:
By replacing setuptools with a bare-bones wheel builder.

Close #823.


henryiii:
Nothing checks for RECORD, I guess?

henryiii (approved):

layday:
Nope, not in `prepare`.

PR #829, opened 2024-10-15
Revert "tests: fix  under pyproject-hooks 1.2"

henryiii:
Reverts pypa/build#824

Testing fix in https://github.com/pypa/setuptools/pull/4680. See https://github.com/pypa/pyproject-hooks/issues/206.

### DIFF
--- a/tests/packages/test-no-prepare/pyproject.toml
+++ b/tests/packages/test-no-prepare/pyproject.toml
@@ -1,9 +1,4 @@
 [build-system]
 build-backend = 'backend_no_prepare'
 backend-path = ['.']
-requires = []
-
-[project]
-name = "test-no-prepare"
-version = "1.0.0"
-description = "Test extracting metadata from a backend w/out `prepare_metadata_for_build_wheel` hook"
+requires = ['setuptools >= 42.0.0']
--- a/tests/packages/test-no-prepare/backend_no_prepare.py
+++ b/tests/packages/test-no-prepare/backend_no_prepare.py
@@ -1,25 +1,4 @@
 # SPDX-License-Identifier: MIT
 
-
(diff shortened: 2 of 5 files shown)


tests/packages/test-no-prepare/pyproject.toml
```diff
@@ -1,9 +1,4 @@
 [build-system]
 build-backend = 'backend_no_prepare'
 backend-path = ['.']
-requires = []
-
-[project]
-name = "test-no-prepare"
-version = "1.0.0"
-description = "Test extracting metadata from a backend w/out `prepare_metadata_for_build_wheel` hook"
+requires = ['setuptools >= 42.0.0']
```
abravalheri, line 4: Maybe 

```suggestion
requires = ['setuptools @ https://github.com/abravalheri/setuptools/archive/refs/heads/issue-pyproject-hooks-206-take2.zip']
```

Or something similar if the idea is to test the specific fix?

(Or maybe it needs to be added to the test dependencies)

layday: This is indeed tested without isolation.

henryiii: First had to verify that this still fails - which it does :)

henryiii: Got a `error: [Errno 2] No such file or directory: 'test_setuptools-1.0.0/test_setuptools.egg-info'` but I think we used to see that occasionally. Most of the jobs passed, though, so it seems to have fixed it!

abravalheri: Thank you very much @henryiii for testing it!

PR #842, opened 2024-11-20
docs: Add method of installation via conda

GasinAn:
On conda-forge, this package is called "python-build", not "build". However, this information is only given in README.md and not in docs/installation.rst.  Users are likely to read the documentation on https://build.pypa.io/ only, and they are likely to think that the name of the package on conda-forge is also "build", which will make them get into trouble.  Fix this problem here.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -2,6 +2,9 @@
 Installation
 ============
 
+Installing via pip
+==================
+
 ``build`` can be installed via `pip`_ or an equivalent:
 
 .. code-block:: sh
@@ -38,6 +41,15 @@ accepted.
    There is also a ``uv`` extra, which can be used for ``--installer=uv`` if
    you don't have another install of ``uv`` available.
 
+Installing via conda
+====================
+
+On conda-forge, this package is called python-build_, therefore it can be installed via `conda`_ or an equivalent:
+
+.. code-block:: sh
+
+   $ conda install conda-forge::python-build
+
 Bootstrapping
 =============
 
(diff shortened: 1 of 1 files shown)


layday:
Python libraries are typically prefixed with "python" in multi-language package indices, and conda-forge is one of many. I'm not sure we need to mention it specifically.

henryiii:
Conda is a very popular, OS agnostic way to install Python libraries. It comes up all the time when discussing packaging, etc, so I think it would be okay to mention. It's called python-build in the defaults channel too (it's still 0.10 there, though).

It was originally called `build`, but they renamed it and eventually dropped the out-of-date build-feedstock.

henryiii:
Also, _very_ few packages have a `python-` prefix in conda.

layday:
Is it too late to rename this package to `pyproject-build`? 😄

Secrus:
Isn't this information already present in the README?

henryiii:
Yes, it is. Maybe we should try to make sure this is more similar to the readme.

henryiii:
I think it was just an oversight that the same info wasn't added to the docs, in general docs should not have less information than the README. A README is an abbreviated version of the docs, usually.

layday:
We also mention pipx and uv in the readme. Let's add those here too.

layday (approved):

henryiii:
> Let's add those here too.

Agreed. Same PR or followup?

henryiii (approved):

layday:
I don't mind, if @GasinAn would prefer to leave that to somebody else, we can do in a follow-up.

GasinAn:
Sorry, I prefer to leave followup to somebody else, because I know few about pipx and uv.😐

PR #861, opened 2025-02-15
fix: disable use of `pip install --python` for debundled pip

mgorny:
In the debundled pip version distributed on Gentoo systems, `pip install --python` is nonfunctional since it expects all of pip's dependencies to be present in the (empty) virtual environment.  Detect this case by the removal of `pip._vendor` package, and use the "no valid outer pip" codepath.

Bug: https://bugs.gentoo.org/934922

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -158,6 +158,12 @@ class _PipBackend(_EnvBackend):
         This checks for a valid global pip. Returns None if pip is missing, False
         if pip is too old, and True if it can be used.
         """
+        # `pip install --python` is nonfunctional on Gentoo debundled pip
+        try:
+            import pip._vendor
+        except ImportError:
+            return False
+
         # Version to have added the `--python` option.
         return _has_dependency('pip', '22.3')
 
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -166,7 +166,7 @@ class _PipBackend(_EnvBackend):
         with warnings.catch_warnings():
             warnings.simplefilter('ignore')
             if importlib.util.find_spec('pip._vendor') is None:
-                return False
+                return False  # pragma: no cover
 
         # Version to have added the `--python` option.
         return _has_dependency('pip', '22.3')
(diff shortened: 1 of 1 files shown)


mgorny:
Looks like distutils import hack from `setuptools` is causing trouble here. Any suggestion how to deal with that?

henryiii:
I don't think this will fix it, but I'd not do `import` if you aren't going to use it later; use `importlib.util.find_spec("pip._vendor")` instead, it's much faster. To me, isn't it a bit broken if `pip install --python` can't be used? You can target a completely empty environment with `pip install --python` normally.

mgorny:
Yes, it is. Unfortunately, I don't see how to fix it — in the end, the `--python` is a huge hack, and we find unbundling dependencies more important than, well, an option that we didn't even knew existed before `build` started using it.

src/build/env.py
```diff
@@ -158,6 +159,15 @@ def _has_valid_outer_pip(self) -> bool | None:
...
         if pip is too old, and True if it can be used.
         """
+        # `pip install --python` is nonfunctional on Gentoo debundled pip.
+        # Detect that by checking if pip._vendor` module exists.  However,
+        # searching for pip could yield warnings from _distutils_hack,
+        # so silence them.
+        with warnings.catch_warnings():
+            warnings.simplefilter('ignore')
+            if importlib.util.find_spec('pip._vendor') is None:
+                return False
```
henryiii, line 169:
```suggestion
                return False  # pragma: no cover
```

henryiii:
What's the warning, out of curiosity?

mgorny:
> What's the warning, out of curiosity?

Basically this:

```py
>>> import distutils
>>> import pip
/usr/lib/python3.10/site-packages/_distutils_hack/__init__.py:30: UserWarning: Setuptools is replacing distutils. Support for replacing an already imported distutils is deprecated. In the future, this condition will fail. Register concerns at https://github.com/pypa/setuptools/issues/new?template=distutils-deprecation.yml
  warnings.warn(
```

henryiii:
> the `--python` is a huge hack

Pip is full of hacks - the build-isolation support is a mess, for example. IMO using `--python` instead of installing pip into every single environment is much more reasonable, and the main reason people don't use it is because it's a newer feature and it's not the way we used to work. `uv` always works this way, and it makes more sense.

Regardless, though, this seems fine.

mgorny:
Thanks!

PR #870, opened 2025-03-06
tests: fix tox min job

henryiii:
Tests have been broken on main for a bit. See if it can be fixed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,7 +54,7 @@ test = [
   "build[uv, virtualenv]",
   "filelock >= 3",
   "pytest >= 6.2.4",
-  "coverage[toml]>=5.1",
+  "pytest-cov >= 2.12",
   "pytest-mock >= 2",
   "pytest-rerunfailures >= 9.1",
   "pytest-xdist >= 1.34",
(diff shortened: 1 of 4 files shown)


gaborbernat (approved):

layday:
I can't tell right now cuz all the tests are failing, but switching from pytest-cov to coverage will reduce coverage if we've got any tests running in subprocesses, unless we set up [implicit coverage](https://coverage.readthedocs.io/en/7.6.12/subprocess.html#implicit-coverage).

henryiii:
I was just trying to solve the "no coverage" found issue on the min jobs (first commit).

henryiii:
Okay, working, here's a summary of the changes:


Min version test:
* Pinned setuptools_scm for older Pythons as it now errors out with old setuptools
* Increased the pin on packaging on older pythons to make setuptools_scm happy
* Used a simpler way to set the constraints in tox but still have them cleared for subprocess calls

I also had to remove the error on `coverage combine`, as it errors out when there's nothing to combine, and that happens on the min version test since we only do one run in CI.

I also added pip to the deps, as it's required when using tox-uv (which is what I'm using locally).

henryiii:
I think there's an issue with threading where setuptools may fight over the same source directory from multiple threads, causing an occasional error we've seen for a while. But not fixing that here.

PR #871, opened 2025-03-07
tests: better support for parallel testing

henryiii:
I think this will fix our flaky tests. Setuptools makes some files in-tree.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -103,8 +103,14 @@ def packages_path():
 
 def generate_package_path_fixture(package_name):
     @pytest.fixture
-    def fixture(packages_path):
-        return os.path.join(packages_path, package_name)
+    def fixture(packages_path, tmp_path):
+        package_path = os.path.join(packages_path, package_name)
+        if 'flit' in package_name:
+            return package_path
+
+        new_path = tmp_path / package_name
+        shutil.copytree(package_path, new_path)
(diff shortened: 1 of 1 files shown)


layday (approved):

tests/conftest.py
```diff
@@ -103,8 +103,14 @@ def packages_path():
 
 def generate_package_path_fixture(package_name):
     @pytest.fixture
-    def fixture(packages_path):
-        return os.path.join(packages_path, package_name)
+    def fixture(packages_path, tmp_path):
+        package_path = os.path.join(packages_path, package_name)
+        if 'flit' in package_name:
```
layday, line 108: We'll end up copying all but one package this way - we only need to do it for setuptools, right?  That's only `test-bad-wheel` and `test-setuptools` AFAICT.

gaborbernat (approved):

tests/conftest.py
```diff
@@ -12,11 +12,17 @@
...
+from pathlib import Path
 
 import pytest
 
 import build.env
 
+if sys.version_info < (3, 11):
+    import tomli as tomllib
+else:
+    import tomllib
```
layday, line 24:
```suggestion
from build._compat import tomllib
```

layday:
pre-commit.ci autofix

PR #874, opened 2025-03-14
allow editable wheel distributions to be built from the command line

greyltc:
adds a  new argument, `-e`/`--editable`, that can't be combined with `--wheel` or `--sdist`

### DIFF
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -300,7 +300,7 @@ def main_parser() -> argparse.ArgumentParser:
                 This is recommended as it will ensure the sdist can be used
                 to build wheels.
 
-                Pass -s/--sdist and/or -w/--wheel to build a specific distribution.
+                Pass -s/--sdist and/or -w/--wheel or -e/--editable to build a specific distribution.
                 If you do this, the default behavior will be disabled, and all
                 artifacts will be built from {srcdir} (even if you combine
                 -w/--wheel with -s/--sdist, the wheel will be built from {srcdir}).
@@ -348,6 +348,15 @@ def main_parser() -> argparse.ArgumentParser:
         action='append_const',
         const='wheel',
         help='build a wheel (disables the default behavior)',
+    )
+    parser.add_argument(
+        '--editable',
+        '-e',
+        dest='distributions',
+        action='append_const',
+        const='editable',
+        help='build an editable wheel (disables the default behavior). '
+        'Can not be combined with other distribuion build types',
     )
     parser.add_argument(
         '--outdir',
@@ -426,6 +435,8 @@ def main(cli_args: Sequence[str], prog: str | None = None) -> None:
         distributions = ['wheel']
 
     with _handle_build_error():
+        if "editable" in distributions and len(distributions) > 1:
+            _error('Building an editable type distribution can not be combined with any other type.')
         built = build_call(
             args.srcdir,
             outdir,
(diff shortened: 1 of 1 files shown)


henryiii:
> An “editable” wheel uses the wheel format not for distribution but as ephemeral communication between the build system and the front end. This avoids having the build backend install anything directly. This wheel must not be exposed to end users, nor cached, nor distributed.

https://peps.python.org/pep-0660/

layday:
Thank you, but we don’t think this should be part of the CLI.

greyltc:
Shame, I find support for PEP 660 very helpful for development. It would have been handy if this tool could have done that directly.

greyltc:
I guess the workaround is to do it using the API then:
```bash
python -c 'import build; build.ProjectBuilder(".").build("editable", "dist")'
```
I'm still not sure I understand why this API feature should not be exposed by the CLI.

greyltc:
Why: I use build instead of pip to build wheels because I like it more for that purpose. I don't use pip for package management. I don't use pip.
How: If this PR were to be merged, I would use it like this: `python -m build --editable`

henryiii:
What do you do with the editable wheel once you build it? Don't you pip install it?

The spec says users should never see or interact with the wheel. "It must not be exposed to end users" is not ambiguous. It's simply a mechanism to allow a tool like pip to implement editable installs. Build doesn't install things, it produces wheels. We would be producing a wheel that the spec says "must not be exposed".

If you are writing a tool, like pip, then our API lets you make editable wheels that you can use to implement installing editable installs (without exposing it to your users).

FFY00:
Editable wheels are meant to be **ephemeral**, this is a key aspect of the design. This effectively means nothing other than code should touch it.

Per its design, and the constraints set by the spec, it should never be exposed in something like a CLI, so that's not something we will do here, sorry 🫤

> Does the spec make any determination between user and developer? Is a "user" just a human?

This is developer tool, the end-users here are the package developers.

> To answer your question, I think I do what any normal dev would do with an editable wheel they've created: I make a virtual environment, activate it and then use `python -m installer` on the wheel.

That is a perfect example of what the spec wanted to prevent... and if you notice, it works only in **your** specific use-case — you either don't have any dependencies, or you are relying on them being injected via `--system-site-packages`.

*sighs*

Look, I understand what you are asking works for your use-case, and I understand the frustration of us not being willing to add a feature that, to you, probably seems fairly simple and innocuous. But we, as a community, have explicitly chosen to design the tooling this way, which wasn't easy — it probably took over a thousand replies in [discuss.python.org](https://discuss.python.org/) alone — and we, as the maintainers, should at minimum respect that.

If you don't want to use a Python-specific package manager, like pip, or uv, my recommendation would be for you to write a simple script using the pypa/build and pypa/installer API to install projects in editable mode. Alternatively, you can also look for a backend that provides something like it (eg. `setup.py develop` from setuptools).

greyltc:
>  you are relying on them being injected via --system-site-packages

That's exactly what I'm doing. I'm managing the dependencies myself, isn't that okay? For example, I might run
```
hatch project metadata | jq -r '.dependencies | .[]'
```
inspect the deps, and install them with my OS' package manager which maintains them properly system-wide. Maybe if there's some conflict or specific older version that's a hard requirement, I might choose to install the dependency in the venv, but I try to avoid that whenever possible.

Should we remove the `--editable` cli option[1] from pip? How can it be that `--editable` can be blessed there but not here?

\[1]: https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-e

PR #880, opened 2025-03-31
chore: use dependency-groups

henryiii:
Use dependency-groups, drop the `[test]`/`[docs]`/`[typing]` extras.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,15 +1,12 @@
 version: 2
 
-sphinx:
-  configuration: docs/conf.py
 
 build:
-  os: ubuntu-20.04
+  os: ubuntu-22.04
   tools:
-    python: "3.10"
-
-python:
-  install:
-    - method: pip
-      path: .
-      extra_requirements: [docs]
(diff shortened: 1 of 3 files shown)


henryiii:
pre-commit.ci autofix

layday:
pip supports dependency groups now - can we use it?

henryiii:
Sure, but where? We are using tox so I don't think we need it anywhere? But yes, `--group` in uv or pip will work.

henryiii:
If you mean in readthedocs, uv is faster. Upgrading pip then pip installing with a group would work, but would be slower.

layday:
Sorry, I did mean in RTD. Just seems slightly odd to rely on asdf here - I'd prefer if there was a "blessed" way to use uv. If there isn't, maybe we should stick to pip?

henryiii:
asdf is the recommendation from the RtD team, and that's what they use to set things up too. If you request python, it gets installed with asdf. They do have caching for some modules (not uv), but it's only a second or two, and cuts the docs time in more than half usually.

PR #891, opened 2025-05-06
feat: drop Python 3.8

henryiii:
Dropping 3.8.


henryiii:
We need to remove `--no-wheel` and make a patch release before we can drop 3.8, since the virtualenv that produces a warning with `--no-wheel` was released for 3.8. Will keep on draft till then.

henryiii:
Actually, I'm not sure that's true, since 3.8 still has wheel, so it should not trigger the warning. We probably can drop at the same time.

henryiii:
Sure. :) If you want to add pypy3.11 while you are at it, you can, but fine if you want to leave that up to us eventually, too.

jayaddison:
Ah; my mistake - because PyPy v3.9 _is_ still supported, there are some older runtime versions of that could be affected by the `pypy3323` bug -- e.g. `pypy3.9-v7.3.8-win64.zip`.

henryiii:
If there's any version with 3.9 that is fine (pypy3.9-v7.3.16?), I'd be tempted to simply say that we don't support anything older than whatever that version is (7.3.16 or 7.3.11?). People should not be using old PyPy versions that are unsupported; if we can say "please update to 7.3.16 to fix the bug", I think that's fine. Depends on the opinion of other maintainers, though.

PR #892, opened 2025-05-06
fix: virtualenv no-wheel

henryiii:
Adapting to the new `--no-wheel` deprecation. See https://github.com/pypa/virtualenv/pull/2868.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -76,7 +76,8 @@ uv = [
   "uv >= 0.1.18",
 ]
 virtualenv = [
-  "virtualenv >= 20.0.35",
+  "virtualenv >= 20.11; python_version < '3.10'",
+  "virtualenv >= 20.17; python_version >= '3.10'",
 ]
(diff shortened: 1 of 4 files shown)


layday:
Are there any plans to actually remove this option or will it forever emit a deprecation warning?

henryiii:
This also bumps the min versions of virtualenv to something we can test (20.11 for Python 3.9 and older, 20.17 for 3.10+), and restored testing on older virutalenv's for the `-min` jobs.

henryiii:
#891 should also be merged before a release, as I didn't add logic to restore this on 3.8, which still supports `--no-wheel`. So this is slightly more wasteful on 3.8.

henryiii:
Also, I'll have to fix coverage in a follow up. The `main` coverage is locked to an old commit, we need to add a token to get it working on `main` again. But this needs fixing to get our CI working again (minus coverage).

PR #898, opened 2025-05-21
fix: Fix `ModuleNotFoundError` when `pip` is not installed

mgorny:
Fix the exception raised if `pip` is not installed at all:

```
[...]
  File "/usr/lib/python3.14/site-packages/build/env.py", line 168, in _has_valid_outer_pip
    if importlib.util.find_spec('pip._vendor') is None:
       ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "<frozen importlib.util>", line 90, in find_spec
ModuleNotFoundError: No module named 'pip'
```

This is a regression from #861 where I did not predict that `find_spec()` will actually raise when the parent package is not available. To fix it, just reorder the conditions to make sure that `pip` is installed first.

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -161,7 +161,7 @@ class _PipBackend(_EnvBackend):
         """
 
         # Version to have added the `--python` option.
-        if not _has_dependency('pip', '22.3'):
+        if not _has_dependency('pip', '22.3'):  # pragma: no cover
             return False
 
         # `pip install --python` is nonfunctional on Gentoo debundled pip.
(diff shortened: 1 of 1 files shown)


henryiii:
Ah, I think this was a regression in an unreleased change? Best kind of fix. :)

gaborbernat (approved):

mgorny:
> Ah, I think this was a regression in an unreleased change? Best kind of fix. :)

Yeah, fortunately I've backported it to Gentoo and managed to notice while we were doing the Python 3.14 testing. Sorry about it and thanks for the quick merge!

PR #899, opened 2025-05-21
chore: remove conftest from coverage

henryiii:
This keeps reporting as 0% on PRs. I don't really think covering test files is
important, and especially conftest.py. We have static checks that detect common
problems.


pyproject.toml
```diff
@@ -96,7 +96,9 @@ exclude = ["**/__pycache__", "docs/_build", "**/*.egg-info", "tests/packages/*/b
 source = [
   "build",
   "tests",
+
```
henryiii, line 99:
```suggestion
```

PR #916, opened 2025-08-01
feat: add --config-json

### LINKED ISSUE #900: Feature proposal: --config-json
I'd like to propose adding a feature from `gpep517`: `--config-json`. This is a replacement for `--config-setting`/`-C` that allows specification of the configure settings as a JSON object in a string. This is more flexible than the `-C` version, both allowing advanced use cases that are allowed by the PEP but not by most tools, as well as being ideal for piping through from tools. PEP 517 allows any dict here, but pip/build/uv have not taken advantage of that (expect for build's API, which does). cibuildwheel could likely use this, mapping its TOML config settings into this option. It also would be useful for testing, as currently backends like scikit-build-core either have to use the Python API or a `gpep517` to test things that are beyond a simple string mapping. (I know about this tool because someone noticed that bools weren't working in config-settings, so I asked how they managed that and they said they were using `gpep517`).

Design:

* I think using JSON is a simple, cross-language string format. Another option would be Python via literal_eval, but I think JSON is more universal across languages.
* I'm fine with the name `--config-json` from `gpep517`, but I don't really have opinions here, happy with whatever.
* I think taking a string is simpler than taking a filename, especially for scripting (which I assume is the main use case).
* What happens when you specify both needs to be specified. It could be an error, or we could merge, with one overriding the other. Not sure which should override which; I'd be fine with an error if both specified TBH.

I'd like some feedback from pip/uv maintainers, since if we add this here, it is conceivable that other frontends might eventually be expected to add it too, especially if backends start expecting nested structures or bool/ints. As well as pip is also a frontend choice for cibuildwheel, and I'm adding native uv soon too.

Example using cibuildwheel:

```toml
[tool.cibuildwheel.config-settings]
build.targets = ["ext1", "ext2"]
```

could transform to

```bash
python -m build --config-json '{"build": {"targets": ["ext1", "ext2"}}'
```

While today it has to be:

```toml
[tool.cibuildwheel.config-settings]
"build.targets" = ["ext1", "ext2"]
```

which is expanded as

```bash
python -m build --config-setting=build.targets=ext1 --config-setting=build.targets=ext2
```

(scikit-build-core has to expand dots manually to mimic the full nested structure of the TOML config, as anything you can pass in TOML you can pass as `--config-setting` except top-level arrays)

henryiii:
Close #900. See that issue for rationale.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,6 @@ $ uvx --from build pyproject-build --installer uv
 
 If you are using [cibuildwheel][], build is integrated and the default builder in 3.0+. If you want to use `uv` as the installer, you can use:
 
-
 ```toml
 [tool.cibuildwheel]
 build-frontend = "build[uv]"
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -95,12 +95,12 @@ ANSI_STRIP = re.compile(r'\x1B(?:[@-Z\\-_]|\[[0-?]*[ -/]*[@-~])')
         ),
         (
             ['--config-json={"one": 1, "two": [2, 3], "three": {"in": "out"}}'],
-            [cwd, out, ['wheel'], {"one": 1, "two": [2, 3], "three": {"in": "out"}}, True, False, None],
+            [cwd, out, ['wheel'], {'one': 1, 'two': [2, 3], 'three': {'in': 'out'}}, True, False, None],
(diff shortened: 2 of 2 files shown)


gaborbernat (changes requested):
Looks good, but you'll need to fix the CI 😊

PR #963, opened 2025-12-12
Allow setting build constraints

layday:


layday:
Still not sure if I'm entirely onboard with this or if we should allow passing arbitrary (or checked) `--installer-args` - or neither.

PR #970, opened 2026-01-02
docs: first argument for _has_dependency is a name

abitrolly:
Fixed function description. It would be nice to get some usage example, how version string looks like. 

EDIT: The code has this example.
```sh
_has_dependency('pip', '22.3')
```
Which checks if `pip >= 22.3`  is installed.

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -41,7 +41,7 @@ class IsolatedEnv(typing.Protocol):
 
 def _has_dependency(name: str, minimum_version_str: str | None = None, /, **distargs: object) -> bool | None:
     """
-    Given a package name, see if it is present and return True if the version is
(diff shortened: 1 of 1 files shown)


henryiii (approved):

src/build/env.py
```diff
@@ -41,7 +41,7 @@ def make_extra_environ(self) -> Mapping[str, str] | None:
 
 def _has_dependency(name: str, minimum_version_str: str | None = None, /, **distargs: object) -> bool | None:
     """
-    Given a path, see if a package is present and return True if the version is
+    Given a package name, see if it is present and return True if the version is
```
```suggestion
    Given a distribution name, see if it is present and return True if the version is
```

To avoid confusion with importable packages.

PR #971, opened 2026-01-02
fix: support UV environment variable

### LINKED ISSUE #967: Respect `UV`
If uv runs a process, it sets (since 0.6) a `UV` variable pointing at itself. We should respect that if set, probably here: https://github.com/pypa/build/blob/ed9c379f51abca7c0c0ae20bee66aa548d597dc3/src/build/env.py#L307.

Might help with https://github.com/pypa/cibuildwheel/pull/2691.

henryiii:
Fix #967.

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -304,7 +304,7 @@ class _UvBackend(_EnvBackend):
 
             self._uv_bin = uv.find_uv_bin()
         except (ModuleNotFoundError, FileNotFoundError):
-            uv_bin = shutil.which(os.environ.get("UV", "") or 'uv')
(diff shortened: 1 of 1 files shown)


src/build/env.py
```diff
@@ -304,7 +304,7 @@ def create(self, path: str) -> None:
 
             self._uv_bin = uv.find_uv_bin()
         except (ModuleNotFoundError, FileNotFoundError):
-            uv_bin = shutil.which('uv')
+            uv_bin = shutil.which(os.environ.get("UV", "") or 'uv')
```
```suggestion
            uv_bin = shutil.which(os.environ.get('UV') or 'uv')
```

PR #1021, opened 2026-04-10
chore: lazy imports

henryiii:
- **chore: use TYPE_CHECKING=False trick**
- **chore: enforce TYPE_CHECKING=False trick**
- **chore: add lazy imports**

## Description

This adds lazy imports.

```console
$ hyperfine --warmup 10 \
    -n FullLazy ".venv/bin/python -X lazy_imports=all    -m build -h" \
    --reference ".venv/bin/python -X lazy_imports=normal -m build -h" \
    -n Eager    ".venv/bin/python -X lazy_imports=none   -m build -h"
...
  .venv/bin/python -X lazy_imports=normal -m build -h ran
    1.31 ± 0.12 times slower than FullLazy
    1.10 ± 0.11 times faster than Eager
```

I used `uvx flake8-lazy —-apply src/**/*.py` to get the lazy lists.

## Changelog

## Checklist

### DIFF
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -16,7 +16,6 @@ import textwrap
 import traceback
 import warnings
 
-from collections.abc import Iterator, Mapping, Sequence
 from functools import partial
 from typing import Any, NoReturn, TextIO
 
@@ -27,8 +26,15 @@ import build
 from . import ProjectBuilder, _ctx
 from . import env as _env
 from ._exceptions import BuildBackendException, BuildException, FailedProcessError
-from ._types import ConfigSettings, Distribution, StrPath, SubprocessRunner
 from .env import DefaultIsolatedEnv
+
+
+TYPE_CHECKING = False
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator, Mapping, Sequence
+
+    from ._types import ConfigSettings, Distribution, StrPath, SubprocessRunner
 
 
 _COLORS = {
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -53,7 +53,7 @@ TYPE_CHECKING = False
 if TYPE_CHECKING:
     from collections.abc import Iterator, Mapping, Sequence
 
-    from ._types import ConfigSettings, Distribution, StrPath, SubprocessRunner
+    from build._types import ConfigSettings, Distribution, StrPath, SubprocessRunner
 
 
 _COLORS = {
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -2,6 +2,25 @@
 
 from __future__ import annotations
 
+
+__lazy_modules__ = [
+    'argparse',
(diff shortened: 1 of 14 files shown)


src/build/__main__.py
```diff
@@ -16,19 +35,25 @@
...
 from typing import Any, NoReturn, TextIO
 
 import pyproject_hooks
 
 import build
+import build.env as _env
+
+from build import ProjectBuilder, _ctx
+from build._exceptions import BuildBackendException, BuildException, FailedProcessError
+from build.env import DefaultIsolatedEnv
```
layday, line 48: Why were these imports made absolute?

src/build/__main__.py
```diff
@@ -16,19 +35,25 @@
...
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator, Mapping, Sequence
 
-from . import ProjectBuilder, _ctx
-from . import env as _env
-from ._exceptions import BuildBackendException, BuildException, FailedProcessError
-from ._types import ConfigSettings, Distribution, StrPath, SubprocessRunner
-from .env import DefaultIsolatedEnv
+    from ._types import ConfigSettings, Distribution, StrPath, SubprocessRunner
```
layday: The actual types get imported in `__init__` AFAICT so there's nothing to be gained from lazifying these.

layday:
Visually, this is just awful. Can maintenance be automated?  Is ruff able to detect extraneous entries?

What if we just flipped on lazy imports globally in `__main__`?

src/build/__main__.py
henryiii: Because one of them already was, and `__main__.py` modues have `__spec__`s that can be None so mypy complains.

src/build/__main__.py
henryiii: The rules for the imports are in flake-lazy. Basically anything that is unused at top level is lazy.

henryiii:
Yes, I wrote a tool for maintenance, it’s flake8-lazy. Being intelligent with this is tricky, and there are a few edge cases, such as the `from an import b` syntax being ambiguous (I assume B is not a module). I would hope it does make its way to ruff in some form.

There is a way to make all imports lazy by adding three lines, but it ideally needs to be in every file.

src/build/__main__.py
layday: I meant tucking it under an if type checking block, sorry for being unclear.

henryiii: I added the TC code, that’s supposed to catch everything that can be moved to type checking blocks.

layday: Okay, so this is for any type checking only imports we might add in the future to go under?  Currently, we don’t have any that’d benefit from this.

src/build/__main__.py
henryiii: I mention this under "tips" at https://iscinumpy.dev/post/flake8-lazy/.

gaborbernat (approved):

PR #1022, opened 2026-04-10
chore: adding more ruff codes

henryiii:
## Description

Adding more ruff codes. The commented ones are ones I'd like to add eventually, but will need some changes.

Based on the output of `uvx --from sp-repo-review[cli] sp-ruff-checks .`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -207,15 +207,36 @@ extend-select = [
   "YTT",    # flake8-2020
   "TRY",    # tryceratops
   "EM",     # flake8-errmsg
+  "DTZ",    # flake8-datetimez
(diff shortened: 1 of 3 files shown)


src/build/_builder.py
```diff
@@ -15,6 +15,8 @@
 
 import pyproject_hooks
 
+from typing_extensions import Self
```
layday: We don't depend on typing_extensions at runtime.

henryiii: Yes, I accidentally added this from testing the currently commented codes.

gaborbernat (approved):

PR #1020, opened 2026-04-09
tests: fully annotate

henryiii:
## Description

Used copilot CLI and gpt-5-mini to add types to the test suite.

## Changelog

- [ ] Added changelog fragment: `docs/changelog/<pr_number>.<type>.rst`
  - Types: `feature`, `bugfix`, `doc`, `removal`, `misc`
  - Example: `123.feature.rst` containing ``Add custom backend support - by :user:`yourname` ``

## Checklist

- [ ] Tests pass locally (`tox`)
- [ ] Code follows project style (`tox -e fix`)
- [ ] Type checks pass (`tox -e type`)
- [ ] Documentation builds (`tox -e docs`)

PR #1023, opened 2026-04-10
tests: improve annotations

henryiii:
## Description

Followup to #1020.

Used copilot CLI and gpt-5-mini to make some changes.

### DIFF
--- a/tests/test_projectbuilder.py
+++ b/tests/test_projectbuilder.py
@@ -174,7 +174,7 @@ class NestedCircularMockDistribution(MockDistribution):
         ('circular_dep', None),
     ],
 )
-def test_check_dependency(monkeypatch: pytest.MonkeyPatch, requirement_string: str, expected: Any) -> None:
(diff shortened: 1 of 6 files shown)


tests/test_projectbuilder.py
```diff
@@ -198,25 +198,25 @@ def test_init(
     test_no_permission: str,
     package_test_bad_syntax: str,
 ) -> None:
-    mocker.patch('pyproject_hooks.BuildBackendHookCaller')
+    mock_buildcaller = mocker.patch('pyproject_hooks.BuildBackendHookCaller', autospec=True)
 
     # correct flit pyproject.toml
     builder = build.ProjectBuilder(package_test_flit)
-    typing.cast(Any, pyproject_hooks.BuildBackendHookCaller).assert_called_with(
+    mock_buildcaller.assert_called_with(
```
henryiii, line 205: FWIW, I'm not sure why `autospec=True` fixes these type annotations, but it does not work with `.side_effect` below.

gaborbernat (approved):

PR #1024, opened 2026-04-10
🐛 fix(env): strip PYTHONPATH from isolated builds

### LINKED ISSUE #405: PYTHONPATH affects isolated builds
# Introduction
I've found that setting the `PYTHONPATH` environment variable affects isolated builds, to the point of breaking functionality in some cases. I'll describe steps to reproduce, and propose a fix, though I should disclaim that my experience with both this package and the `PYTHONPATH` environment variable are limited, so I might be missing something. This ticket follows from the assumption that setting the `PYTHONPATH` environment variable should not affect build behavior (as seems to be confirmed by https://github.com/pypa/build/issues/373#issuecomment-939545616). If this assumption is wrong, the ticket is invalid, though I'd be interested to know why this is the case.

# Environment
Python 3.6.15. The same steps produce an error for 3.9 as well, but it is a different one (similar to the error in #266) so I'll create a separate ticket for that if it is agreed upon that this is a bug.

# Steps to reproduce
```sh
# clone the sample project
git clone https://github.com/pypa/sampleproject.git
cd sampleproject
# create a new virtual environment
python3.6 -m venv .env
.env/bin/pip install -U pip build
# build the sample project with the env's libs dir in PYTHONPATH
PYTHONPATH=.env/lib/python3.6/site-packages/ .env/bin/python -m build --sdist
```

# Expected behavior
An sdist is built

# Observed behavior
```
* Creating venv isolated environment...
* Installing packages in isolated environment... (setuptools>=40.8.0, wheel)

Usage:   
  /tmp/build-env-a0xlzsq3/bin/python -m pip install [options] <requirement specifier> [package-index-options] ...
  /tmp/build-env-a0xlzsq3/bin/python -m pip install [options] -r <requirements file> [package-index-options] ...
  /tmp/build-env-a0xlzsq3/bin/python -m pip install [options] [-e] <vcs project url> ...
  /tmp/build-env-a0xlzsq3/bin/python -m pip install [options] [-e] <local project path> ...
  /tmp/build-env-a0xlzsq3/bin/python -m pip install [options] <archive url/path> ...

no such option: --use-pep517

  File "/home/sander/documents/projects/sampleproject/.env/lib/python3.6/site-packages/build/__main__.py", line 373, in main
    args.srcdir, outdir, distributions, config_settings, not args.no_isolation, args.skip_dependency_check
  File "/home/sander/documents/projects/sampleproject/.env/lib/python3.6/site-packages/build/__main__.py", line 202, in build_package
    out = _build(isolation, builder, outdir, distribution, config_settings, skip_dependency_check)
  File "/home/sander/documents/projects/sampleproject/.env/lib/python3.6/site-packages/build/__main__.py", line 140, in _build
    return _build_in_isolated_env(builder, outdir, distribution, config_settings)
  File "/home/sander/documents/projects/sampleproject/.env/lib/python3.6/site-packages/build/__main__.py", line 108, in _build_in_isolated_env
    env.install(builder.build_system_requires)
  File "/home/sander/documents/projects/sampleproject/.env/lib/python3.6/site-packages/build/env.py", line 211, in install
    _subprocess(cmd)
  File "/home/sander/documents/projects/sampleproject/.env/lib/python3.6/site-packages/build/env.py", line 81, in _subprocess
    raise e
  File "/home/sander/documents/projects/sampleproject/.env/lib/python3.6/site-packages/build/env.py", line 78, in _subprocess
    subprocess.check_output(cmd, stderr=subprocess.STDOUT)
  File "/usr/lib/python3.6/subprocess.py", line 356, in check_output
    **kwargs).stdout
  File "/usr/lib/python3.6/subprocess.py", line 438, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['/tmp/build-env-a0xlzsq3/bin/python', '-Im', 'pip', 'install', '--use-pep517', '--no-warn-script-location', '-r', '/tmp/build-reqs-ont40sbg.txt']' returned non-zero exit status 2.

ERROR Command '['/tmp/build-env-a0xlzsq3/bin/python', '-Im', 'pip', 'install', '--use-pep517', '--no-warn-script-location', '-r', '/tmp/build-reqs-ont40sbg.txt']' returned non-zero exit status 2.
```

# Likely cause
`IsolatedEnvBuilder` sets up an isolated environment to perform the build using `create_isolated_env_venv` (`virtualenv` is not installed for this scencario). Package installation within this environment is then handled by `_IsolatedEnvVenvPip.install`, which passes Python's `-I` flag to ensure isolation: https://github.com/pypa/build/blob/90dbd8bd71e510c18b0a472e72f9e3503ee5a90f/src/build/env.py#L200-L209
However, this same flag is absent during setup: `create_isolated_env_venv` performs a `pip install pip` (simplified) to ensure a sufficiently recent `pip` is installed:
https://github.com/pypa/build/blob/90dbd8bd71e510c18b0a472e72f9e3503ee5a90f/src/build/env.py#L275
As a result, the `pip install pip` during the set up of the environment still takes the `PYTHONPATH` environment variable into account, deciding that a matching `pip` is already installed. As a result the outdated `pip` in the isolated environment is not updated, which causes the failure seen above.

# Proposed fix
Add the `-I` flag to both `pip install` commands in `create_isolated_env_venv`. I've confirmed that this resolves this issue.

gaborbernat:
Closes #405

When `PYTHONPATH` is set in the host environment, it leaks into isolated build subprocesses. The build backend and package installers then discover packages from the host instead of the isolated venv, breaking the isolation contract. 🔒 This causes hard-to-diagnose build failures when the host has incompatible package versions — for example, an old pip that lacks `--use-pep517`.

The fix clears `PYTHONPATH` in `make_extra_environ()` so all backend hook calls via pyproject_hooks strip it from their subprocess environment. The uv backend's install path, which explicitly passes `os.environ`, also gets the same treatment. Internal pip calls already use `-Im` (isolated mode) which ignores `PYTHONPATH`, so they were unaffected.

Setting `PYTHONPATH` to an empty string (rather than removing it) ensures the key is present in the environment dict and overrides whatever value the host had, even when the runner merges it with `os.environ.copy()`. Non-isolated builds (`--no-isolation`) are unaffected since they don't go through `DefaultIsolatedEnv`.

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -154,7 +154,11 @@ class DefaultIsolatedEnv(IsolatedEnv):
         return {
             'PATH': os.pathsep.join([self._env_backend.scripts_dir, path])
             if path is not None
-            else self._env_backend.scripts_dir
+            else self._env_backend.scripts_dir,
+            # Set PYTHONPATH to empty to override any host value. An empty
+            # PYTHONPATH is treated as unset by CPython's path initialization
+            # (the ``if pythonpath_env:`` check makes it a no-op).
+            'PYTHONPATH': '',
         }
 
     def install(self, requirements: Collection[str], constraints: Collection[str] = []) -> None:
@@ -383,7 +387,9 @@ class _UvBackend(_EnvBackend):
 
                 cmd += ['-c', os.path.abspath(constraint_file.name)]
 
-            run_subprocess(cmd, env={**os.environ, 'VIRTUAL_ENV': self._env_path})
+            env = {k: v for k, v in os.environ.items() if k != 'PYTHONPATH'}
+            env['VIRTUAL_ENV'] = self._env_path
+            run_subprocess(cmd, env=env)
 
     @property
     def display_name(self) -> str:
--- a/docs/changelog/405.bugfix.rst
+++ b/docs/changelog/405.bugfix.rst
@@ -1,0 +1,2 @@
+Strip ``PYTHONPATH`` from the environment during isolated builds to prevent host packages from leaking into the build -
+by :user:`gaborbernat`
--- a/tests/test_env.py
+++ b/tests/test_env.py
@@ -28,6 +28,25 @@ MISSING_UV = importlib.util.find_spec('uv') is None and not shutil.which('uv')
 MISSING_VIRTUALENV = importlib.util.find_spec('virtualenv') is None
 
 
+def test_make_extra_environ_overrides_pythonpath() -> None:
+    with build.env.DefaultIsolatedEnv() as env:
+        extra = env.make_extra_environ()
+        assert extra['PYTHONPATH'] == ''
+        assert env._env_backend.scripts_dir in extra['PATH']
+
+
+def test_uv_install_strips_pythonpath(
+    mocker: pytest_mock.MockerFixture,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    monkeypatch.setenv('PYTHONPATH', '/some/leaky/path')
+    run_subprocess = mocker.patch('build.env.run_subprocess')
+    with build.env.DefaultIsolatedEnv(installer='uv') as env:
+        env.install(['some-package'])
+    (install_call,) = run_subprocess.call_args_list
+    assert 'PYTHONPATH' not in install_call.kwargs['env']
+
+
 @pytest.mark.isolated
 def test_isolation() -> None:
     subprocess.check_call([sys.executable, '-c', 'import build.env'])


src/build/env.py
```diff
@@ -149,7 +149,8 @@ def make_extra_environ(self) -> dict[str, str]:
         return {
             'PATH': os.pathsep.join([self._env_backend.scripts_dir, path])
             if path is not None
-            else self._env_backend.scripts_dir
+            else self._env_backend.scripts_dir,
+            'PYTHONPATH': '',
```
henryiii, line 161: Wouldn't it be better to unset it then set it to an empty string?

src/build/env.py
```diff
@@ -104,6 +104,9 @@ def __init__(
         self.installer: Installer = installer
 
     def __enter__(self) -> DefaultIsolatedEnv:
+        # Strip PYTHONPATH for the duration of the isolated build to prevent
+        # the host environment from leaking into build subprocesses.
+        self._old_pythonpath = os.environ.pop('PYTHONPATH', None)
```
henryiii: Do we care about thread safety here? Modifying os.environ will affect other threads. I thought the previous version of the fix didn't do this?

gaborbernat: Yeah, we should probably care 🤔

henryiii: I think the previous version of the fix, but stripping out this variable from a copy of the environment, would be best. :)

gaborbernat:
Updated the approach per your feedback. The fix now:

1. **`make_extra_environ()`** returns `PYTHONPATH: ''` — this overrides the host value when pyproject_hooks' runner does `os.environ.copy().update(extra_environ)`. An empty `PYTHONPATH` is treated as unset by CPython's path initialization (`if pythonpath_env:` is falsy for empty string), so no paths are added to `sys.path`.

2. **`_UvBackend.install_dependencies()`** builds the env dict by filtering out `PYTHONPATH` entirely (it passes `env=` directly to subprocess, so true removal is possible here).

No `os.environ` mutation, fully thread-safe. The `PYTHONPATH=''` approach through `extra_environ` is the only way to override the value without bypassing the pyproject_hooks runner (which handles stdout/stderr display for verbosity control).

henryiii:
Okay, that's reasonable. Maybe could be a future thing to do in pyproject_hooks.

henryiii (approved):

PR #1026, opened 2026-04-10
chore: add ruff PYI

henryiii:
## Description

Adding the PYI check.

### DIFF
--- a/src/build/_util.py
+++ b/src/build/_util.py
@@ -6,7 +6,8 @@ import re
 TYPE_CHECKING = False
 
 if TYPE_CHECKING:
-    from collections.abc import Iterator, Set
+    from collections.abc import Iterator
+    from collections.abc import Set as AbstractSet
 
 
 _WHEEL_FILENAME_REGEX = re.compile(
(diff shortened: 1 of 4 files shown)


gaborbernat (approved):

src/build/_util.py
```diff
@@ -6,7 +6,8 @@
 TYPE_CHECKING = False
 
 if TYPE_CHECKING:
-    from collections.abc import Iterator, Set
+    from collections.abc import Iterator
+    from collections.abc import Set as AbstractSet
```
layday, line 10: Why `Set as AbstractSet`?

henryiii: It's a ruff rule to keep it from looking like dict/Dict or list/List, which are the same thing, but just typing versions. While `Set` is not a `set`, it's an immutable set-like ABC. `frozenset` is also a valid `Set`. It's more like `Sequence`, etc.

Not my favorite rule (as `Dict`/`List` and similar classes become less common, it's less important), but it's fine. (BTW, let me know if you know another name vs. AbstractSet, this is just the suggestion).

layday: Ah, I didn't realise this was part of the ruleset.  That's fair enough then, let's leave it be.

PR #1028, opened 2026-04-10
chore: add ruff SIM/RET

henryiii:
## Description

Adding RET and SIM codes.

AI usage disclaimer: the last four changes didn't have an auto fix, so I used copilot CLI and gpt-5-mini to fix the ruff errors.

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -59,6 +59,9 @@ def get_project(name: str, tmp_path: Path) -> Path:
 
     # for other projects download from github and cache it
     tar_store = os.path.join(ROOT, '.integration-sources')
+    # Checking with exists is not parallel safe so just ignore,
+    # if the creation failed we will have another failure soon
+    # that will notify the user.
     with contextlib.suppress(OSError):
         os.makedirs(tar_store)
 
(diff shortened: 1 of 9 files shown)


tests/test_main.py
```diff
@@ -663,11 +662,9 @@ def test_logging_output_env_subprocess_error(
     stderr_body: list[str],
     stderr_error: str,
 ) -> None:
-    try:
+    with contextlib.suppress(ModuleNotFoundError):  # colorama might not be available  # pragma: win32 no cover
```
henryiii:
```suggestion
    with contextlib.suppress(ModuleNotFoundError):  # colorama might not be available
```

tests/test_integration.py
```diff
@@ -58,10 +59,8 @@ def _ignore_folder(base: str, filenames: list[str]) -> list[str]:
 
     # for other projects download from github and cache it
     tar_store = os.path.join(ROOT, '.integration-sources')
-    try:
+    with contextlib.suppress(OSError):
```
henryiii, line 65:
```suggestion
   # Checking with exists is not parallel safe so just ignore,
   # if the creation failed we will have another failure soon
   # that will notify the user.
    with contextlib.suppress(OSError):
```

Why is there a comment about Python 2 here anyway?

gaborbernat (approved):

PR #1029, opened 2026-04-10
chore: use ruff ALL

henryiii:
## Description

Use Ruff's ALL setting, opting out instead of opting in.

AI usage disclaimer: no AI used.

### DIFF
--- a/src/build/_compat/importlib.py
+++ b/src/build/_compat/importlib.py
@@ -7,15 +7,14 @@ TYPE_CHECKING = False
 
 if TYPE_CHECKING:
     import importlib_metadata as metadata
-else:
-    if sys.version_info >= (3, 10, 2):
-        from importlib import metadata  # pragma: no cover
-    else:  # pragma: no cover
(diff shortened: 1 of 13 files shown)


src/build/_compat/importlib.py
```diff
@@ -7,15 +7,14 @@
...
-else:
-    if sys.version_info >= (3, 10, 2):
-        from importlib import metadata  # pragma: no cover
-    else:  # pragma: no cover
-        try:
-            import importlib_metadata as metadata
-        except ModuleNotFoundError:
-            # helps bootstrapping when dependencies aren't installed
-            from importlib import metadata
+elif sys.version_info >= (3, 10, 2):
```
gaborbernat, line 10: beter to add branch pragms from covdefaults

henryiii: `covdefaults` doesn't support three digit checks, only two.

gaborbernat (approved):

PR #1033, opened 2026-04-13
chore: tox toml

henryiii:
## Description

This is an attempt to convert the tox.ini to tox.toml. I started
by pointing copilot at the tox documentation for configuration.
I then asked it to use a generative list, telling it it was okay to reverse
the order `...-min` to `min-...`.

Still need to:

- [x] Check the minimum tox version this requires (update: used last tox-uv that supported 3.9)
- [ ] Consider putting it in pyproject.toml

🤖 I'm manually using Linux's `Assisted-by` trailer for comments. Used `Copilot:claude-sonnet-4.6`

## Changelog

## Checklist


tox.toml
```diff
@@ -0,0 +1,116 @@
+requires = ["tox>=4.22", "tox-uv", "virtualenv>=20.0.34"]
```
gaborbernat: Probably can remove both tox and virtualenv here. And pin tox-uv.

tox.toml
```diff
@@ -0,0 +1,116 @@
...
+env_list = [
+    "fix",
+    "type",
+    "docs",
+    "path",
+    { product = [["py314", "py313", "py312", "py311", "py310", "py39", "pypy311", "pypy310", "pypy39"]] },
```
gaborbernat: Can we move to use `3.x` instead?

henryiii: Sure, I think it's a fine time to modernize.

tox.toml
```diff
@@ -0,0 +1,116 @@
...
+    { product = [["min"], ["py314", "py313", "py312", "py311", "py310", "py39", "pypy311", "pypy310", "pypy39"]] },
+]
+skip_missing_interpreters = true
+
+[env_run_base]
+description = "run test suite with {basepython}"
+deps = ["pip"]
```
gaborbernat: We should not add this, instead set `uv_seed`.

tox.toml
```diff
@@ -0,0 +1,116 @@
...
+commands = [
+    ["-", "coverage", "combine", "{toxworkdir}"],
+    ["coverage", "report", "--skip-covered", "--show-missing", "-i"],
+    ["coverage", "xml", "-o", "{toxworkdir}/coverage.xml", "-i"],
+    ["coverage", "html", "-d", "{toxworkdir}/htmlcov", "-i"],
+    ["python", "-m", "diff_cover.diff_cover_tool", "--compare-branch", "{env:DIFF_AGAINST:origin/main}", "{toxworkdir}/coverage.xml"],
+]
+depends = [
+    "path",
+    "py314", "py313", "py312", "py311", "py310", "py39", "pypy311", "pypy310", "pypy39",
```
gaborbernat: can use range expansion here

tox.toml
gaborbernat: can use range expansion here

tox.toml
henryiii: If we are on a platform that doesn't have wheels, I'm guessing that would mean it just wouldn't work. But we are already requiring tox-uv, so it's that way now.

henryiii: How tightly should it be pinned? Lower + upper bound?

gaborbernat: Lower is fine.

.github/workflows/reusable-pytest.yml
```diff
@@ -100,4 +88,4 @@ jobs:
           files: .tox/coverage.xml
           flags: tests
           env_vars: PYTHON
-          name: ${{ matrix.py }} - ${{ matrix.os }}
+          name: ${{ matrix.py }} - ${{ matrix.os }} - ${{ matrix.tox-target }}
```
henryiii, line 92: Unrelated change, but seems like this is needed, otherwise two jobs have the same name.

tox.toml
```diff
@@ -0,0 +1,119 @@
...
+
+[env_run_base.set_env]
+COVERAGE_CORE = "sysmon"
+COVERAGE_FILE = "{toxworkdir}/.coverage.{envname}"
+PYPY3323BUG = "1"
+PYTHONWARNDEFAULTENCODING = "1"
+TEST_STATUS_DIR = "{envtmpdir}"
+
+[env_base.min]
+factors = [["3.9", "3.10", "3.11", "3.12", "3.13", "3.14", "pypy3.9", "pypy3.10", "pypy3.11"]]
```
henryiii, line 36: I don't think I can use range here.

Also, I don't like that we have to make this list so many times, it would be nice if I could reuse it somehow. But maybe it's easy enough to update with AI. `¯\_(ツ)_/¯`

abitrolly:
@henryiii did tox get a decent CLI UX with the new config format recently? Like show the commands I can run without learning about tox.ini and stuff?

gaborbernat:
`tox l` always did that 🤔

henryiii:
Yes, `tox list` (or I guess `tox l`, that's a bit closer in typing to `nox -l`, I guess!) is great.

abitrolly:
It takes sooo long. I couldn't wait.

```sh
# time tox l
ROOT: will run in automatically provisioned tox, host /usr/local/bin/python3.14 is missing [requires (has)]: tox-uv
ROOT: install_deps> python -I -m pip install tox tox-uv 'tox>=4.22' 'virtualenv>=20.0.34'
^C
real	0m8.066s
user	0m1.767s
sys	0m0.184s
```

It appears it can not just run itself to show help and needs to create virtualenv. Or no, the rabbit hole. I am not going to get down there. :D
```
subcommands:
  tox command to execute (by default legacy)

  {run,r,run-parallel,p,depends,de,man,list,l,devenv,d,schema,config,c,quickstart,q,exec,e,legacy,le}
    run (r)                   run environments
    run-parallel (p)          run environments in parallel
```
I made a fix to remove that argparse "feature" 10 years ago, but didn't post it anywhere.

gaborbernat:
> It appears it can not just run itself to show help and needs to create virtualenv.

Mean, you have an old outdated version of the tool that doesn't support the new features; so needs to install a newer version of itself following https://tox.wiki/en/4.53.0/explanation.html#auto-provisioning, the speed here will depend a lot on your network.

abitrolly:
@gaborbernat it was `pip install tox` in a clear Python container.
```
Downloading tox-4.53.0-py3-none-any.whl (212 kB)
```
But I can't repeat that anymore. I guess it was its local state in mounted dir that had to be updated.

tox.toml
```diff
@@ -0,0 +1,119 @@
...
+
+[env_base.min]
+factors = [["3.10", "3.11", "3.12", "3.13", "3.14", "3.15", "pypy3.10", "pypy3.11"]]
+description = "check minimum versions required of all dependencies"
+set_env = { PIP_CONSTRAINT = "{toxinidir}/tests/constraints.txt", UV_CONSTRAINT = "{toxinidir}/tests/constraints.txt" }
+
+[env.fix]
+description = "run static analysis and style checks"
+skip_install = true
+deps = ["prek"]
```
layday: Can we use dep groups for all deps? This one could be called "lint".

tox.toml
```diff
@@ -0,0 +1,119 @@
...
+commands = [
+    ["python", "-m", "pip", "list", "--format=columns"],
+    ["python", "-c", "import sys; print(sys.executable)"],
+]
+dependency_groups = ["docs", "test"]
+
+[env.coverage]
+description = "combine coverage from test environments"
+skip_install = true
+deps = ["coverage[toml]>=5.1", "diff_cover>=3"]
```
layday: As above.

tox.toml
```diff
@@ -0,0 +1,119 @@
...
+    { product = [{ prefix = "3.", start = 10, stop = 15 }]},
+    { product = [{ prefix = "pypy3.", start = 10, stop = 11 }]},
+    { product = [{ prefix = "min-3.", start = 10, stop = 15 }]},
+    { product = [{ prefix = "min-pypy3.", start = 10, stop = 11 }]},
+]
+
+[env.bump]
+description = "bump versions, pass major/minor/patch"
+skip_install = true
+deps = ["bump-my-version>=0.10"]
```
layday: Ditto.

layday (approved):

gaborbernat (approved):

PR #1036, opened 2026-04-13
feat: drop 3.9, require 3.10+

henryiii:
## Description

Pip is about to drop 3.9, so we probably can too. I don't think there are any pressing fixes we need to get out for 3.9.

- **feat: drop Python 3.9 support**
- **chore: more 3.10 changes the model missed**
- **chore: use pattern matching in a few places**

:robot: Used Copilot:gpt-5-mini for the first commit.

## Changelog

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -220,6 +220,11 @@ ignore = [
 [tool.ruff.lint.flake8-tidy-imports.banned-api]
 "typing.TYPE_CHECKING".msg = "Use TYPE_CHECKING=False instead"
 "os.path.commonprefix".msg = "Use os.path.commonpath or manually implement string manip"
+"typing.Callable".msg = "Use collections.abc.Callable instead."
+"typing.Iterator".msg = "Use collections.abc.Iterator instead."
+"typing.Mapping".msg = "Use collections.abc.Mapping instead."
+"typing.Sequence".msg = "Use collections.abc.Sequence instead."
+"typing.Set".msg = "Use collections.abc.Set instead."
 
 [tool.ruff.lint.per-file-ignores]
 "tasks/**.py" = ["T20", "INP", "S607"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,7 @@ build-backend = "flit_core.buildapi"
 name = "build"
 description = "A simple, correct Python build frontend"
 readme = "README.md"
-requires-python = ">= 3.9"
+requires-python = ">= 3.10"
 license = "MIT"
 authors = [
   { name = "Filipe Laíns", email = "lains@riseup.net" },
@@ -17,7 +17,7 @@ authors = [
 classifiers = [
   "Programming Language :: Python :: 3",
   "Programming Language :: Python :: 3 :: Only",
-  "Programming Language :: Python :: 3.9",
+
   "Programming Language :: Python :: 3.10",
   "Programming Language :: Python :: 3.11",
   "Programming Language :: Python :: 3.12",
@@ -176,7 +176,7 @@ filterwarnings = [
 [tool.mypy]
 files = ["src", "tests"]
 exclude = ["tests/packages"]
-python_version = "3.9"
+python_version = "3.10"
 strict = true
 enable_error_code = ["ignore-without-code", "truthy-bool", "redundant-expr"]
 warn_unreachable = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -220,11 +220,6 @@ ignore = [
 [tool.ruff.lint.flake8-tidy-imports.banned-api]
 "typing.TYPE_CHECKING".msg = "Use TYPE_CHECKING=False instead"
 "os.path.commonprefix".msg = "Use os.path.commonpath or manually implement string manip"
-"typing.Callable".msg = "Use collections.abc.Callable instead."
-"typing.Iterator".msg = "Use collections.abc.Iterator instead."
-"typing.Mapping".msg = "Use collections.abc.Mapping instead."
-"typing.Sequence".msg = "Use collections.abc.Sequence instead."
-"typing.Set".msg = "Use collections.abc.Set instead."
 
 [tool.ruff.lint.per-file-ignores]
 "tasks/**.py" = ["T20", "INP", "S607"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,7 +49,6 @@ keyring = [
   "keyring",
 ]
 virtualenv = [
-  "virtualenv >= 20.11; python_version < '3.10'",
   "virtualenv >= 20.17; python_version >= '3.10' and python_version < '3.14'",
   "virtualenv >= 20.31; python_version >= '3.14'",
 ]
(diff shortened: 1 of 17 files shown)


gaborbernat (approved):

src/build/__main__.py
```diff
@@ -105,15 +105,17 @@ def _make_logger() -> _ctx.Logger:
...
-                initial_indent = '> ' if kind[1] == 'cmd' else '< '
-                for line in message.splitlines():
-                    _cprint('{dim}{}{reset}', fill(line, initial_indent=initial_indent), file=sys.stderr)
+            match kind:
+                case ('step', *_):
+                    _cprint('{bold}{}{reset}', fill(message, initial_indent='* '), file=sys.stderr)
+                case ('subprocess', 'cmd'):
+                    for line in message.splitlines():
+                        _cprint('{dim}{}{reset}', fill(line, initial_indent='> '), file=sys.stderr)
+                case ('subprocess', _):
```
layday:
```suggestion
                case ('subprocess', 'stdout'):
```

src/build/_builder.py
```diff
@@ -114,21 +114,24 @@ def _parse_build_system_table(pyproject_toml: Mapping[str, Any]) -> Mapping[str,
...
+            msg = '`build-backend` must be a string'
+            raise BuildSystemTableValidationError(msg)
+        case _:
+            _find_typo(build_system_table, 'build-backend')
+            # If ``build-backend`` is missing, inject the legacy setuptools backend
+            # but leave ``requires`` intact to emulate pip
+            build_system_table['build-backend'] = _DEFAULT_BACKEND['build-backend']
+
+    match build_system_table:
+        case {'backend-path': list() as lst} if all(isinstance(i, str) for i in lst):
```
layday:
```suggestion
        case {'backend-path': list() as backend_path} if all(isinstance(i, str) for i in backend_path):
```

src/build/_types.py
```diff
@@ -6,16 +6,14 @@
 
 __all__ = ['ConfigSettings', 'Distribution', 'StrPath', 'SubprocessRunner']
 
-ConfigSettings = typing.Mapping[str, typing.Union[str, typing.Sequence[str]]]
+ConfigSettings = typing.Mapping[str, str | typing.Sequence[str]]
```
layday:
```suggestion
ConfigSettings = collections.abc.Mapping[str, str | collections.abc.Sequence[str]]
```

src/build/_types.py
```diff
@@ -6,16 +6,14 @@
...
 
 TYPE_CHECKING = False
 
 if TYPE_CHECKING:
     from pyproject_hooks import SubprocessRunner
 else:
-    SubprocessRunner = typing.Callable[
-        [typing.Sequence[str], typing.Optional[str], typing.Optional[typing.Mapping[str, str]]], None
-    ]
+    SubprocessRunner = typing.Callable[[typing.Sequence[str], str | None, typing.Mapping[str, str] | None], None]
```
layday:
```suggestion
    SubprocessRunner = collections.abc.Callable[[collections.abc.Sequence[str], str | None, collections.abc.Mapping[str, str] | None], None]
```

layday (changes requested):

src/build/__main__.py
henryiii: Nice, also I made a mistake here, I meant to do `(subprocess, *_)`, but if it's always this, that's better.

src/build/_types.py
henryiii: Ohh, I could probably make Ruff require this.

henryiii:
I'm going to put this in draft until the question on the direct url PR gets resolved - if we need a patch release, it's a bit easier if this isn't in yet. ;)

(Addressing review soon).

henryiii:
I added the banned API to Ruff, and verified that telling copilot gpt-5-mini to run and fix `prek -a` (0x token usage, if you are curious as to why that old model!) does produce the same output (but merged your suggestion instead, just saving my pyproject.toml edits).

I like being able to turn nearly any check report into an autofix. :)

src/build/__main__.py
```diff
@@ -105,15 +105,17 @@ def _make_logger() -> _ctx.Logger:
...
-                initial_indent = '> ' if kind[1] == 'cmd' else '< '
-                for line in message.splitlines():
-                    _cprint('{dim}{}{reset}', fill(line, initial_indent=initial_indent), file=sys.stderr)
+            match kind:
+                case ('step', *_):
+                    _cprint('{bold}{}{reset}', fill(message, initial_indent='* '), file=sys.stderr)
+                case ('subprocess', 'cmd'):
+                    for line in message.splitlines():
+                        _cprint('{dim}{}{reset}', fill(line, initial_indent='> '), file=sys.stderr)
+                case ('subprocess', 'stdout'):
```
henryiii: Is this actually 
```suggestion
                case ('subprocess', 'stdout' | 'stderr'):
```
?

layday: We used to have `stderr` but we merged it into `stdout` because it wasn't possible to keep the original line order with two streams.

layday: Ah, you are right, `log_subprocess_error` does still read from/emit to `stderr` - good catch.

henryiii: I asked gpt-5-mini and it said we still have the stderr case (I'm asking because the tests are failing and I think it's this case), it found it here:

https://github.com/pypa/build/blob/d0bb805cb4ffca56b7dc07dc25f2e3ffa9f6d9b9/src/build/_ctx.py#L42-L45

henryiii: Maybe at runtime they are merged but tests still has both?

henryiii: It says that the verbose path can't hit it (merged streams), but the non-verbose path (verbosity <= 0) does hit it, as we read them from the error and the error has both separately.

tests/packages/test-metadata/pyproject.toml
```diff
@@ -11,4 +11,4 @@ description = 'hello!'
 [tool.black]
 line-length = 127
 skip-string-normalization = true
-target-version = ['py39', 'py38', 'py37', 'py36']
+target-version = ['py310', 'py38', 'py37', 'py36']
```
layday, line 14: Let's just remove this whole section?

layday:
A couple of < 3.10 constraints in tests/constraints.txt can also go.

tests/packages/test-metadata/pyproject.toml
henryiii: Oh, wow. This is old.

pyproject.toml
```diff
@@ -222,6 +220,11 @@ ignore = [
 [tool.ruff.lint.flake8-tidy-imports.banned-api]
 "typing.TYPE_CHECKING".msg = "Use TYPE_CHECKING=False instead"
 "os.path.commonprefix".msg = "Use os.path.commonpath or manually implement string manip"
+"typing.Callable".msg = "Use collections.abc.Callable instead."
+"typing.Iterator".msg = "Use collections.abc.Iterator instead."
+"typing.Mapping".msg = "Use collections.abc.Mapping instead."
+"typing.Sequence".msg = "Use collections.abc.Sequence instead."
+"typing.Set".msg = "Use collections.abc.Set instead."
```
gaborbernat: How come we change these?

henryiii: Because those are old, deprecated locations for them that added generics, but the standard library versions are generic in 3.10+. However, I think there might actually be a built-in Ruff check of those now.

Ah, yes, `UP035 deprecated-import` catches these.

gaborbernat (approved):
Not sure if the match over the if is better but I'm not against it.

layday (approved):

Issue #1037, opened 2026-04-14
Error "Backend 'X' is not available" when PYTHONPATH contains a build dependency


AWhetter:
# Problem

I have a situation where  PYTHONPATH is set such that it contains a directory that has a build dependency in it, and running `python -m build` fails with the error "Backend 'X' is not available".

As an example:

```bash
$ cat <<EOF > pyproject.toml
[build-system]
requires = ["setuptools"]
build-backend = "setuptools.build_meta"

[project]
name = "myemptypkg"
version = "0.1.0"
description = "Nothing"
EOF
$ python -m venv .venv
$ .venv/bin/pip install --upgrade pip build
$ .venv/bin/pip list
Package         Version
--------------- -------
build           1.4.3
packaging       26.0
pip             26.0.1
pyproject_hooks 1.2.0
setuptools      65.5.1
$ PYTHONPATH=.venv/lib/python3.11/site-packages .venv/bin/python -m build
* Creating isolated environment: venv+pip...
* Installing packages in isolated environment:
  - setuptools
* Getting build dependencies for sdist...

Traceback (most recent call last):
  File "/u/asw/temp/build_test/.venv/lib/python3.11/site-packages/pyproject_hooks/_impl.py", line 402, in _call_hook
    raise BackendUnavailable(
pyproject_hooks._impl.BackendUnavailable: Cannot import 'setuptools.build_meta'

ERROR Backend 'setuptools.build_meta' is not available.
```

Note that the use of pip >=22.3 is important to trigger this code path: https://github.com/pypa/build/blob/d0bb805cb4ffca56b7dc07dc25f2e3ffa9f6d9b9/src/build/env.py#L218

# Explanation of the issue

This comes about because the command to install the build dependencies has PYTHONPATH available to it, via `_PipBackend.install_dependencies` (https://github.com/pypa/build/blob/d0bb805cb4ffca56b7dc07dc25f2e3ffa9f6d9b9/src/build/env.py#L357). This differs from the environment that is used to execute `pyproject_hooks`, where PYTHONPATH is unset by `DefaultIsolatedEnv.make_extra_environ`.

Here's the tracebacks of where the two different subprocesses occur:
```
* Creating isolated environment: venv+pip...
* Installing packages in isolated environment:
  - setuptools
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
[...]

PR #1038, opened 2026-04-14
fix: PYTHONPATH only partially removed

henryiii:
## Description

Remove the `PYTHONPATH` setting from `_pip_env`. From the "possible solution", I think this should help #1037?

:robot: Assisted-by: OpenCode:Kimi-K2.5 - used to quickly implement the suggestion and a few tests, then iterated by hand on this to just remove vs. override.

## Changelog

## Checklist

### DIFF
--- a/docs/changelog/1038.bugfix.rst
+++ b/docs/changelog/1038.bugfix.rst
@@ -1,1 +1,3 @@
-Keep ``PYTHONPATH`` out of the return for ``_pip_env`` - by :user:`henryiii`
+Keep ``PYTHONPATH`` out of the return for ``_pip_env``; also exclude ``PYTHONPATH``-only paths when detecting a usable
+outer pip, so environments where pip is only reachable via ``PYTHONPATH`` (e.g. Nix dev shells) no longer attempt to
+invoke it with ``PYTHONPATH`` cleared - by :user:`henryiii`
--- a/docs/changelog/1038.bugfix.rst
+++ b/docs/changelog/1038.bugfix.rst
@@ -1,0 +1,1 @@
+Keep ``PYTHONPATH`` out of the return for ``_pip_env`` - by :user:`henryiii`
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -196,10 +196,14 @@ def _has_keyring_cli() -> bool:
     return shutil.which('keyring') is not None
 
 
-def _pip_env() -> dict[str, str] | None:
+def _pip_env() -> dict[str, str]:
+    env = os.environ.copy()
+    # Clear PYTHONPATH to avoid interference from the outer environment.
+    # The isolated build environment should not inherit PYTHONPATH.
+    env.pop('PYTHONPATH', None)
     if 'PIP_KEYRING_PROVIDER' not in os.environ and _has_keyring_cli():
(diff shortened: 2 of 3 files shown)


gaborbernat (approved):

gaborbernat (approved):

henryiii:
Once @layday reviews, this is good to merge, and we can have a quick patch release.

layday:
What happens if the outer pip is on the Python path? `_has_dependency` respects `PYTHONPATH`.

This will always be the case e.g. in a Nix dev shell.

henryiii:
As long as its path gets resolved correctly, shouldn't it be fine? You don't been PYTHONPATH set to call pip.

layday:
You do if pip itself is on `PYTHONPATH`.  This is with Nix:

```
$ echo $PYTHONPATH
/nix/store/dfkim8vahgdqhk1lrxz1j4mcighlrk99-python3-3.14.3/lib/python3.14/site-packages:/nix/store/yic25g8alpkdjyfkbim9741cpnj8yin8-python3.14-pip-25.3/lib/python3.14/site-packages
$ python -m pip -V
pip 25.3 from /nix/store/yic25g8alpkdjyfkbim9741cpnj8yin8-python3.14-pip-25.3/lib/python3.14/site-packages/pip (python 3.14)
$ PYTHONPATH= python -m pip -V
/nix/store/dfkim8vahgdqhk1lrxz1j4mcighlrk99-python3-3.14.3/bin/python: No module named pip
```

`_PipBackend._has_valid_outer_pip` will be true but when you try to run it it won't be there.

henryiii:
Ahh, we do use `python -m pip`, you are correct.

Should we just revert the PYTHONPATH changes in the last release?

Or we could disable PYTHONPATH when looking for pip. But it looks like we use `importlib.metadata.distributions`, so that not be easy.

We could also just refuse to look for pip if PYTHONPATH is set.

layday:
> Should we just revert the PYTHONPATH changes in the last release?

Meaning #1024?  It looks like it isolates the backend runner, but the linked issue (#405) is about `PYTHONPATH` leaking into `ensurepip`, which was fixed in https://github.com/python/cpython/pull/92857.

> Or we could disable PYTHONPATH when looking for pip.

How do we do that other than by attempting to parse and subtract `PYTHONPATH` from `sys.path` or by spawning a subprocess without `PYTHONPATH` and performing the dependency check there? Neither option sounds particularly appealing to me.

We could just try passing `--ignore-installed` to pip, but how does that interact with `--python`?

henryiii:
I've not used that before, we could try? Hopefully it's not new so we don't get bit with even more version-specific behavior. :) (Edit: looks like it's old, good)

henryiii:
Coverage is a bit short, but I tried something different. I tried to create a couple of tests to mimic this (with AI help), and implemented its suggestion as a fix.

For `--ignore-installed`, I think it's a net positive (as nothing should be there if we are making an isolated environment), but couldn't that still have issues with leakage, if something overrides or adds to the environment?

layday:
Can you be a bit more specific about `--ignore-installed`?

henryiii:
Oh, I think I was thinking about usage too, yeah, that might work. I'll make an alternate PR with that.

PR #1043, opened 2026-04-20
docs: reorganize testing docs for copy/paste

abitrolly:
## Description

Make it easier to scan and copy/paste the needed command. I forgot how to run tests, and my reading skills were all time low today.

## Changelog

## Checklist

### DIFF
--- a/docs/development/contributing.rst
+++ b/docs/development/contributing.rst
@@ -61,7 +61,7 @@ It is also possible to run only the integrations tests in parallel.
 
     tox -- -n auto --only-integration
 
-The test matrix coverts all supported Python versions and implementations, and with the module being invoked directly
+The test matrix covers all supported Python versions and implementations, and with the module being invoked directly
 from path, sdist install, or wheel install. To run tests only for Python 3.9.
 
 .. code-block:: console
--- a/docs/development/contributing.rst
+++ b/docs/development/contributing.rst
@@ -49,32 +49,27 @@ check. To run tests we use ``tox``.
 
     tox
 
-Integration tests take a long time to run, so they are disabled by
-default. Passing either ``--run-integration`` or ``--only-integration`` arguments through ``tox`` to ``pytest`` will run
-them, where the latter will disable unit tests and only run integration tests. Even though these tests are disabled by
-default, they will be run in CI where test suite run durations are not a big issue. CI still runs both test suites.
+Tests run in parallel by default, but if you pass any arguments, you need to include ``-n auto`` if you want to keep
+parallel runs.
 
-.. code-block:: console
-
-    tox -- --run-integration
-
-It is also possible to run only the integrations tests in parallel.
+Integration tests take a long time to run, so they are disabled by default. Passing either ``--run-integration`` or
+``--only-integration`` arguments through ``tox`` to ``pytest`` will run them, where the latter will disable unit tests
+and only run integration tests. CI still runs both test suites.
 
 .. code-block:: console
 
     tox -- -n auto --only-integration
 
 The project has a fairly large environment matrix, running tests for all supported Python versions and implementations,
-and with the module being invoked directly from path, sdist install, or wheel install. To run tests only for Python 3.9.
-
-.. code-block:: console
-
-    tox -e py39
-and with the module being invoked directly from path, sdist install, or wheel install. To run tests only for Python 3.14:
+and with the module being invoked directly from path, sdist install, or wheel install. To run tests only for Python
+3.14:
 
 .. code-block:: console
 
     tox -e py314
+
+and with the module being invoked directly from path, sdist install, or wheel install.
+
 Additionally, there are environments for type checking and documentation building, plus extras like checking code with
 minimum versions of dependencies. For type checking,
 
@@ -82,7 +77,7 @@ minimum versions of dependencies. For type checking,
 
     tox -e type
 
-You can also run unit tests against a specific Python version with wheel installation using ``tox -e py39-wheel``. Code
+You can also run unit tests against a specific Python version with wheel installation using ``tox -e py314-wheel``. Code
 coverage is tracked to ensure all code paths are tested. Aim for complete coverage of any new code you add. The CI
 system will report coverage metrics on your pull request and runs the test suite across all supported operating systems.
 
--- a/docs/development/contributing.rst
+++ b/docs/development/contributing.rst
@@ -41,27 +41,41 @@ issues using GitHub's issue linking syntax.
  Running Tests
 ***************
 
-Due to its nature, build has a somewhat complex test suite with two sets of tests: unit tests and integration tests.
-Unit tests verify the actual code implementation, while integration tests run build on real world projects as a sanity
-check. Integration tests take a long time to run and are not very helpful for tracking down issues, so they are disabled
-by default. They can be enabled by passing either ``--run-integration`` or ``--only-integration`` arguments to pytest,
-where the latter will disable unit tests and only run integration tests. Even though these tests are disabled by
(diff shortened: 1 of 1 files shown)


gaborbernat (changes requested):
I disagree with removing existing information. I understand that I would prefer if we extended it, not clean it up. Your addition here should be strictly additive.

abitrolly:
@gaborbernat I think I was careful enough to compress info. but not remove. I removed explanation like ``--only-integration` means only integration tests are run", because it is logical if not obvious. And merged examples directly into text that previously explained it. Actually I needed only copy/pasteable examples on one page to quickly fix tests while I have time. So feel free to format how do you see fit as long as the code blocks are here. The PR allows edits for maintainers, so we should not spend too much time here.

By the way, did you use any AI to make https://github.com/pypa/build/pull/988 ? It is pretty awesome piece of change that I would personally find trouble doing by hand. But I still haven't figure out how to run agents securely.

henryiii:
What do you mean by "securely? If you mean access to things on your computer, then you can run in a container (I haven't done that yet, just been manually okaying operations), or GitHub's got various mechanisms to run agents (see copilot-instructions.yml, or codespaces).

For private code, then I think you'd need to run locally; I don't really do much that's not open source, so not something I've had to worry about yet. I've been playing with Qwen 3.6 locally and have had issue with it mis-formatting tool calls and paths. (It's pretty smart, it just randomly loses characters; it usually iterates and fixes it, but it's really annoying and makes it take 3x longer). I'm thinking it might be an issue with a small model + OpenCode's 10K token preamble, I'm tempted to try Pi instead, which has a 1K preamble. I also have access to Kimi-K2.5 on nrp.ai, which is really nice (open weight!), but not small enough to run locally. I'm planning on trying Gemma 4 soon. Kimi-K2.5 feels pretty close to GitHub Copilot + claude-haiku-4.5, maybe even a bit above it.

(I think that particular PR looked like it was assisted by Claude Code, at a guess)

henryiii:
To me, it looks pretty similar in content, with just some lost text for the introduction that should be pretty easy to simply put back in.

abitrolly:
I've heard some scary stories about agents following rogue instructions while doing lookups on the internet, and sending out private keys and other sensitive data. Also masking instructions with Unicode symbols to make them invisible. Also the statements like that Aider code is 80% written by AI, and also projects like `beads` + GasTown + Wasteland that run completely on autopilot using repository credentials make me feel really uncomfortable. Especially when code reviews are also done by agents. `llama.cpp` that exposes API through a web-server that is, well, probably written in C++? I already [run all things in a container](https://github.com/yakshaveinc/linux/blob/master/bin/runin-podman.sh), but I've also heard that containers also have attack vectors, and with every layer of security I just fall into a trap of analysis paralysis.

henryiii:
I've used Kimi-K2.6 to restore missing information while keeping the original structure intact (that was basically my prompt!). Maybe that's better? Might be okay to trim a bit.

docs/development/contributing.rst
```diff
@@ -41,26 +41,45 @@ issues using GitHub's issue linking syntax.
...
+    tox -e py39
+
+Additionally, there are environments for type checking and documentation building, plus extras like checking code with
+minimum versions of dependencies. For type checking,
+
+.. code-block:: console
+
+    tox -e type
+
+You can also run unit tests against a specific Python version with wheel installation using ``tox -e py39-wheel``. Code
```
henryiii: This isn't in a code block, but I think it's fine (is a specific example).

Also all these will change after #1033.

abitrolly: Other code blocks are there, so it is automatic LGTM for me.

> Might be okay to trim a bit.

It always good to ask LLMs to be "impeccable with words", because for commercial "pay by token" proposals there is an incentive to generate verbose answers that will convert into more tokens in a follow-up chat. 

Humans also have to spend time and energy to read stuff, that's why they [skim, don't read](https://www.theguardian.com/commentisfree/2018/aug/25/skim-reading-new-normal-maryanne-wolf). So for those of us who already feel overwhelmed with info, less is more.

docs/development/contributing.rst
```diff
@@ -41,26 +41,45 @@ issues using GitHub's issue linking syntax.
...
-installation using ``tox -e py39-wheel``.
-
-Code coverage is tracked to ensure all code paths are tested. Aim for complete coverage of any new code you add. The CI
+check. To run tests we use ``tox``.
+
+.. code-block:: console
+
+    tox
+
+Integration tests take a long time to run and are not very helpful for tracking down issues, so they are disabled by
```
layday:
```suggestion
Integration tests take a long time to run, so they are disabled by
```

Depends on the kind of issue - doesn't really add value.

docs/development/contributing.rst
```diff
@@ -41,26 +41,45 @@ issues using GitHub's issue linking syntax.
...
+default. Passing either ``--run-integration`` or ``--only-integration`` arguments through ``tox`` to ``pytest`` will run
+them, where the latter will disable unit tests and only run integration tests. Even though these tests are disabled by
+default, they will be run in CI where test suite run durations are not a big issue. CI still runs both test suites.
```
layday: This section is repetitive. There's the commands, the elaboration, and a final sentence which restates the second half of the previous sentence.

docs/development/contributing.rst
```diff
@@ -41,26 +41,45 @@ issues using GitHub's issue linking syntax.
...
+
+.. code-block:: console
+
+    tox -- --run-integration
+
+It is also possible to run only the integrations tests in parallel.
```
layday: It's possible to run all the tests in parallel - why are we singling out integration tests?

docs/development/contributing.rst
```diff
@@ -41,26 +41,45 @@ issues using GitHub's issue linking syntax.
...
+
+    tox -- -n auto --only-integration
+
+The project has a fairly large environment matrix, running tests for all supported Python versions and implementations,
+and with the module being invoked directly from path, sdist install, or wheel install. To run tests only for Python 3.9.
+
+.. code-block:: console
+
+    tox -e py39
+
```
layday, line 70:
```suggestion
and with the module being invoked directly from path, sdist install, or wheel install. To run tests only for Python 3.14:

.. code-block:: console

    tox -e py314
```

We're about to drop support for 3.9.

layday (changes requested):
I agree with the general sentiment that the docs are too verbose and visually indistinct. A few suggestions below.

docs/development/contributing.rst
henryiii: Good tip! FWIW, I'm using a open source model, and I don't think it's necessarily trying to optimize for more tokens, it's just trying to keep the original text while also keeping the new structure (which is exactly what I asked it to do).

docs/development/contributing.rst
henryiii: Fine, but as soon as we do, we are going to rework this to be `tox run -e 3.14`, so it's going to change again.

docs/development/contributing.rst
henryiii: I've removed some. Also removed the an example or two that was just a repeat.

docs/development/contributing.rst
henryiii: I've redone this to only include the example with parallel - the test suite already runs in parallel if you don't include flags, but as soon as you do, that removes the parallel running. So updated to say that.

gaborbernat (approved):

docs/development/contributing.rst
layday: Understood.

layday (approved):

PR #1049, opened 2026-04-21
tests: keep environment from leaking in Python 3.15

### LINKED ISSUE #1047: test_isolation fails when PYTHONPATH is set (exposed by __lazy_modules__ on Python 3.15)
test_isolation calls subprocess.check_call without env=, so the parent's PYTHONPATH leaks into the isolated venv. The test expects import build.env to fail inside the venv, but it succeeds because build is reachable via PYTHONPATH.                                        
                                                                                             
This wasn't visible before 1.4.3 because eager imports in build/__init__.py would pull in pyproject_hooks, which isn't in the venv, causing the import to fail for a different reason. The __lazy_modules__ added in 1.4.3 defers that import on Python 3.15, unmasking the PYTHONPATH leak.

Suggested fix: use make_extra_environ() (which already clears PYTHONPATH) for the subprocess:

```                                                                                          
  @@ -3,6 +3,7 @@                                           
   import importlib.util                                                                     
   import logging                                                                            
  +import os                                                                                 
   import shutil                                                                             
   import subprocess                                        
  @@ -52,7 +53,8 @@                                                                          
       debug = 'import sys; import os; print(os.linesep.join(sys.path));'                    
       with build.env.DefaultIsolatedEnv() as env,                                           
  pytest.raises(subprocess.CalledProcessError):                                              
  -        subprocess.check_call([env.python_executable, '-c', f'{debug} import build.env']) 
  +        isolated_env = {**os.environ, **env.make_extra_environ()}                         
  +        subprocess.check_call([env.python_executable, '-c', f'{debug} import build.env'], 
  env=isolated_env)                                                                          
```

henryiii:
## Description

Fix #1047.

:robot: Use a bit of copilot with both gpt-5-mini and Kimi-K2.6, mostly to automate running and appying the example. Hand wrote most of the config changes and some of the code changes.

## Changelog

## Checklist


henryiii:
@copilot investigate the CI failure.

layday:
I feel both of these tests are too generic to be meaningful. What are we testing, that a venv is a venv? `env.python_executable != sys.executable` is probably sufficient.

henryiii:
I can probably combine the tests. The first one is the original one, with the added protection from #1047. The second is the same test but with an injected PYTHONPATH to verify it fails regardless of the external PYTHONPATH setting - but I think we could just add that to the first test.

henryiii:
I believe this issue wouldn't show up if we just checked `env.python_executable != sys.executable`. This demonstrates that you need to apply the extra env. If you remove it, the test fails.

layday:
The issue only exists in the test. Whether we guard against PYTHONPATH in this one test or not has no impact on the actual running of the tool, so I don’t think it matters much either way.

henryiii:
It shows that we need to guard it in the code itself (yes, this is way too unit-testy of a unit-test, testing the internals by mimicking the internals instead of using them!)

If we want to drop the test or change it, I think that should be a followup. This only fixes the current test to work on 3.15, and adds 3.15 to our CI.

layday (approved):

PR #1053, opened 2026-04-27
Replace kind=('...',) with kind='...'

abitrolly:
## Description

To improve readability of log calls.

No AI is used (unfortunately).

## Changelog

## Checklist

### DIFF
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -103,18 +103,19 @@ def _make_logger() -> _ctx.Logger:
 
     fill = partial(textwrap.fill, subsequent_indent='  ', width=max_terminal_width)
 
-    def log(message: str, *, kind: tuple[str, ...] | None = None) -> None:
+    def log(message: str, *, kind: str | None = None) -> None:
         if _ctx.verbosity >= -1:
-            match kind:
-                case ('step', *_):
+            split = kind.split('-') if kind else None
+            match split:
+                case ['step']:
                     (first, *rest) = message.splitlines()
                     _cprint('{bold}{}{reset}', fill(first, initial_indent='* '), file=sys.stderr)
                     for line in rest:
                         print(fill(line, initial_indent='  '), file=sys.stderr)  # noqa: T201
-                case ('subprocess', 'cmd'):
+                case ['subprocess', 'cmd']:
                     for line in message.splitlines():
                         _cprint('{dim}{}{reset}', fill(line, initial_indent='> '), file=sys.stderr)
-                case ('subprocess', 'stdout' | 'stderr'):
+                case ['subprocess', 'stdout' | 'stderr']:
                     for line in message.splitlines():
                         _cprint('{dim}{}{reset}', fill(line, initial_indent='< '), file=sys.stderr)
                 case _:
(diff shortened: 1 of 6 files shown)


layday:
I don't find this to be an improvement from a readability nor from a CQ perspective.

abitrolly:
@layday I have no idea what `CQ perspective` is. Google says it is some kind "cultural intelligence". No sure if that's how you look at this, but the cognitive load from visually parsing that braces and commas are in correct order are much higher that just scanning a quoted string.

henryiii:
CQ: Code Quality. I'd agree, a tuple beats a string, and we can improve the static typing for a tuple, while we are stuck with `str` for a string.

gaborbernat (changes requested):
I prefer the existing format.

abitrolly:
I see you don't have to type `_ctx.log(msg, kind=('debug'),)` too often.

gaborbernat:
Given 2/3 active maintainers are against it closing it as won't fix for now, thanks for opening the PR though!

abitrolly:
> we can improve the static typing for a tuple

@henryiii do you have a specific suggestion how to improve typing in this particular case when both `('one', 'two')` and `('one')` are accepted? If there is no suggesting, then it is just a compication on the syntax.

abitrolly:
> Given 2/3 active maintainers are against it closing it as won't fix for now, thanks for opening the PR though!

When voting replaces arguments it is not that I am particularly happy.

gaborbernat:
My argument is that explicit is better than implicit, and I find (personally) here easier to read (and type) the existing. It's fine to disagree on these, but ultimately the maintainers vote tends to win because they are the ones that have to deal with the code for longer period of time, so their opinion is weighted heavier.

layday:
> When voting replaces arguments it is not that I am particularly happy.

Honestly, I don’t feel particularly compelled to argue something so minor.  It doesn’t matter that much if it’s a string or tuple, but a structured data type can express a descendent relationship better than a hyphenated string might (which you end up having to split in order to match).

abitrolly:
> a structured data type can express a descendent relationship better than a hyphenated string might

Why say "might"? Make it structured, so that it can be serialized as JSON. Here is the Go way for cross tool compatibility https://go.dev/blog/slog

henryiii:
> do you have a specific suggestion how to improve typing in this particular case when both `('one', 'two')` and `('one')` are accepted?

```python
example: tuple[Literal['one'], Literal['two']] | tuple[Literal['one']]
```

abitrolly:
> example: tuple[Literal['one'], Literal['two']] | tuple[Literal['one']]

So what would be the line for the current `kind=` values?

https://github.com/pypa/build/blob/42da4c4b483563819fabdb8ecf2f6df53ed82123/src/build/__main__.py#L109-L121

With string it would be this:

```
None | Literal['step'] | Literal['subprocess-cmd'] | Literal['subprocess-stdout'] | Literal['subprocess-stderr']
```

layday:
> Why say "might"? Make it structured, so that it can be serialized as JSON. Here is the Go way for cross tool compatibility https://go.dev/blog/slog

It already is structured. A tuple is a data structure. JSON serialisation is not a goal; "kind" is just metadata for the CLI to know how to format the message.

Why insist on an incredibly nitpicky change to a project you are otherwise uninvolved in? What does it matter?

abitrolly:
@layday I want to bring back https://github.com/pypa/build/pull/969 and to debug what is needed for it to work I have to insert a dozen of multilevel log statements.

abitrolly:
> Why insist on an incredibly nitpicky change to a project you are otherwise uninvolved in? What does it matter?

I am bored / feel lonely / dying and need attention? :D

I don't know, seems like a useless waste of time, unconscious distraction in pursuit of more accessible source of dopamine instead of doing something useful, but what requires a considerably more efforts with equally lower chances of dopamine release.

PR #1056, opened 2026-04-29
fix: api should not ignore installed, only CLI

henryiii:
## Description

Our recent change broke pyodide-build, which is manually using our API and pre-installing some dependencies that are patched for pyodide. This is a proposed fix, which keeps the API explicit (users need to opt-in with `fresh=True`), but our CLI opts into it.

When asking copilot CLI to do the fix, it also suggested `--reinstall` for uv to match pip's behavior. I haven't worked out yet if that's the best thing here, but that might not hurt to have.

:robot: Assisted-by: Copilot:GPT-5.4

## Changelog

## Checklist

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -166,14 +166,12 @@ class DefaultIsolatedEnv(IsolatedEnv):
         requirements: Collection[str],
         constraints: Collection[str] = [],
         *,
-        fresh: bool = False,
+        _fresh: bool = False,  # Used internally by CLI to support preset PYTHONPATH
     ) -> None:
         """
         Install packages from PEP 508 requirements in the isolated build environment.
 
         :param requirements: PEP 508 requirement specification to install
-        :param fresh: Whether to overwrite already installed packages while installing
-            these requirements
 
         :note: Passing non-PEP 508 strings will result in undefined behavior, you *should not* rely on it. It is
                merely an implementation detail, it may change any time without warning.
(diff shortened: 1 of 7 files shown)


henryiii:
A few points to consider in review:
- Do we want to backport this for a 1.4.x release, or just release 1.5 (Python 3.8 has been dropped)
- Do we want to add `--reinstall` for uv (current state)?

gaborbernat (approved):

gaborbernat:
I'm alright to release a new version without backporting, personally.

layday:
I'm not sure about this.  I would prefer to drop `--ignore-installed` than muddy the interface with a parameter that'll probably see no use outside of pyodide.  Unless a good solution for `PYTHONPATH` can be found, I would rather simply mention it as a caveat in the docs.  The `DefaultIsolatedEnv` could also notify the user that dependencies might fail to install if it detects that `PYTHONPATH` is set.

henryiii:
This does drop `--ignore-installed` from programmatic use (unless you know about this parameter), and makes sure only the CLI uses it, where we do control what is happening. I feel like it's a good balance between keeping `python -m build` working and keeping third party usage of our API stable.

How about we make this parameter private for now, and only make it public if someone needs it? That gives us the ability to solve this another way in the future, while still fixing pyodide and keeping previous fix (which seems to be working, at least no one has complained).

src/build/env.py
```diff
@@ -394,6 +421,8 @@ def install_dependencies(  # pragma: no cover -- uv tests are skipped on PyPy, c
                 cmd += [f'-{"v" * min(2, verbosity - 1)}']
 
             cmd += ['install', *requirements]
+            if fresh:
+                cmd += ['--reinstall']
```
layday: I don't think uv is affected by `PYTHONPATH`, so let's leave it as it was before.

layday:
> How about we make this parameter private for now, and only make it public if someone needs it? That gives us the ability to solve this another way in the future, while still fixing pyodide and keeping previous fix (which seems to be working, at least no one has complained).

That's a good idea - let's do it that way.

layday (approved):

PR #1057, opened 2026-05-05
✨ feat(cli): accept sdist tarball as srcdir argument

### LINKED ISSUE #311: [FR] Implement building wheels from sdist/tarball artifacts
This is an important use case for building platform-specific matrixes of wheels. In this scenario, one would build an sdist in a separate job, save it as an artifact, and then a matrix of subsequent jobs would create wheels using the same artifact. As a bonus, other binary artifacts from external ecosystems, like RPM, normally also get built from sdist so such jobs would also be runnable in parallel.

I suggest supporting the following invocation:
```console
$ python -m build --wheel dist/some-tarball.tar.gz
```
I think this UX is good. And since the CLI already accepts a directory-type source in that place, pointing at a file would be natural.

Refs:
* https://github.com/pypa/build/issues/304#issuecomment-856736419
* https://github.com/pypa/build/issues/257#issuecomment-835349308

gaborbernat:
Closes #311.

Anyone testing a release artifact has had to extract the sdist by hand or run a separate `pip wheel` against the tarball before they could exercise the wheel build path. The positional `srcdir` argument now accepts a `.tar.gz` source distribution directly, matching the "build a wheel from an sdist" recipe that came out of the discussion on #311. 📦

When `srcdir` points at a `.tar.gz`, build first validates it: the filename must parse per PEP 625, the archive must open as tar, contain a single top-level directory, and include a `PKG-INFO`. The validated archive is extracted into a temporary directory and a wheel is built from the extracted source. The default `--outdir` for archive input is the directory containing the archive, matching the convention `pip wheel` already uses, so the wheel lands next to the sdist that produced it. ✨

The default for archive input is wheel-only rather than sdist+wheel: regenerating an sdist from another sdist is not a defined operation across backends, and the issue thread converged on rejecting it for now. Passing `--sdist` against an sdist input is therefore an error with a message pointing back at #311. `--metadata` and `--wheel` work as expected; `--config-setting`, `--no-isolation`, `--installer`, and friends are passed through unchanged. The public Python API (`build_package`, `build_package_via_sdist`, `ProjectBuilder`) keeps its directory-only contract — this is a CLI-only affordance, since broadening the API surface was outside the scope FFY00 had approved on the issue.

| Input | default | `--wheel` | `--sdist` | `--metadata` |
|---|---|---|---|---|
| directory | sdist + wheel via sdist | wheel | sdist | metadata |
| `.tar.gz` | wheel | wheel | error | metadata |

## Validation against real PyPI sdists

Built five sdists pulled from PyPI:

| Package | Backend | Result |
|---|---|---|
| `attrs-26.1.0` | hatch-vcs | wheel built |
| `click-8.3.3` | flit-core | wheel built |
| `idna-3.13` | flit-core | wheel built |
| `wheel-0.47.0` | flit-core | wheel built |
| `markupsafe-3.0.3` | setuptools + C ext | platform wheel built (native ext compiled) |

Error paths exercised manually:

- `--sdist <archive>`: parser rejects with the #311-pointed error
- corrupt `.tar.gz`: `BuildException`: "failed to read source distribution" with tarfile diagnostics
- filename without a version (`notavalid.tar.gz`): "does not look like a source distribution"
- `--metadata <archive>`: PKG-INFO surfaced as JSON
- default no-flag: wheel lands next to the input archive

## Source-built vs sdist-built wheel equivalence

Built the same package both ways and compared:

| Package | Backend | sha256 (source) == sha256 (sdist) |
|---|---|---|
| `build` (this repo) | flit-core | `87361bc1…` (byte-identical) |
| `attrs-26.1.0` | hatch-vcs | `94d456bc…` (byte-identical) |

For pure-Python wheels with deterministic backends, the sdist path produces a bit-for-bit identical wheel.

### DIFF
--- a/docs/how-to/basic-usage.rst
+++ b/docs/how-to/basic-usage.rst
@@ -76,6 +76,22 @@ Or explicitly:
 
     $ python -m build --srcdir path/to/project
 
+*************************************
+ Building from a source distribution
+*************************************
+
+The positional argument also accepts a ``.tar.gz`` `source distribution
+<https://packaging.python.org/en/latest/specifications/source-distribution-format/>`_. Build checks the filename and the
+presence of ``PKG-INFO``, extracts the archive into a temporary directory, and runs the wheel build against the
+extracted source. The wheel lands next to the input archive.
+
+.. code-block:: console
+
+    $ python -m build dist/mypackage-1.0.0.tar.gz
+
+No flag means the same as ``--wheel``. ``--metadata`` works the same way. ``--sdist`` errors out, since the archive
+already is an sdist.
+
 *********************************
  Specifying the output directory
 *********************************
@@ -234,7 +250,7 @@ Build from the sdist to ensure it's complete:
 .. code-block:: console
 
     $ python -m build --sdist
-    $ python -m build --wheel dist/mypackage-1.0.0.tar.gz
+    $ python -m build dist/mypackage-1.0.0.tar.gz
 
 Or test installation:
 
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -10,10 +10,13 @@ __lazy_modules__ = [
     'functools',
     'json',
     'os',
+    'packaging.utils',
+    'packaging.version',
     'platform',
     'pyproject_hooks',
     'shutil',
     'subprocess',
+    'tarfile',
     'tempfile',
     'textwrap',
     'traceback',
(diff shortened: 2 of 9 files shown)


src/build/__main__.py
```diff
@@ -332,29 +339,22 @@ def build_package_via_sdist(
...
-                        distribution,
-                        config_settings,
-                        skip_dependency_check,
-                        dependency_constraints_txt,
-                        installer,
-                    )
-                    built.append(os.path.basename(out))
-            finally:
-                shutil.rmtree(sdist_out, ignore_errors=True)
+        with _extract_sdist(sdist, sdist_name[: -len('.tar.gz')]) as extracted_srcdir:
```
```suggestion
        with _extract_sdist(sdist, sdist_name.removesuffix('.tar.gz') as extracted_srcdir:
```

docs/how-to/basic-usage.rst
```diff
@@ -234,7 +250,7 @@ Build from the sdist to ensure it's complete:
 .. code-block:: console
 
     $ python -m build --sdist
-    $ python -m build --wheel dist/mypackage-1.0.0.tar.gz
+    $ python -m build dist/mypackage-1.0.0.tar.gz
```
henryiii, line 253: I take it these docs were wrong, that wasn't supported till now?

gaborbernat: Indeed 😊 wishful thinking I guess? 😆

henryiii (approved):

PR #1058, opened 2026-05-07
test: cover config settings forwarding in build_package and build_package_via_sdist

terminalchai:
## Problem

`build_package` and `build_package_via_sdist` both accept a `config_settings` parameter and forward it to `_build`, but there were no tests verifying that this forwarding actually happens. A refactor that silently dropped `config_settings` from the `_build` call would pass the existing test suite.

## Fix

Add two unit tests:

- **`test_build_package_passes_config_settings_to_build`** — calls `build_package` with `config_settings={'--flag': 'value'}` and asserts both the `sdist` and `wheel` `_build` calls receive the settings in the correct position.
- **`test_build_package_via_sdist_passes_config_settings_to_build`** — same for `build_package_via_sdist`, also covering the `sdist`→`wheel` two-step flow (temp dir creation, extraction, cleanup).

Both tests also cover the other recently-added parameters (`dependency_constraints_txt`, `installer`) to make the call signatures explicit.


gaborbernat (changes requested):
Merge conflict 🤔

gaborbernat:
@terminalchai can you fix the CI and also the merge conflict?

gaborbernat (approved):

layday:
It looks like nothing's been changed here?

Issue #455, opened 2022-04-07
Add option for cleaning outdir before using it


ssbarnea:
As many bugs can happen when `dist/` output folder ends-up collecting artifacts from previous runs, it is extremely common to want to clean the folder before running build, so it will only contain the outcome of a single build command execution.

As cleaning a folder from shell is not really portable across all platforms people end-up using ugly python snippets before running build. Examples like https://github.com/ansible/ansible-lint/blob/main/tox.ini#L173-L181 are quite common.

In order to simplify use, I would like to propose adding an option to build that performs the cleaning, so users will not longer have to implement such workarounds. I would not mind even making the cleaning enabled by default but I guess it would be found dangerous as some might try to use `.` as output directory.

WDYT? Should I try to make pull-request for it?

ssbarnea:
I am talking about cleaning files created by build itself, on previous runs, not about leftovers from other tools. If the package version changes, new files are created and when using setuptools-scm, version would change on any commit.

Few years back I was considering writing a tox plugin that can clean some folders like `dist/ ` on runs but the reality is that issue is not related to tox, it related to build itself, so it would be better to improve the tool and allow users to avoid this issue regardless if they use tox or another task runner.

On CI/CD cleaning is usually not a problem as you likely start from clean. Even so I do remember some cases where even on these we ended up with multiple artifacts (release pipelines that create additional release commits being such an example).

Regarding implementation, we could make the option to clean only file patterns that are matching the package name, file `foo-*.{whl,tag.gz}`.

gaborbernat:
This is a valid feature request, but it should require an explicit opt-in flag to avoid accidental data loss. Something like `--clean-outdir` that users must explicitly provide.

The implementation should only clean the output destination folder (the `--outdir` argument, which defaults to `dist/`). It should not touch any build backend generated directories or temporary build directories. This keeps the scope focused and predictable - you're only cleaning the final artifact destination.

This is particularly useful in CI/CD pipelines where you want to ensure a clean artifact directory, and it's common practice in other build systems to have a clean target. The explicit flag requirement means users won't accidentally lose files, but they can opt in when they know what they're doing.

The safest implementation would be to delete the entire output directory only when the flag is provided, with clear documentation that it removes everything in that directory. This makes the behavior predictable and avoids complexity around pattern matching specific package names.

Keeping open as a valid enhancement request - an opt-in `--clean-outdir` flag would be useful.

henryiii:
[...]

PR #1061, opened 2026-05-15
✨ feat(cli): add --clean-outdir-artifacts flag

gaborbernat:
When building repeatedly into the same output directory, stale `.tar.gz` and `.whl` files from previous runs accumulate alongside new artifacts. This makes it easy to accidentally publish or install an outdated wheel. Users previously had to `rm -rf dist/` manually before each build, which risks deleting non-build files if `--outdir` points somewhere shared.

`--clean-outdir-artifacts` removes only `.tar.gz` and `.whl` files from the resolved output directory before the build runs. Other files in the directory are left untouched. The flag is CLI-only — it does not appear on the `build_package` or `build_package_via_sdist` signatures, keeping the programmatic API focused on building. 🧹 A public `clean_outdir_artifacts(outdir)` function is exposed for callers who want the same behavior without invoking the CLI.

The flag is a no-op when the output directory does not yet exist.

### DIFF
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -365,7 +365,7 @@ def clean_outdir_artifacts(outdir: StrPath) -> None:
         for entry in outdir_path.iterdir():
             if entry.name.endswith(('.tar.gz', '.whl')):
                 entry.unlink()
-                _ctx.log(f'Removed stale artifact {entry.name}', kind=('step',))
+                _ctx.log(f'Removed stale artifact {outdir_path / entry.name}', kind=('step',))
 
 
 def _build_metadata(
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -731,6 +731,9 @@ if __name__ == '__main__':  # pragma: no cover
 
 
 __all__ = [
+    'build_package',
+    'build_package_via_sdist',
+    'clean_outdir_artifacts',
     'main',
     'main_parser',
 ]
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -29,6 +29,7 @@ import contextlib
 import contextvars
 import json
 import os
+import pathlib
 import platform
 import shutil
 import subprocess
@@ -360,15 +361,11 @@ def build_package_via_sdist(
 
 def clean_outdir_artifacts(outdir: StrPath) -> None:
     """Remove ``.tar.gz`` and ``.whl`` files from *outdir* so the build starts from a clean slate."""
-    outdir_path = os.fspath(outdir)
-    if os.path.isdir(outdir_path):
-        removed = False
-        for entry in os.listdir(outdir_path):
-            if entry.endswith(('.tar.gz', '.whl')):
-                os.remove(os.path.join(outdir_path, entry))
-                removed = True
-        if removed:
-            _ctx.log(f'Cleaned build artifacts from {outdir_path}', kind=('step',))
+    if (outdir_path := pathlib.Path(outdir)).is_dir():
+        for entry in outdir_path.iterdir():
+            if entry.name.endswith(('.tar.gz', '.whl')):
+                entry.unlink()
+                _ctx.log(f'Removed stale artifact {entry.name}', kind=('step',))
 
 
 def _build_metadata(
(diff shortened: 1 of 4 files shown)


src/build/__main__.py
```diff
@@ -358,6 +367,13 @@ def build_package_via_sdist(
     return [sdist_name, *built]
 
 
+def _clean_outdir(outdir: StrPath) -> None:
```
henryiii: Instead of adding another keyword argument to the two build functions above, why not make this function public? That's a simpler, less coupled interface.

henryiii: Also, this is very simple, so I'd also be fine if it just wasn't public at all, and programatic users just had to delete themselves. The use case is specifically for CLI on Windows were you might have a different command to clear it.

layday:
I'd personally flip this around, erroring when `outdir` is not empty, but not doing any of the deleting ourselves. Maximising convenience is not always worthwhile; I think safety should come first in this case.  If your `outdir` is not empty, perhaps you've passed in the wrong path or another tool's using the "dist" directory and you might wanna separate them.

gaborbernat:
> I'd personally flip this around, erroring when `outdir` is not empty, but not doing any of the deleting ourselves. Maximising convenience is not always worthwhile; I think safety should come first in this case.  If your `outdir` is not empty, perhaps you've passed in the wrong path or another tool's using the "dist" directory and you might wanna separate them.

I personally do not agree with this. We should not force the user to do their own deletion when they can explicitly specify and request it. Given they have to specifically request this as an additional flag, it's safe to assume that they know what they are doing, and we are safe to go ahead.

src/build/__main__.py
gaborbernat: I'm confused. This makes it not public, and it's what you're describing in your second comment, so what's wrong with that?

henryiii: I don't want the keyword argument to be added, programmatic users can delete this just fine by themselves.

henryiii:
Should we also error out if the directory is not otherwise empty? I like this version better, much less likely to delete something unexpected. :)

gaborbernat:
> Should we also error out if the directory is not otherwise empty? I like this version better, much less likely to delete something unexpected. :)

Neah I think the current implementation is good enough.

henryiii:
If I approve this it gets automerged, so consider me approving, I want to make sure @layday is happy with it too. I think it's a good balance this way.

layday:
My opinion's that the value proposition of this change is pretty low.  The tool has acquired a fair few options recently and I feel we're on the cusp of overwhelming our users. New options should be evaluated on whether they're adding something that we'd not otherwise be able to accomplish (with ease). A non-standard shortcut for `rm -r` doesn't really cross that bar in my mind. 

If this were opt-out, I'd consider it an improvement.  If we don't believe it should be the default, then it's not really worth having almost at all.

gaborbernat:
> A non-standard shortcut for `rm -r` doesn't really cross that bar in my mind.

rm is not cross platform 🤔 and making it opt-out could cause issues with existing workflows, so needs to remain opt-in. So I'm strongly in favor of it.

layday:
`rm` is an interface. There are Windows tools which implement this interface. BusyBox is one. There are cross-platform tools which implement other interfaces.  The JS community has rifraf.  Personally, I prefer `git clean` since I'm always working in a git repo.  It makes more sense to me to approach something like this holistically - your mileage may of course vary.

I'm -1, but if you feel strongly about this, I won't insist.

henryiii:
I'm a a 0 currently. I don't feel strongly enough to break the tie. For compiled packages, you can't use this anyway because you will often be intentionally building multiple times with different Pythons and such.

Do we have real examples of CI jobs that would benefit from this? I expect if you are using an AI, it would probably prefer the standard tricks over a custom flag it has to look up.

gaborbernat:
https://github.com/pypa/build/issues/455 the issue has 7 thumbs up so there's definitely a need here

henryiii:
I gave an example of how it's easy to do in tox today; unless I see an example of someone actually needing it and not having a trivial workaround, I'm 0. If @gaborbernat still feels strongly about this, it's fine (due to the comment "but if you feel strongly about this, I won't insist").

PR #1068, opened 2026-05-22
ci: secure GitHub Actions workflows

henryiii:
Used [my secure-ci skill](https://github.com/henryiii/skills/blob/main/secure-ci/SKILL.md), and tweaked a bit.

:robot: *Human guided, AI assisted PR ([using this skill](https://github.com/henryiii/skills/blob/main/branch-and-pr/SKILL.md)). AI text below.* :robot:

## Summary

This PR secures the repository's GitHub Actions workflows following zizmor pedantic auditing and best practices.

### Changes

**Permissions hardening:**
- Added `permissions: {}` to all reusable workflows (`reusable-change-detection.yml`, `reusable-pytest.yml`, `reusable-docs.yml`, `reusable-type.yml`) to avoid default broad permissions
- Added explanatory comments to all non-default permission declarations per zizmor requirements

**Concurrency:**
- Added concurrency groups to `cd.yml` and `pre-release.yml` (previously missing)

**Template injection fixes:**
- `pre-release.yml`: Moved `inputs.bump` and `github.repository` from template expansion to environment variables
- `reusable-pytest.yml`: Moved `matrix.py` and `matrix.tox-target` from template expansion to environment variables

**Job naming:**
- Added explicit `name` fields to all jobs missing them (zizmor `anonymous-definition` audit)

**Pre-commit:**
- Updated zizmor pre-commit hook to SHA-pinned `v1.24.1` with `files` filter and `--persona=pedantic` args
- Froze all pre-commit hook revs to SHA pins via `prek auto-update --freeze`

**Dependabot:**
- Changed schedule from `daily` to `monthly`
- Renamed group from `actions` to `github-actions`
- ~~Added `pre-commit` ecosystem with monthly schedule and 7-day cooldown~~

**Cooldowns:**
- Added `exclude-newer = "7 days"` to `[tool.uv]` in `pyproject.toml`

Assisted-by: OpenCode:glm-5

PR #1074, opened 2026-05-29
Revert "build(deps): bump pip from 22.3 to 26.1"

layday:
Reverts pypa/build#1072

### DIFF
--- a/tests/constraints.txt
+++ b/tests/constraints.txt
@@ -1,6 +1,6 @@
 importlib-metadata==4.6
 packaging==24.0
-pip==26.1; python_version < "3.12"
+pip==22.3; python_version < "3.12"
 pip==23.2; python_version >= "3.12" and python_version < "3.15"
 pip==25.3; python_version >= "3.15"
 pyproject_hooks==1.0


gaborbernat:
Why?

layday:
These PRs should not have been merged. The versions in this file are supposed to be the minimums defined in pyproject.toml and are used in our min version tests.

gaborbernat:
Why was the PR generated in the first place, can we fix that first?

layday:
I don't know. I thought it might be related to #1068 given the timing of the PRs but it doesn't looks like it's changed anything to have caused this.

layday:
The [alerts](https://github.com/pypa/build/security/dependabot) do say that they were created from #1068, but I don't see where we might've opted into pip alerts and updates for Dependabot.

gaborbernat (approved):

gaborbernat:
> The [alerts](https://github.com/pypa/build/security/dependabot) do say that they were created from #1068, but I don't see where we might've opted into pip alerts and updates for Dependabot.

@henryiii ideas?

layday:
I can only guess that "Dependabot security updates" gets flicked on automatically when you update your dependabot config.  I don't have access to this repo's settings to see though.

henryiii:
We might be able to move over to using min versions with uv and drop the file. Though the pyproject.toml is also generating warnings due to the lower bounds. I didn't change any settings other than the name of the pass job in branch protection.

PR #1077, opened 2026-06-01
Fix metadata extraction for build-tagged wheels

xujiantop-crypto:
## Description

When a backend does not implement prepare_metadata_for_build_wheel, metadata_path() builds a wheel and derives the .dist-info directory name from the wheel filename. The current filename parser is greedy enough to misread wheels with a build tag, for example pkg-1.0-1-py3-none-any.whl, so it looks for the wrong .dist-info directory.

This change validates wheel filenames with packaging.utils.parse_wheel_filename() and then splits the validated filename into the original wheel filename components. That keeps the escaped distribution/version strings needed for the .dist-info path while correctly handling the optional build tag.

## Changelog

- [x] Added changelog fragment: docs/changelog/build-tag-wheel-metadata.bugfix.rst

## Checklist

- [x] Tests pass locally (	ox)
  - PYTHONPATH=src python -m pytest tests/test_projectbuilder.py::test_check_dependency tests/test_projectbuilder.py::test_metadata_path_no_prepare tests/test_projectbuilder.py::test_metadata_path_no_prepare_with_build_tag tests/test_projectbuilder.py::test_metadata_invalid_wheel -q
- [x] Code follows project style (	ox -e fix)
  - python -m ruff check src/build/_util.py tests/test_projectbuilder.py tests/packages/test-no-prepare/backend_no_prepare.py tests/packages/test-no-prepare-build-tag/backend_no_prepare.py
- [ ] Type checks pass (	ox -e type)
- [ ] Documentation builds (	ox -e docs)


layday:
Please open a ticket describing the issue and a human will perform whatever changes may be necessary.

Issue #789, opened 2024-06-18
Print artifact digest after build process is completed


elad-pticha:
I think it would be beneficial to add artifact digests to the Successfully built message. For instance:
``` bash
Successfully built test-0.0.0.tar.gz@sha256:960b9adda66023aed657c0da9626a6b8de71e433843181a5397431465adb57a7 and test-0.0.0-py3-none-any.whl@sha256:c856716babc6d603769f6b4f1a7122a61b870e7f627b0e2b4aa8f48b712c7770
```

![image](https://github.com/pypa/build/assets/124190990/1cb1f36c-e13e-4045-b9a3-c85218fed4eb)

Including the digest information strengthens the link between the build system and the package repository.

Let's take, for example, a GitHub Actions pipeline that builds and pushes packages to PyPI. Users can check the package's hash in the package repository but have no idea if that artifact is really built from the source repository it claims to be related to. If a digest is printed at the end of the build process, it could indicate that the package in the package registry has been created in a specific build pipeline.

This also relates to the world of [SLSA](https://slsa.dev/) and artifact integrity, allowing users to link between an artifact to its source.

elad-pticha:
If you think this is a good idea, I can submit a PR to implement it.

henryiii:
I’d like to have a machine readable output (json probably), but it can’t be printed, since it is problematic to try to pick it out of stdout. Probably an argument that allows writing the built package info to a stream or file. Having the name(s) and maybe other info is useful too when building pipelines.

henryiii:
If you had such a machine readable output, then you could process it and produce the digest in your preferred format. No need to make a list of formats.

elad-pticha:
I agree that build logs are not a place to put security information and that this does not come any close to creating a proper SLSA attestation, but maybe it is better than having no information at all for projects that don't produce attestations.

Another option is that this issue should be under the [twine](https://github.com/pypa/twine) repository and that the digest should be printed after the package has been pushed.

PR #1079, opened 2026-06-03
✨ feat(cli): add --json-output build report

### LINKED ISSUE #198: Listing filenames of produced distributions in the CLI

gaborbernat:
Anyone scripting a release hits the same wall: the files `build` produces have names they cannot predict. The version, Python tag and platform all land in the filename, and only the backend decides them at build time. So the moment a pipeline wants to upload, install, or attach what it just built, it has to go guess. 📦

Today people work around this, and every route has a sharp edge. Globbing `dist/*` quietly sweeps up stale artifacts from an earlier run, so the fix becomes `rm -rf dist` before every build. Choosing a fresh output folder then globbing runs into `pip install foo/*` not expanding globs at all. `twine upload dist/*` only works because twine happens to expand globs itself. Asking pip to resolve names with `--find-links` means knowing the project name up front and paying for a full pip invocation. Parsing standard output looks tempting until a backend writes its own chatter there and corrupts whatever you scraped. Writing a manifest into `dist/` then gets picked up by `twine dist/*` and breaks the upload. None of these is safe to hand a new contributor.

This adds `--json-output PATH`, which records what the build actually produced once it finishes: each distribution's filename and absolute path, as JSON. Point it at a file, or pass `-` to send the report to standard output (the success summary steps aside to standard error so the JSON stays clean). A release job can now read the exact artifacts straight from the report instead of guessing. ✨

```sh
python -m build --json-output report.json
twine upload $(jq -r '.artifacts[].path' report.json)
```

The report sits where you ask, so it stays out of `dist/` and away from tools that glob it. It reports rather than predicts, which is the only honest option when the backend owns both the filenames and the output streams. Reference, how-to, tutorial and explanation docs all cover it.

Closes #198

### DIFF
--- a/docs/changelog/198.feature.rst
+++ b/docs/changelog/198.feature.rst
@@ -1,0 +1,3 @@
+Add a ``--json-output PATH`` option that writes a JSON report of the built distributions, listing each filename and
+absolute path, to ``PATH`` or to standard output when ``PATH`` is ``-``. A script can read the artifact names from it
+instead of globbing ``dist/`` or parsing the build backend's output - by :user:`gaborbernat`.
--- a/docs/explanation/how-it-works.rst
+++ b/docs/explanation/how-it-works.rst
@@ -230,6 +230,23 @@ This:
 
 Useful for tools that need package metadata without building the full package.
 
+***************************
+ Reporting Built Artifacts
+***************************
+
+The backend settles a distribution's filename only when it produces the file: the name encodes the version, Python tag
+and platform, and the backend may emit a pure-Python wheel, a platform-specific wheel, or something in between.
+``--json-output`` reports the filenames once the build finishes rather than predicting them:
+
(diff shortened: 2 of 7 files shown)


gaborbernat:
Superseded by #1080, which takes the same approach for #198 but writes to a frontend-owned file only (no `-`/stdout, since the backend subprocess can write to stdout too) and enriches the report with each artifact's `kind`, `size` and SHA-256 hash — folding in #789 — under a versioned schema with an atomic write. Continuing the work there.

PR #1080, opened 2026-06-03
✨ feat(cli): machine-readable build report, and metadata from a wheel

### LINKED ISSUE #198: Listing filenames of produced distributions in the CLI

gaborbernat:
Closes #198.

Anyone wiring `build` into a release pipeline hits the same wall: once the build finishes, they need to know *which* files came out so they can upload, sign, or checksum them. The filenames are not predictable, since they carry the version and, for wheels, the Python, ABI and platform tags. So today people reach for workarounds that each have a sharp edge. They glob `dist/*` and trust that the next tool expands it (only `twine` does; `pip install dist/*` does not). They `rm -rf dist` before every build so a stale artifact from last time does not get published by accident. They scrape the coloured `Successfully built …` line off the terminal. Or they drop down to the Python API and reimplement a chunk of the CLI. 📦 None of these is something we want to be the recommended path.

This adds a first-class answer: `python -m build --report dist/report.json` writes a small JSON file describing exactly what was produced — each artifact's name, path, kind (`sdist` or `wheel`), size, and SHA-256 hash. A pipeline can then do `twine upload $(jq -r '.artifacts[].path' dist/report.json)` and be certain it is shipping this build's output and nothing else. Including the hash also gives projects a ready hook for provenance and checksums (the ask from #789, folded in here). ✨

The report goes to a file the frontend owns rather than to stdout, because the build backend runs as a subprocess that can print to stdout itself — so a structured payload there could never be parsed safely. The report is JSON and its `version` field is a schema version, so the format can grow later without breaking readers. The default `python -m build` experience is untouched and the feature is entirely opt-in.

Building on that, `--metadata` now also accepts a built wheel: `python -m build --metadata dist/mypackage-1.0.0-py3-none-any.whl` reads the wheel's `METADATA` and prints it as JSON without building anything. A pipeline can lift a wheel's path straight out of `report.json` and inspect what it shipped — no rebuild, no unzip.

The fragile stopgaps people reach for in the issue threads until this lands:

- [#198 (comment)](https://github.com/pypa/build/issues/198#issuecomment-740115413) — `rm -rf dist` before each build, then `twine upload dist/*` relying on twine's own glob expansion to avoid shipping stale artifacts.
- [#198 (comment)](https://github.com/pypa/build/issues/198#issuecomment-762794454) — the catch: `pip install dist/*` does *not* glob-expand, so CI still has to discover the dynamically named files some other way.
- [#198 (comment)](https://github.com/pypa/build/issues/198#issuecomment-740102639) — the original ask: write a manifest JSON and pipe its path through `jq` into the installer.
- [#198 (comment)](https://github.com/pypa/build/issues/198#issuecomment-2067611620) — hand-rolling a JSON file in `dist/`, which then breaks because `twine dist/*` tries to upload the JSON to PyPI.
- [#198 (comment)](https://github.com/pypa/build/issues/198#issuecomment-2074156261) — confirmation that stdout/stderr scraping cannot work (the backend writes there too), with [uv#2314](https://github.com/astral-sh/uv/pull/2314) and [twine#281](https://github.com/pypa/twine/issues/281) showing the same problem hit elsewhere.
- [#789 (comment)](https://github.com/pypa/build/issues/789#issuecomment-2179420077) — the rebuttal to scraping the coloured `Successfully built …` line for digests: build logs are the wrong place; machine-readable output is the real fix.

Release pipelines wipe `dist/` first, then re-discover the produced files by globbing, because nothing tells them the names:

- [pypa/packaging `noxfile.py` (L400-L406)](https://github.com/pypa/packaging/blob/23a1cc498e28c7c915319f16ce89f0a6563d3ead/noxfile.py#L400-L406) — after `python -m build`, reconstructs the expected `dist/{name}-{version}-...whl`/`.tar.gz` names by hand (`rmtree("dist")` at L420). Even pypa's own tooling guesses the filenames.
- [pypa/cibuildwheel `noxfile.py` (L198-L205)](https://github.com/pypa/cibuildwheel/blob/2f49f9a050c42805d0f0c5cfd7083637c11a5b4e/noxfile.py#L198-L205) — `shutil.rmtree` on `build/` and `dist/` before `python -m build` to tell new output from stale.
- [Python-Markdown `makefile` (L21-L24)](https://github.com/Python-Markdown/markdown/blob/ddead47873e7918829e1d8b0eaee6ab725793cc9/makefile#L21-L24) — `rm -rf build dist`, then `python -m build`, then `twine upload dist/*`.
- [openfisca-core `tasks/publish.mk` (L14-L16)](https://github.com/openfisca/openfisca-core/blob/4f7f09833afe7e8b6856e8d7a3016c04a931009b/tasks/publish.mk#L14-L16) — `python -m build`, then `find dist -name "*.whl"` to locate the wheel to install.
- [python-a2a `release.sh` (L531-L536)](https://github.com/themanojdesai/python-a2a/blob/40b969a465f85e8156793b7cfac16c5e23ed0b86/release.sh#L531-L536) — `find dist -name "*.whl"` to install the just-built wheel, after `rm -rf dist` (L78) and `python -m build` (L349).

### DIFF
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -25,8 +25,8 @@ __lazy_modules__ = [
     'tempfile',
     'textwrap',
     'traceback',
-    'typing',
     'warnings',
+    'zipfile',
 ]
 
 import argparse
@@ -43,6 +43,7 @@ import tempfile
 import textwrap
 import traceback
 import warnings
+import zipfile
 
 from functools import partial
 from tarfile import TarError
@@ -422,7 +423,9 @@ def _build_metadata(
     installer: _env.Installer = 'pip',
     env_dir: str | None = None,
 ) -> list[str]:
-    import packaging.metadata
+    if os.path.isfile(srcdir) and os.fspath(srcdir).lower().endswith('.whl'):
+        _print_metadata(_wheel_metadata(srcdir))
+        return []
 
     def run_subprocess(cmd: Sequence[StrPath], cwd: str | None = None, extra_environ: Mapping[str, str] | None = None) -> None:
         env = os.environ.copy()
@@ -448,12 +451,27 @@ def _build_metadata(
             'rb',
         ) as metadata_file,
     ):
-        valid_metadata, _ = packaging.metadata.parse_email(metadata_file.read())
+        _print_metadata(metadata_file.read())
+
+    return []
+
+
+def _wheel_metadata(wheel: StrPath) -> bytes:
+    with zipfile.ZipFile(wheel) as archive:
+        names = [name for name in archive.namelist() if name.count('/') == 1 and name.endswith('.dist-info/METADATA')]
+        if len(names) != 1:
+            msg = f'{os.fspath(wheel)!r} is not a valid wheel: expected one .dist-info/METADATA, found {len(names)}'
+            raise BuildException(msg)
+        return archive.read(names[0])
+
+
+def _print_metadata(raw_metadata: bytes) -> None:
+    import packaging.metadata
+
+    valid_metadata, _ = packaging.metadata.parse_email(raw_metadata)
     print(  # noqa: T201
         json.dumps(valid_metadata, ensure_ascii=False, indent=2),
     )
-
-    return []
 
 
 def main_parser() -> argparse.ArgumentParser:
@@ -510,7 +528,8 @@ def main_parser() -> argparse.ArgumentParser:
         type=str,
         nargs='?',
         default=os.getcwd(),
-        help='source directory or .tar.gz source distribution (defaults to the current working directory)',
+        help='source directory, .tar.gz source distribution, or (with --metadata) a .whl to read metadata from '
+        '(defaults to the current working directory)',
     )
 
     global_group = parser.add_argument_group('global options')
@@ -578,7 +597,8 @@ def main_parser() -> argparse.ArgumentParser:
     build_group.add_argument(
         '--metadata',
         action='store_true',
-        help="print out a wheel's metadata in JSON format. Cannot be used in conjunction with ``--sdist`` or ``--wheel``",
+        help="print out a wheel's metadata in JSON format, building it first unless the source argument is already a "
+        '.whl. Cannot be used in conjunction with ``--sdist`` or ``--wheel``',
     )
     build_group.add_argument(
         '--report',
@@ -687,8 +707,9 @@ def main(cli_args: Sequence[str], prog: str | None = None) -> None:
     _setup_cli(verbosity=args.verbosity)
 
     sdist_input = os.path.isfile(args.srcdir) and os.fspath(args.srcdir).lower().endswith('.tar.gz')
+    wheel_input = os.path.isfile(args.srcdir) and os.fspath(args.srcdir).lower().endswith('.whl')
     build = partial(
-        _select_build(parser, args, sdist_input=sdist_input),
+        _select_build(parser, args, sdist_input=sdist_input, wheel_input=wheel_input),
         config_settings=_resolve_config_settings(args),
         isolation=not args.no_isolation,
         skip_dependency_check=args.skip_dependency_check,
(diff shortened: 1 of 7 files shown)


layday:
I like this, but I think we should make it work with `--metadata` in some fashion. Perhaps `--metadata` could accept a wheel as input, the path of which would be extracted from the report, and generate its metadata. (This can happen in a follow-up.)

src/build/__main__.py
```diff
@@ -518,6 +520,17 @@ def __call__(
...
         help="print out a wheel's metadata in JSON format. Cannot be used in conjunction with ``--sdist`` or ``--wheel``",
     )
+    build_group.add_argument(
+        '--report',
+        help='write a machine-readable report of the built artifacts (name, path, kind, size and SHA-256 hash) to PATH. '
+        'Cannot be used together with ``--metadata``',
+        metavar='PATH',
+    )
+    build_group.add_argument(
+        '--output-format',
```
layday: I feel this is too generic an option name for the report's format.  It might be better to just drop this for now and re-introduce it (under a different name) once/if we end up supporting multiple formats.

gaborbernat: Good call — dropped `--output-format`; the report stays JSON and we can reintroduce a format switch under a clearer name if we ever add another. Done in the latest push.

layday:
I'm just thinking if this shouldn't be co-located with the built dists.  I feel like having to pass a path to `--report` might be a little too much friction?  I'd personally never not want to produce a report; if build would do this by default, I'd be happy. I could live with having to specify a flag, but having to pass the path feels a little redundant.

gaborbernat:
Good point on the friction — reworked. `--report` is now a no-arg flag: by default it writes `build-report.json` next to the distributions in `--outdir`, so the common case is just `python -m build --report`. An optional `--report PATH` is still there for a custom location.

I kept it opt-in rather than always-on, since writing a file the user didn't ask for felt surprising — but it's now zero-config when you do want it.

On concurrency: the write is atomic (temp file + `os.replace`), so a reader never sees a partial file, but two builds sharing one `--outdir` overwrite each other's `build-report.json` (last wins) — the same way their distributions would. I documented that parallel builds with differing args should use separate `--outdir` (or distinct `--report PATH`), since a per-build output dir is the safer pattern regardless.

henryiii:
I'm not sure putting it in the outdir is always the best (I'm at least happy there's a report path option!) `twine upload dist/*` is the standard way to upload ([even explicitly done that way in the intro tutorial!](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#upload-your-distributions)), and that would probably break if there's suddenly a non-wheel/sdist file. In fact, the whole argument for #455 was to keep the outdir clean, and now we are putting an extra non-uploadable file into it?

I'd prefer the path be required, I think. I can't think of a good reason to put it in the `dist/` folder (breaks uploads) or in the source folder (would often get picked up and placed in the sdist), so just forcing a choice is probably best?

(If twine does ignore non-wheel non-sdist files even when passed explicitly via shell expansion, that would be better. Also if it was prefixed by a dot, then shell expansion wouldn't pick it up, that's another option).

layday:
That's a good point. Maybe we should discuss this with the twine team, to see if there's any appetite for uploading files from a manifest.

gaborbernat:
My initial design was to let the report flag take a directory, and now that another maintainer also prefered that, I'll revert to it. I think is a cleaner approach 😊

henryiii:
Why not just make it take the full path (`--report=dist/anything.json`), then we can work on a "flag-less" version later once we see if twine can ignore it and pick default names and location then?

gaborbernat:
Done — `--report` now requires the file path (`--report build-report.json`), there is no default location. The docs steer the examples away from the output directory so `twine upload dist/*` workflows stay safe, and a flag-less default can come later if twine learns to ignore non-distribution files. Also rebased onto latest main.

layday (approved):

PR #1081, opened 2026-06-03
✨ feat(cli): warn on valueless config settings

### LINKED ISSUE #640: Config settings without an argument
Currently, build supports argument less config-settings:

```
-Csomething
```

Is equivalent to

```
-Csomething=""
```

(since config settings is a dict in the spec). This is used, for example, to allow `--config-setting=--with-c-extensions` to be supported. However, pip does not support it.

I don't have a strong option on if this should be supported or not (I think I kind of like it?), but I think we should be consistent. What do people think here? If we like supporting this, I think we should ask pip to support it too. If we don't like supporting it (if it was added by mistake, for example), we could produce a warning when it's detected describing how to write it in a pip compatible way.

gaborbernat:
People reach for the same `-C` / `--config-settings` flag whether they build with build or install with pip, and they expect a command that works in one to work in the other. build quietly accepts a valueless setting such as `-C--with-c-extensions` and treats it as an empty value, while pip refuses anything without an `=`. Someone who copies a working build command into a pip invocation hits a confusing rejection. Today people get past it by reading pip's error closely enough to learn they must append `=`, or by lifting the workaround from issue threads such as cibuildwheel's. Henry raised the inconsistency in #640 and asked the two tools to agree.

build now warns when a setting arrives without a value and shows the `-C KEY=` form that both tools accept. ✨ The shorthand keeps working, so existing scripts and pipelines keep running, and the warning moves people toward a spelling they can reuse with pip. The how-to guide, the backend explanation, and the CLI help now give the same guidance, so anyone who reads one of them lands on the portable form.

Someone relying on the bare form sees a single warning with the exact fix, and nothing they run today stops working. Closes #640.

Issue #640 has no comments; the discussion and the workaround live in the cibuildwheel thread it points to:

- [cibuildwheel#1513 (comment)](https://github.com/pypa/cibuildwheel/issues/1513#issuecomment-1591178953) — the portable form people settle on: append `=` so `-C--with-c-extensions=` is accepted by both build and pip.
- [cibuildwheel#1513 (comment)](https://github.com/pypa/cibuildwheel/issues/1513#issuecomment-1594343986) — the proposal to normalize everything to `key=value` pairs to paper over the mismatch.
- [pip `--config-settings` docs](https://pip.pypa.io/en/stable/cli/pip_install/) — pip's "Settings take the form `KEY=VALUE`" rule that forces the workaround in the first place.

Build scripts write config settings in the portable `--config-setting=KEY=VALUE` form that both build and pip accept, never the bare `-C KEY`:

- [Mercurial `build-one-linux-wheel.sh` (L23-L35)](https://github.com/saschazepter/hg/blob/c274076ac4fc65076f13e14c6d3756cd65966d3d/contrib/build-one-linux-wheel.sh#L23-L35) — passes the flag as `--config-setting=--global-option=--rust` to `python -m build`.
- [Streamlink `build-and-sign.sh` (L60)](https://github.com/streamlink/streamlink/blob/79324b9e8ae37650e71764590211a9b5f2d840cc/script/build-and-sign.sh#L60) — `python -m build ... --config-setting="--build-option=--plat-name=..."` with an explicit value.
- [LightGBM `build-python.sh` (L145)](https://github.com/lightgbm-org/LightGBM/blob/1c0970bbb764a140cca525f5c26e85bc01624779/build-python.sh#L145) — every setting carries an explicit `=`, even boolean flags like `cmake.define.USE_CUDA=ON`, fed to `python -m build` (L379).

### DIFF
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -538,8 +538,10 @@ def main_parser() -> argparse.ArgumentParser:
         action='append',
         help='settings to pass to the backend.  Multiple settings can be provided. '
         'Settings beginning with a hyphen will erroneously be interpreted as options to build if separated '
-        'by a space; use ``--config-setting=--my-setting -C--my-other-setting`` instead',
-        metavar='KEY[=VALUE]',
+        'by a space; use ``--config-setting=--my-setting -C--my-other-setting`` instead. '
+        'A setting passed without ``=VALUE`` is given an empty value, but this form is not supported by pip; '
+        'write ``KEY=`` instead for compatibility',
+        metavar='KEY=[VALUE]',
     )
     config_exclusive_group.add_argument(
         '--config-json',
@@ -582,15 +584,20 @@ def _parse_config_settings(raw_config_settings: list[str]) -> dict[str, str | li
     config_settings = dict[str, str | list[str]]()
 
     for arg in raw_config_settings:
-        setting, _, value = arg.partition('=')
-        if setting not in config_settings:
+        setting, sep, value = arg.partition('=')
+        if not sep:
+            warnings.warn(
+                f'Config setting {setting!r} was passed without a value; this form is not supported by pip. '
+                f'Write {setting}= instead to be compatible with both tools',
+                stacklevel=2,
+            )
+        existing = config_settings.get(setting)
+        if existing is None:
             config_settings[setting] = value
+        elif isinstance(existing, list):
+            existing.append(value)
         else:
-            existing = config_settings[setting]
-            if isinstance(existing, list):
-                existing.append(value)
-            else:
-                config_settings[setting] = [existing, value]
+            config_settings[setting] = [existing, value]
 
     return config_settings
 
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -97,7 +97,7 @@ def make_kwargs(
             ['--installer', 'uv'], (cwd, out), make_kwargs(installer='uv'), 'build_package_via_sdist', id='installer'
         ),
         pytest.param(
-            ['-C--flag1', '-C--flag2'],
+            ['-C--flag1=', '-C--flag2='],
             (cwd, out),
             make_kwargs(config_settings={'--flag1': '', '--flag2': ''}),
             'build_package_via_sdist',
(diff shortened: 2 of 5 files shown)


.github/workflows/cd.yml
```diff
@@ -26,7 +26,7 @@ jobs:
     permissions:
       contents: read # to checkout
     steps:
-      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
henryiii, line 29: Why the unrelated changes?

gaborbernat: For some reason, the ZIS-MOR was failing locally, and this fixed it. I can separate it.

henryiii: Making separate PRs would make review easier. :)

henryiii: Also seems like there's an unrelated refactor in test main?

gaborbernat: I would not say unrelated. I was extending the tests to validate this feature, so I fixed them to be more compact and fixed the not-so-well-defined type definitions.

src/build/__main__.py
```diff
@@ -524,7 +524,9 @@ def __call__(
         action='append',
         help='settings to pass to the backend.  Multiple settings can be provided. '
         'Settings beginning with a hyphen will erroneously be interpreted as options to build if separated '
-        'by a space; use ``--config-setting=--my-setting -C--my-other-setting`` instead',
+        'by a space; use ``--config-setting=--my-setting -C--my-other-setting`` instead. '
+        'A setting passed without ``=VALUE`` is given an empty value, but this form is not supported by pip; '
+        'write ``KEY=`` instead for compatibility',
         metavar='KEY[=VALUE]',
```
layday:
```suggestion
        metavar='KEY=[VALUE]',
```

tests/test_main.py
```diff
@@ -36,242 +36,112 @@
 
 ANSI_STRIP = re.compile(r'\x1B(?:[@-Z\\-_]|\[[0-?]*[ -/]*[@-~])')
 
+JSONValue = str | int | list['JSONValue'] | dict[str, 'JSONValue']
+
+
+class BuildKwargs(TypedDict):
```
layday, line 50: Seems a shame to have this here but leave the runtime untyped.

tests/test_main.py
```diff
@@ -885,6 +757,24 @@ def test_parse_config_settings_triple_duplicate() -> None:
...
+)
+def test_parse_config_settings_pip_compatibility(arg: str, value: str, warns: bool, recwarn: pytest.WarningsRecorder) -> None:
+    assert build.__main__._parse_config_settings([arg]) == {'--flag': value}
+
+    messages = [str(warning.message) for warning in recwarn]
+    if warns:
+        assert any("Config setting '--flag' was passed without a value" in message for message in messages)
+    else:
+        assert messages == []
+
```
layday: Either all of the warnings are config setting warnings or none of them are, right? This says that if there is a config setting warning, other warnings might come up too; but if there's no config setting warning, there must be no warnings at all.

.github/workflows/cd.yml
gaborbernat: Done — split the action pin-comment changes into #1091.

gaborbernat: Also split — the test_parse_args compaction (typed make_kwargs helper) is now its own PR, #1092. Once it lands the feature PRs can rebase onto it; this PR is back to just the warning + its test.

src/build/__main__.py
gaborbernat: Applied the suggestion (`KEY=[VALUE]`).

tests/test_main.py
gaborbernat: Moved out: the make_kwargs/BuildKwargs/JSONValue typing now lives in #1092, so this PR no longer adds typed test helpers on top of an untyped runtime.

tests/test_main.py
gaborbernat: Made it symmetric — it now checks the config-setting warning is present iff `warns`, and asserts no unrelated warnings appear in either case:
```python
config_warnings = [m for m in messages if "Config setting '--flag' was passed without a value" in m]
assert messages == config_warnings
assert bool(config_warnings) == warns
```

layday (approved):

PR #1082, opened 2026-06-03
✨ feat(deps): show wanted vs found in missing-deps error

### LINKED ISSUE #504: build fails with a bogus missing dependencies list
I am trying to build a wheel with build for the [scanpy](https://github.com/scverse/scanpy).

It prints this error message:
```
===>  Building for py39-scanpy-1.9.1
cd /disk-samsung/freebsd-ports/biology/py-scanpy/work/scanpy-1.9.1 && /usr/bin/env XDG_DATA_HOME=/disk-samsung/freebsd-ports/biology/py-scanpy/work  XDG_CONFIG_HOME=/disk-samsung/freebsd-ports/biology/py-scanpy/work  XDG_CACHE_HOME=/disk-samsung/freebsd-ports/biology/py-scanpy/work/.cache  HOME=/disk-samsung/freebsd-ports/biology/py-scanpy/work PATH=/disk-samsung/freebsd-ports/biology/py-scanpy/work/.bin:/home/yuri/bin:/sbin:/bin:/usr/sbin:/usr/bin:/usr/local/sbin:/usr/local/bin PKG_CONFIG_LIBDIR=/disk-samsung/freebsd-ports/biology/py-scanpy/work/.pkgconfig:/usr/local/libdata/pkgconfig:/usr/local/share/pkgconfig:/usr/libdata/pkgconfig NO_PIE=yes MK_DEBUG_FILES=no MK_KERNEL_SYMBOLS=no SHELL=/bin/sh NO_LINT=YES PREFIX=/usr/local  LOCALBASE=/usr/local  CC="cc" CFLAGS="-O2 -pipe -fno-omit-frame-pointer  -fstack-protector-strong -fno-strict-aliasing "  CPP="cpp" CPPFLAGS="-fno-omit-frame-pointer"  LDFLAGS=" -fstack-protector-strong " LIBS=""  CXX="c++" CXXFLAGS="-O2 -pipe -fno-omit-frame-pointer -fstack-protector-strong -fno-strict-aliasing -fno-omit-frame-pointer  "  MANPREFIX="/usr/local" BSD_INSTALL_PROGRAM="install  -s -m 555"  BSD_INSTALL_LIB="install  -s -m 0644"  BSD_INSTALL_SCRIPT="install  -m 555"  BSD_INSTALL_DATA="install  -m 0644"  BSD_INSTALL_MAN="install  -m 444" /usr/local/bin/python3.9 -m build --wheel --no-isolation --outdir /disk-samsung/freebsd-ports/biology/py-scanpy/work
* Getting dependencies for wheel...

ERROR Missing dependencies:
	anndata>=0.7.4
	matplotlib>=3.4
	kiwisolver>=1.0.1
	scikit-learn>=0.22
	threadpoolctl>=2.0.0
	umap-learn>=0.3.10
	pynndescent>=0.5 -> scikit-learn>=0.18 -> threadpoolctl>=2.0.0
	umap-learn>=0.3.10
	scikit-learn>=0.22 -> threadpoolctl>=2.0.0
	seaborn
	matplotlib>=2.2 -> kiwisolver>=1.0.1
*** Error code 1
```

Many dependencies that it lists are installed:
py39-anndata-0.8.0
py39-matplotlib-3.4.3_3
py39-kiwisolver-1.4.3,1
py39-umap-learn-0.5.3
py39-seaborn-0.11.2

There is one real problem: threadpoolctl is missing.

But why does build complain about dependencies that are installed?

py39-build-0.8.0
FreeBSD 13.1

gaborbernat:
Packagers and offline builders who run `python -m build --no-isolation` rely on the dependency check to tell them what is missing before a build wastes their time. Today that check lists requirement names and nothing else, so a dependency that is installed at the wrong version, or installed for a different Python than the one running build, still gets reported as simply "missing." In issue #504 a FreeBSD packager stepped through build with a debugger only to discover the package was present the whole time, invisible to the interpreter build was using.

To get past this today people reach for `--skip-dependency-check` and hope nothing is genuinely absent, run `pip list` by hand against each name in the error, or guess which interpreter build actually inspected. None of those answer the real question: what does build see, and why is it not enough?

This change makes the error answer that question. 📦 For every unmet requirement it shows the version that was wanted next to the version that was found (or that the package is not installed), and it names the interpreter the check ran against. A wrong-version dependency now reads as `found: 1.0.0` against `wanted: >=2.0`, and a dependency installed for another Python reads as `found: not installed` beside an interpreter path the user recognizes. ✨ What used to need a debugger is now visible at a glance.

Closes #504.

How people diagnose a present-but-unrecognised dependency until this lands:

- [#504 (comment)](https://github.com/pypa/build/issues/504#issuecomment-2383365311) — step through `build` under `pdb` to discover a dependency was installed but at the wrong version.
- [#504 (comment)](https://github.com/pypa/build/issues/504#issuecomment-3994378332) — fall back to `--skip-dependency-check`, then `pip list` by hand and `-vvv` for more detail.
- [#504 (comment)](https://github.com/pypa/build/issues/504#issuecomment-4524877911) — manually reason about which interpreter the check actually ran against versus the caller's environment.
- [pip#9794](https://github.com/pypa/pip/issues/9794) — the companion request on the pip side to validate build dependencies under `--no-build-isolation`.

Packagers reach for `--skip-dependency-check` to get past the opaque error, accepting they must vet the deps themselves:

- [zeromq/pyzmq cross-compile Dockerfile (L98)](https://github.com/zeromq/pyzmq/blob/73f49bb18588e09dee4e76abb5a4a1beeb418fb4/docs/source/howto/cross.Dockerfile#L98) — `cross-python -m build --no-isolation --skip-dependency-check --wheel`, the wrong-interpreter case where the check would inspect the host Python, not the cross one.
- [HaikuPorts `build` recipe (L90)](https://github.com/haikuports/haikuports/blob/a68cac38b2e05c29b0629826963735b94f198b99/dev-python/build/build-1.5.0.recipe#L90) — `$python -m build --no-isolation --skip-dependency-check` across several interpreters in a loop.
- [FreedomBox `Makefile` (L74)](https://github.com/freedombox/FreedomBox/blob/a7584b465dadd1909abc0ac1039927b1ddf44d6c/Makefile#L74) — `python -m build --no-isolation --skip-dependency-check --wheel` to bypass the check rather than decode it.

### DIFF
--- a/src/build/_util.py
+++ b/src/build/_util.py
@@ -74,7 +74,7 @@ def check_dependency(
 
 def format_unmet_dependencies(unmet: AbstractSet[tuple[str, ...]]) -> str:
     body = ''.join(_format_missing_dependency(chain) for chain in sorted(unmet))
-    return f'Missing dependencies (checked against {sys.executable}):{body}'
+    return f'Unmet dependencies (checked against {sys.executable}):{body}'
 
 
 def _format_missing_dependency(dep_chain: tuple[str, ...]) -> str:
--- a/src/build/_util.py
+++ b/src/build/_util.py
@@ -1,6 +1,11 @@
 from __future__ import annotations
 
 import re
+import sys
+
+import packaging.requirements
+
+from ._compat import importlib
 
 
 TYPE_CHECKING = False
@@ -27,10 +32,6 @@ def check_dependency(
     :param parent_extras: Extras (eg. "test" in myproject[test])
     :yields: Unmet dependencies
     """
-    import packaging.requirements
-
-    from ._compat import importlib
-
     req = packaging.requirements.Requirement(req_string)
     normalised_req_string = str(req)
 
@@ -62,6 +63,25 @@ def check_dependency(
             for other_req_string in dist.requires:
                 # yields transitive dependencies that are not satisfied.
                 yield from check_dependency(other_req_string, (*ancestral_req_strings, normalised_req_string), req.extras)
+
+
+def format_unmet_dependencies(unmet: AbstractSet[tuple[str, ...]]) -> str:
+    body = ''.join(_format_missing_dependency(chain) for chain in sorted(unmet))
+    return f'Missing dependencies (checked against {sys.executable}):{body}'
+
+
+def _format_missing_dependency(dep_chain: tuple[str, ...]) -> str:
+    requirement = packaging.requirements.Requirement(dep_chain[-1])
+    try:
+        found = importlib.metadata.distribution(requirement.name).version
+    except importlib.metadata.PackageNotFoundError:
+        found = 'not installed'
+    wanted = str(requirement.specifier) if requirement.specifier else 'any'
+    return f'\n\t{_format_dep_chain(dep_chain)}\n\t\twanted: {wanted}\n\t\tfound: {found}'
+
+
+def _format_dep_chain(dep_chain: tuple[str, ...]) -> str:
+    return ' -> '.join(dep.partition(';')[0].strip() for dep in dep_chain)
 
 
 def parse_wheel_filename(filename: str) -> re.Match[str] | None:
--- a/src/build/_util.py
+++ b/src/build/_util.py
@@ -1,5 +1,12 @@
 from __future__ import annotations
 
+
+__lazy_modules__ = [
+    f'{__spec__.parent}._compat',
+    'packaging',
+    'packaging.requirements',
+]
+
 import re
 import sys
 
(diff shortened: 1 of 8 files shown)


src/build/_util.py
```diff
@@ -64,5 +72,24 @@ def check_dependency(
                 yield from check_dependency(other_req_string, (*ancestral_req_strings, normalised_req_string), req.extras)
 
 
+def format_unmet_dependencies(unmet: AbstractSet[tuple[str, ...]]) -> str:
+    body = ''.join(_format_missing_dependency(chain) for chain in sorted(unmet))
+    return f'Missing dependencies (checked against {sys.executable}):{body}'
```
layday: I think "unmet" is a better qualifier:

```suggestion
    return f'Unmet dependencies (checked against {sys.executable}):{body}'
```

gaborbernat: Agreed — renamed to `Unmet dependencies` (message, docs and tests). Thanks.

layday (approved):

Issue #613, opened 2023-05-07
Can't build with `--no-isolation` without installing the builder first


mjpieters:
Minimal reproducable test case using this project itself:

```
# setup; temp dir with venv and build 0.10.0 source
mkdir /tmp/testnonisolated
cd /tmp/testnonisolated
git clone --depth 1 --branch 0.10.0 https://github.com/pypa/build.git
python -m venv venv
source venv/bin/activate
# install build with pip (isolated build)
pip install build/
# attempt to use build to build build (say that 10 times really fast)
python -m build --no-isolation build
```

This fails with

```
* Getting build dependencies for sdist...

Traceback (most recent call last):
  File "/private/tmp/testnonisolated/venv/lib/python3.11/site-packages/pyproject_hooks/_impl.py", line 321, in _call_hook
    raise BackendUnavailable(data.get('traceback', ''))
pyproject_hooks._impl.BackendUnavailable: Traceback (most recent call last):
  File "/private/tmp/testnonisolated/venv/lib/python3.11/site-packages/pyproject_hooks/_in_process/_in_process.py", line 77, in _build_backend
    obj = import_module(mod_path)
          ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/mj/.pyenv/versions/3.11.2/lib/python3.11/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1206, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1178, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1128, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1206, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1178, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1142, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'flit_core'

ERROR Backend 'flit_core.buildapi' is not available.
```

[...]

Issue #614, opened 2023-05-10
Add option to specify where the sdist tarfile should be expanded


cquike:
Currently, if no -s or -w option is given, a sdist is created which is then extracted in a /tmp/build-sdist-xxx directory. It would be useful to have an option to specify _this_ directory. That's important when compiling large C extensions which can benefit from using ccache. When a random temporary directory is created then ccache is of not use.

In other words, my request is to have an option to specify the directory sdist_out in this line:  https://github.com/pypa/build/blob/f5d5f15f1e8a4e501ad5044cd2d82e0c7995171c/src/build/__main__.py#L238 . Obviously the default should be the current behaviour and only when the option is specified that directory should be used.

I know that the current -w option helps in that sense, but it is desirable to have off-source builds since the extension compilation can potentially create many files and mess around with the sources.

layday:
Ah right, that makes more sense - you don't actually want to keep `foo` around, you just want the path to be deterministic between builds in order to hit the cache.  Hmm, I don't know if adding a new option is the best option.  Perhaps build can default to using `$TMPDIR/pyproject-build/$(basename $PWD)` if it's unoccupied and fall back on `mkdtemp` and emit a warning if it isn't?

PR #1083, opened 2026-06-03
✨ feat(env): control the isolated build environment location

### LINKED ISSUE #655: Controlling location of isolated build environments
Is there a way to control exactly where isolated build environments are created?

For example, I'd like to ask python-build to create its isolated build environments inside `.local/tmp/python-build` inside my project.

Sorry if this is documented somewhere and I didn't see it.

gaborbernat:
Today `build` puts every isolated build environment in a random temporary directory and removes it the moment the build finishes. People who need a say in that location have no good option. Some set `TMPDIR`, which moves the parent folder but keeps the random name and still wipes it; others give up on isolation and assemble a virtual environment by hand. #655 collects the three situations where this hurts.

🛠️ Teams building compiled projects want their compilation cache to pay off. ccache and sccache decide what to reuse from the paths in each compiler command, so a build-environment path that changes on every run looks brand new and the cache stays cold. A stable location turns those misses into hits.

📦 People building inside containers often have a read-only filesystem with one small writable corner. They need the environment to land in that corner.

🔍 Anyone debugging a failed build wants to open the environment afterward and see what went in. A directory that disappears on exit takes the evidence with it.

This change lets you choose the location with `--env-dir` (or the `BUILD_ENV_DIR` environment variable), and exposes the same `path` setting on the API. You point it at an empty folder, `build` fills it, clears it after a clean build so the next run can reuse the spot, and leaves it in place after a failure so you can look inside. The documentation also corrects an older claim that the temporary directory survives a failed build.

Closes #655.

The stopgaps people reach for in the issue thread and across the toolchain until this lands:

- [#655 (comment)](https://github.com/pypa/build/issues/655#issuecomment-1883525892) — set `TMPDIR`/`TEMP`/`TMP` to steer the env into a chosen (or writable) parent folder.
- [#655 (comment)](https://github.com/pypa/build/issues/655#issuecomment-1884132233) — the requester reports `TMPDIR` had no effect, which is exactly why a dedicated location flag is needed.
- [#655 (comment)](https://github.com/pypa/build/issues/655#issuecomment-2411875026) — RAPIDS describes every compile missing the sccache cache because the random env path appears in the compile line.
- [#655 (comment)](https://github.com/pypa/build/issues/655#issuecomment-2412046685) — confirmation that it is the build venv (not sdist expansion) that needs a stable path for scikit-build-core / meson-python.
- [ccache `base_dir`](https://ccache.dev/manual/latest.html#config_base_dir) / [sccache `SCCACHE_BASEDIR`](https://github.com/mozilla/sccache/blob/main/docs/Configuration.md) — strip the volatile path before hashing (a partial mitigation, not a fix for the container or debugging cases).
- [#613](https://github.com/pypa/build/issues/613) — build a venv by hand and use `python -m build --no-isolation` to get a controllable, persistent environment.

Projects steer the build environment by hand — relocating the temp dir or skipping isolation — to keep a compiler cache warm:

- [PyTorch `macos-build.sh` (L26-L47)](https://github.com/pytorch/pytorch/blob/35c82ffe9178a10ae833a4628bf5bd1e1d7efe9e/.ci/pytorch/macos-build.sh#L26-L47) — install sccache shims on `PATH`, then `python -m build --wheel --no-isolation` so the cache sees a stable environment.
- [Apache Arrow `python_wheel_macos_build.sh` (L170)](https://github.com/apache/arrow/blob/5896c13e31c3993cba8b182c12a7c2d5f01f1ae9/ci/scripts/python_wheel_macos_build.sh#L170) — `python -m build --sdist --wheel . --no-isolation` with `-DARROW_USE_CCACHE=ON` (L124).
- [ecmwf/tensogram release CI (L14-L15)](https://github.com/ecmwf/tensogram/blob/d90b083e2764dcea8f9bb12fe772587fbd4daf2b/.github/workflows/release-preflight.yml#L14-L15) — point `TMPDIR` at a writable `$GITHUB_WORKSPACE/.tmp` (created at L40) so `python -m build` (L204) lands in the one writable spot, with sccache stats at L217.
- [MSYS2 `mingw-w64-meson` PKGBUILD (L67)](https://github.com/msys2/MINGW-packages/blob/694602d7796194e979434787ac51fc626b785841/mingw-w64-meson/PKGBUILD#L67) — `python -m build --wheel --skip-dependency-check --no-isolation` so the wheel builds in the prepared env where ccache/sccache (L26-L27) live.

### DIFF
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -88,6 +88,7 @@ if TYPE_CHECKING:
         dependency_constraints_txt: str | None
         skip_dependency_check: bool
         sdist_extract_dir: str | None
+        env_dir: str | None
 
 
 _COLORS = {
@@ -197,10 +198,11 @@ def _bootstrap_build_env(
     skip_dependency_check: bool,
     dependency_constraints_txt: os.PathLike[str] | None,
     installer: _env.Installer,
+    env_dir: str | None = None,
     runner: SubprocessRunner | None = None,
 ) -> Iterator[ProjectBuilder]:
     if isolation:
-        with DefaultIsolatedEnv(installer=installer) as env:
+        with DefaultIsolatedEnv(installer=installer, path=env_dir) as env:
             make_builder = partial(ProjectBuilder.from_isolated_env, env, srcdir)
             if runner:
                 make_builder = partial(make_builder, runner=runner)
@@ -252,6 +254,7 @@ def _build(
     skip_dependency_check: bool,
     dependency_constraints_txt: os.PathLike[str] | None,
     installer: _env.Installer,
+    env_dir: str | None = None,
 ) -> str:
     with _bootstrap_build_env(
         isolation,
@@ -261,6 +264,7 @@ def _build(
         skip_dependency_check,
         dependency_constraints_txt,
         installer,
+        env_dir,
         pyproject_hooks.quiet_subprocess_runner if _ctx.verbosity < 0 else None,
     ) as builder:
         return builder.build(distribution, outdir, config_settings)
@@ -311,6 +315,7 @@ def build_package(
     skip_dependency_check: bool = False,
     dependency_constraints_txt: os.PathLike[str] | None = None,
     installer: _env.Installer = 'pip',
+    env_dir: str | None = None,
 ) -> list[str]:
     """
     Run the build process.
@@ -321,6 +326,7 @@ def build_package(
     :param config_settings: Configuration settings to be passed to the backend
     :param isolation: Isolate the build in a separate environment
     :param skip_dependency_check: Do not perform the dependency check
+    :param env_dir: Location of the isolated build environment (a temporary directory is used when not set)
     """
     built: list[str] = []
     for distribution in distributions:
@@ -333,6 +339,7 @@ def build_package(
             skip_dependency_check,
             dependency_constraints_txt,
             installer,
+            env_dir,
         )
         built.append(os.path.basename(out))
     return built
@@ -348,6 +355,7 @@ def build_package_via_sdist(
     dependency_constraints_txt: os.PathLike[str] | None = None,
     installer: _env.Installer = 'pip',
     sdist_extract_dir: StrPath | None = None,
+    env_dir: str | None = None,
 ) -> list[str]:
     """
     Build a sdist and then the specified distributions from it.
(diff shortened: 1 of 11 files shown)


src/build/env.py
```diff
@@ -105,12 +105,22 @@ def __init__(
...
         self.installer: Installer = installer
+        self._requested_path = path
 
     def __enter__(self) -> Self:
-        try:
+        if self._requested_path is None:
             path = tempfile.mkdtemp(prefix='build-env-')
+        else:
+            path = self._requested_path
+            if os.path.isdir(path) and os.listdir(path):
```
layday: Use `os.scandir` perhaps to avoid traversing the entire directory.

src/build/env.py
```diff
@@ -105,12 +105,22 @@ def __init__(
...
+                msg = f'Build environment location is not empty: {path}'
+                raise BuildException(msg)
```
layday: `BuildException` says it's for the `ProjectBuilder` - maybe we should reword its description if we're using it as a generic error produced anywhere by `build`?

src/build/__main__.py
```diff
@@ -563,6 +583,14 @@ def __call__(
         help='disable building the project in an isolated virtual environment. '
         'Build dependencies must be installed separately when this option is used',
     )
+    install_group.add_argument(
+        '--env-dir',
+        default=os.environ.get('BUILD_ENV_DIR'),
```
layday: Why are we also exposing this as an env var?

henryiii: Also note we didn't add an envvar for #614, which I see as rather similar.

gaborbernat: Agreed, and good consistency point. Dropped `BUILD_ENV_DIR` in the latest push — `--env-dir` is now flag-only, matching `--sdist-extract-dir` from #614. If we ever want env-var control for both, we can add it uniformly in a follow-up.

src/build/__main__.py
```diff
@@ -625,6 +651,9 @@ def main(cli_args: Sequence[str], prog: str | None = None) -> None:
         parser.prog = prog
     args = cast('_Args', parser.parse_args(cli_args))
 
+    if args.env_dir is not None and args.no_isolation:
+        parser.error('--env-dir: not allowed with --no-isolation')
+
```
henryiii, line 671: Can't we use `install_group.add_mutually_exclusive_group` instead of manually checking this?

src/build/env.py
gaborbernat: Done — switched to `os.scandir` with a `next(entries, None)` emptiness probe, so we stop at the first entry instead of listing the whole directory.

src/build/env.py
gaborbernat: Reworded its docstring to "Exception raised by ``build`` when a build cannot proceed" — it stopped being `ProjectBuilder`-specific the moment the env layer started raising it too.

src/build/__main__.py
gaborbernat: I tried that first, but argparse can't express this one: `--no-isolation` already lives in the mutually exclusive group with `--installer`, and an action can't belong to two groups. Folding `--env-dir` into that existing group would also forbid `--env-dir --installer uv`, which is a valid (arguably the primary) combination — env location and installer choice are orthogonal. The exclusivity we need is pairwise (`env-dir` × `no-isolation` only), so the manual `parser.error` check stays.

henryiii (approved):

PR #1084, opened 2026-06-04
⚠️ feat(util): deprecate project_wheel_metadata

### LINKED ISSUE #557: Validate dependencies in non-isolated builds
Currently, if a non-isolated build is invoked, it simply attempts to invoke the build hook expecting the invoker to have supplied the dependencies. But if the environment doesn't have the dependencies and if the lack of dependencies doesn't create a failure in the build, the build proceeds without its declared dependencies (see https://github.com/jaraco/jaraco.packaging/pull/9#issuecomment-1261433024 for an example).

Better would be for build to enforce that the declared dependencies are met before invoking the build.

gaborbernat:
Closes #557.

Teams that read a project's metadata without a full isolated build, for example a docs site that needs the current version or a packager pulling the name and dependencies, hit a quiet failure today. When they call `project_wheel_metadata` with `isolated=False`, build trusts whatever happens to be in the current environment and runs the backend anyway. If a declared build dependency is absent, the backend does not fail; it returns a wrong answer. The version comes back as `0.0.0` when something like `setuptools_scm` is not installed, and nobody notices until a release looks broken. 📉

What people do today is pick the lesser of two problems. They either pay for a throwaway virtualenv on every metadata read, which rules out offline and air-gapped machines and slows docs builds, or they switch to `isolated=False` and accept that the result might be silently wrong. The original report walked through exactly this: a documentation build moved to the non-isolated path to work offline, and the version it recorded dropped from a real value to `0.0.0` with no warning.

This adds `build.util.wheel_metadata`, which returns a project's metadata as a parsed, JSON-serialisable mapping — the same structure `python -m build --metadata` prints — and takes a `check_dependencies` flag. ✅ With `isolated=False` and `check_dependencies=True`, build confirms the declared build dependencies are present against the interpreter actually running, and stops with a clear `DependencyError` when one is missing, instead of returning a misleading result. The command line already runs this check under `--no-isolation`, so the two now behave the same way.

`project_wheel_metadata` is deprecated in favour of it. As [noted in review](https://github.com/pypa/build/pull/1084#issuecomment-4636476990), it returns an `email.message.Message`-style object and had fallen behind the CLI, which already parses metadata into the friendlier JSON structure; rather than grow a second function the right move was to expose that capability and retire the old one. 📦

When a check fails, the message names the interpreter it inspected and, for each requirement, the version that was wanted against the version that was found. That tells a user whether a package is genuinely absent or merely the wrong version, and it points at the common gotcha where a dependency was installed for a different Python than the one running build.

How people cope with silent wrong metadata on the non-isolated path until this lands:

- [jaraco.packaging#9 (comment)](https://github.com/jaraco/jaraco.packaging/pull/9#issuecomment-1261433024) — the canonical reproduction: a missing `setuptools_scm` makes the recorded version collapse from a real value to `0.0.0` with no error.
- [#556 (comment)](https://github.com/pypa/build/issues/556#issuecomment-2028942951) — set `BUILD_ENVIRONMENT=current` (jaraco.packaging ≥ 9.5) to read metadata against the current interpreter.
- [#556 (comment)](https://github.com/pypa/build/issues/556#issuecomment-2028797519) — pre-stage build deps as wheels and force pip offline with `PIP_NO_INDEX=1 PIP_FIND_LINKS=.`.
- [jaraco.packaging#11 (comment)](https://github.com/jaraco/jaraco.packaging/pull/11#issuecomment-1627557237) — a per-library env var to bypass isolation, the wrong-layer patch that motivated fixing `build` itself.
- [#557 (comment)](https://github.com/pypa/build/issues/557#issuecomment-1367957108) / [clarification](https://github.com/pypa/build/issues/557#issuecomment-1367963519) — pass `--no-isolation -x`, accepting that `project_wheel_metadata` does no dependency check at all.
- [pytest-checkdocs#19 (comment)](https://github.com/jaraco/pytest-checkdocs/issues/19#issuecomment-1627556931) — a downstream consumer left broken by the per-project patch, showing why the fix belongs in `build`.

Packagers reading metadata or building without isolation pre-provide the build deps by hand:

- [openSUSE `python-jaraco.stream.spec` (L62-L64)](https://github.com/bmwiedemann/openSUSE/blob/b6b49ac3dc7e59c26e54a6e458aac232879ea35e/packages/p/python-jaraco.stream/python-jaraco.stream.spec#L62-L64) — `export BUILD_ENVIRONMENT=current` so the Sphinx metadata read uses the rpm `BuildRequires` instead of an isolated build.
- [Alpine aports `twine` APKBUILD (L48)](https://github.com/universalis-llc/aports/blob/178d1598f63c23f299986f0206889e447b91f252/community/twine/APKBUILD#L48) — `python3 -m build --no-isolation --skip-dependency-check`, with `py3-setuptools_scm` hand-added to `makedepends` (L24-L27).
- [MacPorts `apache-arrow` Portfile (L353)](https://github.com/macports/macports-ports/blob/524d5d3db70ca75e9747cbf19623e8a23f722301/devel/apache-arrow/Portfile#L353) — `-m build --no-isolation` paired with a manually listed `setuptools_scm` build dep (L250).
- [tqcli packaging runbook (L444)](https://github.com/tqcli/tqcli/blob/594eb1903605a451c24b56410fdf2c764792ef31/docs/prompts/ship_turboquant_wheels.md#L444) — writes down that under `--no-isolation` the venv must contain every `[build-system].requires` entry, after a silent failure from the missing ones.

### DIFF
--- a/docs/changelog/557.removal.rst
+++ b/docs/changelog/557.removal.rst
@@ -1,0 +1,2 @@
+Deprecate :func:`build.util.project_wheel_metadata`; it returns an ``email.message.Message``-style object and has fallen
+behind the CLI. Generate metadata with the ``python -m build --metadata`` command instead - by :user:`gaborbernat`.
--- a/src/build/util.py
+++ b/src/build/util.py
@@ -3,10 +3,11 @@
 from __future__ import annotations
 
 
-__lazy_modules__ = [f'{__spec__.parent}._compat', f'{__spec__.parent}.env', 'pathlib', 'tempfile']
+__lazy_modules__ = [f'{__spec__.parent}._compat', f'{__spec__.parent}.env', 'pathlib', 'tempfile', 'warnings']
 
 import pathlib
 import tempfile
+import warnings
 
 import pyproject_hooks
 
@@ -40,13 +41,20 @@ def project_wheel_metadata(
     Uses the ``prepare_metadata_for_build_wheel`` hook if available,
     otherwise ``build_wheel``.
 
+    .. deprecated:: 1.5.0
+       Generate metadata with the ``python -m build --metadata`` command instead.
+
     :param source_dir: Project source directory
     :param isolated: Whether or not to run invoke the backend in the current
                      environment or to create an isolated one and invoke it
                      there.
     :param runner: An alternative runner for backend subprocesses
     """
-
+    warnings.warn(
+        'project_wheel_metadata is deprecated; generate metadata with the `python -m build --metadata` command instead',
+        DeprecationWarning,
+        stacklevel=2,
+    )
     if isolated:
         with DefaultIsolatedEnv() as env:
             builder = ProjectBuilder.from_isolated_env(
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -12,6 +12,9 @@ import pytest_mock
 import build.util
 
 
+pytestmark = pytest.mark.filterwarnings('ignore:project_wheel_metadata is deprecated:DeprecationWarning')
+
+
 @pytest.mark.pypy3323bug
 @pytest.mark.parametrize('isolated', [False, pytest.param(True, marks=[pytest.mark.network, pytest.mark.isolated])])
 def test_wheel_metadata(package_test_setuptools: str, isolated: bool) -> None:
@@ -74,3 +77,11 @@ def test_project_wheel_metadata_installs_build_requires_fresh(mocker: pytest_moc
         mocker.call({'dep1'}, _fresh=True),
         mocker.call({'dep2'}),
     ]
+
+
+def test_project_wheel_metadata_is_deprecated(mocker: pytest_mock.MockerFixture) -> None:
+    result = mocker.patch('build.util._project_wheel_metadata')
+    mocker.patch('build.util.ProjectBuilder')
+
+    with pytest.warns(DeprecationWarning, match=r'python -m build --metadata'):
+        assert build.util.project_wheel_metadata('/tmp/project', isolated=False) is result.return_value


layday:
My preference would be to deprecate and eventually remove `project_wheel_metadata`.  This function has fallen behind the CLI's implementation (which parses the metadata into a user-friendlier JSON structure) and was conceived at a time when we were reluctant to extend the CLI with a metadata extraction command. `__main__._build_metadata` can be made public to compensate.

gaborbernat:
Done. Added `build.util.wheel_metadata`, which returns the parsed, JSON-serialisable metadata the `--metadata` CLI prints (a `packaging.metadata.RawMetadata`), and moved the `check_dependencies` option onto it. Deprecated `project_wheel_metadata` (DeprecationWarning + changelog removal note) pointing at the new function.

I exposed it as `build.util.wheel_metadata` rather than literally promoting `__main__._build_metadata`, to keep the public surface in `build.util` next to the function it replaces and avoid leaking the CLI-only knobs (installer, constraints, config-settings). Happy to rename or relocate if you'd prefer a different home.

layday:
No, I think if we aren't committing to providing utility functions on top of the builder and env, we shouldn't have one just for metadata.  Now that the CLI can generate metadata, there isn't anything special about this function when compared to the other CLI command functions.  Perhaps they *should* all have programmatic equivalents, but that'd probably take quite a bit of refactoring so that we don't end up letting the implementations drift again.

gaborbernat:
So what do you think we should do today then? 🤔 close this and the related issue as won't fix? 🤔

layday:
That we should simply deprecate `project_wheel_metadata` - let me know if you think otherwise though.

gaborbernat:
And what should we guide users towards, using the cli?

layday (approved):

astrojuanlu:
Doesn't this mean that now there's no public, programmatic, non-deprecated API to get the metadata?

Promoting `__main__._build_metadata` doesn't seem very useful without some code changes, it's clearly made for the CLI:

https://github.com/pypa/build/blob/43c337a6bf55cbf2737bf938f3968be46ebc48b7/src/build/__main__.py#L477-L479

astrojuanlu:
In the meantime, I rescued @gaborbernat 's implementation from the `git push --force` abyss and preserved it in https://github.com/astrojuanlu/pypa-build/commit/701775ce7796ccc165d81559dab4612fe9fb64f3

gaborbernat:
> No, I think if we aren't committing to providing utility functions on top of the builder and env, we shouldn't have one just for metadata. Now that the CLI can generate metadata, there isn't anything special about this function when compared to the other CLI command functions. Perhaps they _should_ all have programmatic equivalents, but that'd probably take quite a bit of refactoring so that we don't end up letting the implementations drift again.

We removed it per this idea 😊

PR #1085, opened 2026-06-04
✨ feat(cli): add --sdist-extract-dir to reuse compiler caches

### LINKED ISSUE #614: Add option to specify where the sdist tarfile should be expanded

gaborbernat:
Teams shipping packages with large C/C++ extensions wait minutes on every rebuild, even when they only touched one file. The default build produces an sdist, extracts it to a random temporary directory, and builds the wheel from there. Compiler caches like `ccache` and `sccache` decide what to reuse from the absolute path of each source file, so a new random path on every run looks like brand-new code and nothing is reused. Closes #614.

Right now the only way out is to drop the convenient one-shot build and run the steps by hand: build the sdist, untar it into a fixed folder, then point a wheel build at that folder, repeating it after every code change. People also reach for `TMPDIR`, but that only moves the parent folder and still adds a random suffix, so the cache keeps missing.

This adds `--sdist-extract-dir`, which lets you choose a stable folder for that intermediate sdist. 🚀 Reuse the same folder across rebuilds and the cache recognizes unchanged files, so only what you actually edited gets recompiled. The familiar `python -m build` workflow stays intact.

The random temporary directory remains the default on purpose, so everyday builds stay isolated and still catch backends that hard-code paths. The new behavior is opt-in: name a folder and build refreshes its contents on each run and keeps it for the next one. The flag also applies when building a wheel straight from a `.tar.gz`.

One thing to set expectations: the isolated build environment still lives at a random path, so packages whose compile lines embed environment include paths (Python or dependency headers) need toolchain-side path remapping on top of this to get full cache hits. Pinning that environment is tracked separately in #655.

These are the stopgaps people reach for in the issue thread and across the toolchain until this lands:

- [#614 (comment)](https://github.com/pypa/build/issues/614#issuecomment-1542388823) — run the steps by hand: `build -s`, untar into a fixed folder, then `build -w` against it, repeated after every edit.
- [#614 (comment)](https://github.com/pypa/build/issues/614#issuecomment-1542406647) — the requester spells out that all they need is a *deterministic* extraction path so the cache key stays stable.
- [#614 (comment)](https://github.com/pypa/build/issues/614#issuecomment-1542455745) — a maintainer's alternative: derive a predictable `$TMPDIR/...` location instead of a random one.
- [ccache `base_dir` / `CCACHE_BASEDIR`](https://ccache.dev/manual/latest.html#config_base_dir) and the [ccache discussion on same-checkout-different-location misses](https://github.com/ccache/ccache/discussions/1370) — rewrite absolute paths before hashing so a moving directory still hits.
- [sccache `SCCACHE_BASEDIR`](https://github.com/mozilla/sccache/blob/main/docs/Configuration.md) — the sccache equivalent for normalizing build paths.

Projects building C/C++ extensions with `python -m build` already script the pin-the-path dance by hand:

- [Arch `python-zstandard` PKGBUILD (L25-L63)](https://github.com/nisavid/arch-strix-halo-pkgs/blob/ab1fff54105f811ca00fc053a2a7b4236600bcf1/packages/python-zstandard-gfx1151/PKGBUILD#L25-L63) — extract the sdist into a fixed `$srcdir`, `export CCACHE_BASEDIR="$srcdir"`, then `python -m build --wheel --no-isolation`.
- [PEP-Repository CI `.gitlab-ci.yml` (L376)](https://github.com/PEP-Repository/core/blob/d8aa9eed4da12031337f3e7435c3b48027d20326/.gitlab-ci.yml#L376) — `export CCACHE_BASEDIR="$PWD"` before `pipx run build --wheel` (L825) so ccache hits survive build's path handling.

### DIFF
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -87,6 +87,7 @@ if TYPE_CHECKING:
         no_isolation: bool
         dependency_constraints_txt: str | None
         skip_dependency_check: bool
+        sdist_extract_dir: str | None
 
 
 _COLORS = {
@@ -331,6 +332,7 @@ def build_package_via_sdist(
     skip_dependency_check: bool = False,
     dependency_constraints_txt: os.PathLike[str] | None = None,
     installer: _env.Installer = 'pip',
+    sdist_extract_dir: StrPath | None = None,
 ) -> list[str]:
     """
     Build a sdist and then the specified distributions from it.
@@ -341,6 +343,8 @@ def build_package_via_sdist(
     :param config_settings: Configuration settings to be passed to the backend
     :param isolation: Isolate the build in a separate environment
     :param skip_dependency_check: Do not perform the dependency check
+    :param sdist_extract_dir: Directory to extract the intermediate sdist into; a temporary directory is used and
+        removed afterwards when ``None``
     """
     if 'sdist' in distributions:
         msg = 'Only binary distributions are allowed but sdist was specified'
@@ -353,7 +357,7 @@ def build_package_via_sdist(
     sdist_name = os.path.basename(sdist)
     built: list[str] = []
     if distributions:
-        with _extract_sdist(sdist, sdist_name.removesuffix('.tar.gz')) as extracted_srcdir:
+        with _extract_sdist(sdist, sdist_name.removesuffix('.tar.gz'), extract_dir=sdist_extract_dir) as extracted_srcdir:
             _ctx.log(f'Building {_natural_language_list(distributions)} from sdist', kind=('step',))
             for distribution in distributions:
                 out = _build(
@@ -510,6 +514,13 @@ def main_parser() -> argparse.ArgumentParser:
         metavar='PATH',
     )
     build_group.add_argument(
+        '--sdist-extract-dir',
+        help='extract the intermediate sdist to PATH (created if missing and kept afterwards) instead of a random '
+        'temporary directory; reusing it across rebuilds gives compiler caches such as ccache/sccache a stable '
+        'source path. Only affects the default (via-sdist) build and building a wheel from an sdist',
+        metavar='PATH',
+    )
+    build_group.add_argument(
         '--sdist',
         '-s',
         dest='distributions',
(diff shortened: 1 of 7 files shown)


src/build/__main__.py
```diff
@@ -689,14 +699,23 @@ def _validate_sdist_archive(archive: StrPath) -> str:
 
 
 @contextlib.contextmanager
-def _extract_sdist(archive: StrPath, top_level: str) -> Iterator[str]:
-    extract_dir = tempfile.mkdtemp(prefix='build-via-sdist-')
-    try:
+def _extract_sdist(archive: StrPath, top_level: str, extract_dir: StrPath | None = None) -> Iterator[str]:
```
```suggestion
def _extract_sdist(archive: StrPath, top_level: str, *, extract_dir: StrPath | None = None) -> Iterator[str]:
```

There are multiple dir-like things here, let's enforce at least this one is by name.

henryiii (commented):
Minor nit. Otherwise looks fine.

src/build/__main__.py
```diff
@@ -339,7 +342,7 @@ def build_package_via_sdist(
     sdist_name = os.path.basename(sdist)
     built: list[str] = []
     if distributions:
-        with _extract_sdist(sdist, sdist_name.removesuffix('.tar.gz')) as extracted_srcdir:
+        with _extract_sdist(sdist, sdist_name.removesuffix('.tar.gz'), sdist_extract_dir) as extracted_srcdir:
```
```suggestion
        with _extract_sdist(sdist, sdist_name.removesuffix('.tar.gz'), extract_dir=sdist_extract_dir) as extracted_srcdir:
```

henryiii (approved):

PR #1086, opened 2026-06-04
✨ feat: report resolved build dependency versions

### LINKED ISSUE #959: Show build backend and backend commands in verbosity mode
I would enjoy if `build` had some machine readable logging about how it interacts with backend, starting with backend and being used. Something like this:

```
$ python -m build --wheel -v
build:backend: setuptools
...
```

gaborbernat:
When a build breaks, the usual next step is a report on the backend's issue tracker, and those trackers ask for the exact backend version you ran. That version is awkward to find: `pyproject.toml` lists only the requirement specifiers (`setuptools >= 40.8.0`), and on an isolated build the resolved versions live in the temporary environment that build deletes when it finishes. The installer does print them, but only at `-vv` and interleaved with the rest of its log, so you have to opt into a wall of output and fish the line out. Closes #959.

build now prints a short summary of its own. 📦 After installing the build dependencies it lists each resolved version as `name==version`, emitted as its own step alongside the other progress lines, at normal verbosity:

```console
$ python -m build --wheel
* Creating isolated environment: venv+pip...
  - setuptools >= 42.0.0
* Getting build dependencies for wheel...
* Installed build dependency versions:
  - setuptools==82.0.1
* Building wheel...
```

Copy the backend's line straight into your bug report — no extra flags, no scraping the installer log. The versions are read from the installed metadata in the environment, so they reflect what the build actually used, including whatever the specifiers resolved to. With `--no-isolation` build installs nothing, so the versions come from your own interpreter instead.

Without this, people wrap `python -m build` to capture the environment themselves rather than rely on the `-vv` installer output:

- [laurenfields/MSIght `publish.yml` (L27-L35)](https://github.com/laurenfields/MSIght/blob/1db01d9455fb50bded0dcc921ebb0046bad536a4/.github/workflows/publish.yml#L27-L35) — brackets `python -m build --no-isolation --verbose` with `pip freeze` and an explicit `import setuptools; print(setuptools.__version__)` to record exactly what built the package.
- [Rupt/tmsh `Makefile` (L37-L41)](https://github.com/Rupt/tmsh/blob/3bd2eb1289d644bea8bfe16426b87fcac677c98f/Makefile#L37-L41) — `pip freeze > requirements.txt` to snapshot the env, then `python -m build --no-isolation` so the recorded freeze *is* the build environment.
- [skorch-dev/skorch `deploy.sh` (L59-L67)](https://github.com/skorch-dev/skorch/blob/f7161d00de841d648c1842d9a638968e2c34d9b0/scripts/deploy.sh#L59-L67) — `python -m pip list` to log the environment right before `python -m build --sdist --wheel`.
- [octodns/octodns `cibuild-module` (L19-L45)](https://github.com/octodns/octodns/blob/14c3563677464537cd121a87d062dba831c0700b/script/cibuild-module#L19-L45) — logs an `## environment & versions` header before `python -m build`, then dumps `pip freeze` as `## installed modules` for debugging.


src/build/env.py
```diff
@@ -33,7 +33,11 @@
 import typing
 import warnings
 
+from packaging.requirements import InvalidRequirement, Requirement
+from packaging.utils import canonicalize_name
+
 from . import _ctx
+from ._compat.importlib import metadata as importlib_metadata
```
layday, line 44: Should be added to `__lazy_modules__ ` above.

layday (dismissed):

src/build/env.py
henryiii: Do we not have flake8-lazy running in CI? I think this was an early adopter, so I might not have added it.

src/build/__main__.py
```diff
@@ -190,7 +197,10 @@ def _bootstrap_build_env(
             # first install the build dependencies
             install(builder.build_system_requires, _fresh=True)
             # then get the extra required dependencies from the backend (which was installed in the call above :P)
-            install(builder.get_requires_for_build(distribution, config_settings))
+            extra_requires = builder.get_requires_for_build(distribution, config_settings)
+            install(extra_requires)
+
+            _log_dependency_versions(env, builder.build_system_requires | extra_requires)
```
layday, line 219: This indents the installed deps but is disconnected from the previous line which is emitted by `builder.get_requires_for_build`, i.e. that one comes after the other is coincidental.  I think this should be coded as a step and arranged like `DefaultIsolatedEnv.install`'s `log` call.

docs/explanation/how-it-works.rst
```diff
@@ -230,6 +230,24 @@ This:
...
 Useful for tools that need package metadata without building the full package.
 
+***************************
+ Diagnostics and Verbosity
+***************************
+
+``pyproject.toml`` declares build dependencies as *specifiers* (``setuptools >= 40.8.0``), not exact versions. The
+installer resolves each specifier to a concrete version at build time, and on an isolated build that version lives only
+in the temporary environment, which build deletes once it finishes. So the most useful piece of information for a
+backend bug report, the exact backend version, is gone by the time you need it.
```
layday: This is not quite accurate. build emits the installer's output with `-vv`. Both pip and uv print the installed versions.

src/build/__main__.py
gaborbernat: Reworked: it's now emitted as its own step (heading plus `- name==version` bullets, like `DefaultIsolatedEnv.install`'s log) instead of loose indented lines, and shown at normal verbosity. Dropped the now-unused `debug` log kind.

docs/explanation/how-it-works.rst
gaborbernat: You're right. Reworded to acknowledge the installer prints versions at `-vv`; the value here is a concise, always-on summary so you don't have to scrape the installer log. Also made every example output a real run.

src/build/env.py
gaborbernat: Declared in `__lazy_modules__` now (the flake8-lazy hook added in #1088 enforces this), keeping the import lazy.

src/build/__main__.py
```diff
@@ -114,25 +114,28 @@ def _make_logger() -> _ctx.Logger:
...
-                        _cprint('{dim}{}{reset}', fill(line, initial_indent='> '), file=sys.stderr)
-                case ('subprocess', 'stdout' | 'stderr'):
-                    for line in message.splitlines():
-                        _cprint('{dim}{}{reset}', fill(line, initial_indent='< '), file=sys.stderr)
-                case _:
-                    print(fill(message, initial_indent='  '), file=sys.stderr)  # noqa: T201
-
-    return log
+
+def _log(message: str, *, fill: Callable[..., str], kind: tuple[str, ...] | None = None) -> None:
```
layday: My motivation for keeping this under `_make_logger` was to avoid it being called directly by accident.

gaborbernat: Good point — reverted to a nested closure inside `_make_logger` (kept a small `emit` helper nested too), so the log callable can't be invoked directly without the verbosity/fill setup. Pushed.

docs/reference/cli.rst
```diff
@@ -53,6 +53,34 @@ interpreter checked and, for each requirement, the version specifier that was ``
...
+Regardless of verbosity, after installing the build dependencies of an isolated build, build prints a summary of the
+resolved versions, one ``name==version`` per line. build reads these from the isolated environment's installed metadata,
+so they reflect what was installed rather than the specifiers in ``pyproject.toml``:
+
+.. code-block:: console
+
+    $ python -m build --wheel
+    * Creating isolated environment: venv+pip...
+    * Installing packages in isolated environment:
+      - setuptools >= 42.0.0
```
henryiii, line 74: Just curious (the answer might be yes!), do we need to list the packages we are planning to install then listing the installed versions too? Could we just do the later, perhaps?

gaborbernat: Good question — they coincide only when a requirement is already an exact pin of a marker-free, extra-free name, which is unusual for `build-system.requires`. The `Installing packages` line is the *request* (the verbatim requirement strings); the summary is the *resolved result*, read back from the installed metadata. They diverge on ranges, markers, extras, and URLs.

For example:

```toml
[build-system]
requires = ["setuptools>=42", "tomli>=1.1.0; python_version<'3.11'"]
build-backend = "setuptools.build_meta"
```

```console
$ python -m build --wheel
  - setuptools>=42
  - tomli>=1.1.0; python_version<'3.11'
* Getting build dependencies for wheel...
* Installed build dependency versions:
  - setuptools==82.0.1
* Building wheel...
```

Two things the request line carries that the summary can't:

- **Membership differs** — `tomli` is requested, but on 3.11+ its marker is false, so nothing installs and it's absent from the summary.
- **Range → pin** — `setuptools>=42` is what we asked for; `setuptools==82.0.1` is what we got, which is the bit you need to reproduce a build or file a bug.

A third case: `installed_versions` skips non-PEP-508 entries, so a `pkg @ git+https://…` build requirement shows up *only* in the request line. So I'd keep both — dropping the request line would silently lose the marker'd and URL requirements.

docs/reference/cli.rst
```diff
@@ -53,6 +53,34 @@ interpreter checked and, for each requirement, the version specifier that was ``
...
+****************
+ Verbose Output
+****************
+
+Repeating ``-v`` raises the verbosity level. Each level adds to the previous one:
+
+- ``-v`` streams the output of the environment-creation and dependency-installation subprocesses.
+- ``-vv`` additionally passes ``-v`` through to the installer.
+
+Regardless of verbosity, after installing the build dependencies of an isolated build, build prints a summary of the
```
henryiii, line 65: Interesting, I was thinking that `--quiet` turned this off, but this is correct, build always prints output no matter what.

henryiii (approved):

abitrolly:
@gaborbernat nice one. The distribution version discovery methods seems different from https://github.com/pypa/build/pull/969/changes#diff-4f1bb46d65de7fbb1573dfc2a2bca9f63c7b03d55e74c53339f0c8ee4fa054d8R56.  No idea which should be the officially endorsed, though.

Which agent + model have you used?

abitrolly:
This output breaks step structure.
```
$ python -m build --wheel
* Creating isolated environment: venv+pip...
  - setuptools >= 42.0.0
* Getting build dependencies for wheel...
* Installed build dependency versions:
  - setuptools==82.0.1
* Building wheel...
```

`* Getting build dependencies for wheel...` is a step, but `* Installed build dependency versions:` is just a debug output for the step, so it should be indented under it. FWIW.

Like that:

```
$ python -m build --wheel
* Creating isolated environment: venv+pip...
  - setuptools >= 42.0.0
* Getting build dependencies for wheel...
  Installed build dependency versions:
  - setuptools==82.0.1
* Building wheel...
```

layday:
No, not necessarily. "Getting build dependencies for wheel" corresponds to `get_requires_for_build_wheel`.  In the vast majority of cases, this will be empty, and the only deps installed are the build system table's.  If `get_requires_for_build_wheel` had been non-empty, you would've seen another "Installing packages" step.

gaborbernat:
Good point on the step structure — the summary is output of the get-requires step, not a phase of its own. Fixed in #1096, which renders it indented under the step exactly as you sketched.

gaborbernat:
@layday you're right — I had the attribution wrong: the summary covers what the "Installing packages" step(s) put into the environment, and it landing after "Getting build dependencies" is coincidental ordering, which is the very reason it became a step during review. Closed #1096 and keeping the step as is.

abitrolly:
@layday I didn't get the logic here.

"Getting build dependencies for wheel" is a step in a workflow. I don't know if it is `get_requires_for_build_wheel` or not and it doesn't seem relevant to me as a user. I see the message "Installed build dependency versions" as a completion of this step. Because otherwise, what is the step the message belongs to?

layday:
It's not unreasonable that "Getting build dependencies for \<dist format\>" should encompass both the build system deps and the dist-specific deps, but it's not how the tool's currently structured. If I have time, I might look into how these can be reorganised to be less confusing.

abitrolly:
Maybe I don't have a picture in my head what a full `build` workflow looks like. For the lack of a diagram, the build log is the diagram, so I may miss that some things are meant to be separate. Like "build system deps" and "dist-specific build deps". I guess "build-system deps" are fetched by bootstrap code before the build system is invoked, and the build system didn't even started at this point, and "dist-specific build deps" are fetched by the bootstrapped build system after it read `pyproject.toml` config and decided what to build.

PR #1087, opened 2026-06-04
✨ feat(cli): surface debug paths when a backend build fails

### LINKED ISSUE #966: keep backend logs?
I can't seem to find in the docs how to keep the backend logs (and even better the whole temporary directory where the building is happening).

It's rather annoying to see an error from meson-python backend but being unable to see the actual meson log.
(and yes, in my case running meson standalone works, but invoking it via build does not, so it's really a debugging need)

gaborbernat:
A plain `python -m build` that hits a backend failure deletes its temporary working directories before anyone can look inside them, so the first failure leaves nothing to debug from. This is what [#966](https://github.com/pypa/build/issues/966) ran into with meson-python: the error was visible but the meson log lived in a directory that was already gone. 🔍

On any backend failure build now prints a `TIP` line. When no working directory is pinned it tells you to re-run with `--env-dir` and `--sdist-extract-dir` so the environment and the extracted sources survive the next run; when they are already pinned it reports where they were kept. The line links to a new `Debug a failed build` how-to that walks through streaming backend output with `-vv`, keeping both working directories, and reaching the backend's own log files. That last part stays the backend's job: PEP 517 gives the frontend no way to discover where a backend writes its logs, so the guide shows the meson-python `builddir` setting as the worked example rather than pretending build can capture them. ✨

The behavior change is limited to the failure path, where build now emits one extra hint before exiting; successful builds are untouched.

Stacked on #1083 (`--env-dir`) and #1085 (`--sdist-extract-dir`) — the hint reuses both flags and should merge after them. Depends on #1083 and #1085.

Closes #966

### DIFF
--- a/docs/how-to/troubleshooting.rst
+++ b/docs/how-to/troubleshooting.rst
@@ -463,13 +463,12 @@ build removes that location after a successful build and keeps it after a failur
 environment in ``.build-env`` for you to examine. Setting ``TMPDIR`` (or ``TEMP`` on Windows) only moves the temporary
 directory; it does not keep the environment around.
 
-To see backend logs in real-time, use verbose output:
+build forwards the backend's output to your terminal as it happens, so compilation logs and error messages are already
+visible. Raise the verbosity to add build's own diagnostics and stream the environment setup live:
 
 .. code-block:: console
 
     $ python -m build -vv
-
-This shows all output from the build backend, including compilation logs for C extensions and detailed error messages.
 
 .. _debug-a-failed-build:
 
@@ -483,8 +482,10 @@ need.
 Stream the backend output
 =========================
 
-build forwards everything the backend writes to its standard output and error. Raise the verbosity to keep the full
-stream, including compiler invocations and backend tracebacks:
+build forwards everything the backend writes to standard output and error straight to your terminal, so compiler
+invocations and backend tracebacks already appear without any extra flag. Raise the verbosity to add build's own
+diagnostics and to stream the environment setup (installing the backend and its dependencies) live instead of only on
+failure:
 
 .. code-block:: console
 
@@ -507,15 +508,15 @@ Find the backend's own logs
 ===========================
 
 build cannot capture log *files* the backend writes inside its own working directory, because PEP 517 gives the frontend
-no way to discover them. Each backend exposes its own setting for keeping them. For ``meson-python``, point its build
-directory at a path you control:
+no way to discover them. Each backend exposes its own setting for keeping them. For ``meson-python`` and
+``scikit-build``, point its build directory at a path you control:
 
 .. code-block:: console
 
-    $ python -m build -C builddir=.meson-build
+    $ python -m build -C build-dir=.meson-build
 
-The meson log then survives at ``.meson-build/meson-logs/meson-log.txt``. Check your backend's documentation for the
-equivalent setting.
+The backend then writes its build tree and logs under ``.meson-build`` (meson-python leaves its log at
+``.meson-build/meson-logs/meson-log.txt``). Check your backend's documentation for the equivalent setting.
 
 **********************
  Still having issues?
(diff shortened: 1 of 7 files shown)


docs/how-to/troubleshooting.rst
```diff
@@ -471,6 +471,52 @@ To see backend logs in real-time, use verbose output:
...
+**********************
+
+When a backend fails, build prints a ``TIP`` line that links back here. Work through these steps to gather what you
+need.
+
+Stream the backend output
+=========================
+
+build forwards everything the backend writes to its standard output and error. Raise the verbosity to keep the full
+stream, including compiler invocations and backend tracebacks:
```
henryiii: "Raise the verbosity to keep the full stream, including compiler invocations and backend tracebacks:" Those are from the build backend, so `-vv` doesn't affect them. Adding `-v` (and `-vv`) only affects a little bit of build's own output.

(not true for pip, which I guess is what this is thinking of; build and uv build are better than pip install)

docs/how-to/troubleshooting.rst
```diff
@@ -471,6 +471,52 @@ To see backend logs in real-time, use verbose output:
...
+Find the backend's own logs
+===========================
+
+build cannot capture log *files* the backend writes inside its own working directory, because PEP 517 gives the frontend
+no way to discover them. Each backend exposes its own setting for keeping them. For ``meson-python``, point its build
+directory at a path you control:
+
+.. code-block:: console
+
+    $ python -m build -C builddir=.meson-build
```
henryiii:
```suggestion
    $ python -m build -C build-dir=.meson-build
```

This is the correct form (https://mesonbuild.com/meson-python/reference/config-settings.html) and also supported by scikit-build-core.

docs/how-to/troubleshooting.rst
```diff
@@ -471,6 +471,52 @@ To see backend logs in real-time, use verbose output:
...
+
+``--env-dir`` keeps the isolated environment (the installed backend and its dependencies) on failure; see `Inspect the
+build environment`_. ``--sdist-extract-dir`` keeps the extracted sdist, which is the directory the backend actually runs
+in, so any files the backend leaves next to your sources stay put.
+
+Find the backend's own logs
+===========================
+
+build cannot capture log *files* the backend writes inside its own working directory, because PEP 517 gives the frontend
+no way to discover them. Each backend exposes its own setting for keeping them. For ``meson-python``, point its build
```
henryiii:
```suggestion
no way to discover them. Each backend exposes its own setting for keeping them. For ``meson-python`` and ``scikit-build``, point its build
```

docs/how-to/troubleshooting.rst
```diff
@@ -471,6 +471,52 @@ To see backend logs in real-time, use verbose output:
...
+directory at a path you control:
+
+.. code-block:: console
+
+    $ python -m build -C builddir=.meson-build
+
+The meson log then survives at ``.meson-build/meson-logs/meson-log.txt``. Check your backend's documentation for the
+equivalent setting.
```
henryiii: Generalize too (or add tabs).

docs/how-to/troubleshooting.rst
```diff
@@ -463,13 +463,60 @@ build removes that location after a successful build and keeps it after a failur
...
+build cannot capture log *files* the backend writes inside its own working directory, because PEP 517 gives the frontend
+no way to discover them. Each backend exposes its own setting for keeping them. For ``meson-python`` and
+``scikit-build``, point its build directory at a path you control:
+
+.. code-block:: console
+
+    $ python -m build -C build-dir=.meson-build
+
+The backend then writes its build tree and logs under ``.meson-build`` (meson-python leaves its log at
+``.meson-build/meson-logs/meson-log.txt``). Check your backend's documentation for the equivalent setting.
```
henryiii:
```suggestion

.. code-block:: console

    $ python -m build -C build-dir=_build

The backend then usually writes its build tree and logs under ``_build``. Check your backend's documentation for the equivalent setting.
```

henryiii (approved):

PR #1088, opened 2026-06-04
👷 ci: enforce lazy imports with flake8-lazy

gaborbernat:
build keeps its command-line startup fast by deferring heavy imports, and each module records which ones in a `__lazy_modules__` list. Nothing has been checking those lists, so they drifted out of sync with the actual imports. 🐌

This adds [`flake8-lazy`](https://github.com/henryiii/flake8-lazy) as a pre-commit hook so the lists are verified on every change. It runs only against `src/`, since deferred imports matter for the shipped package rather than the test suite. Bringing the current code back into compliance meant declaring the parent packages the hook now expects (`packaging`, `build._compat`) in `__main__` and giving the `tarfile` compat shim a `__lazy_modules__` entry for `pathlib`.

The lists are declarative, so there is no runtime behavior change; the value is catching a missing or stale declaration before it ships.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -42,6 +42,11 @@ repos:
       - id: ruff-check
         args: [--fix, --show-fixes]
       - id: ruff-format
+  - repo: https://github.com/henryiii/flake8-lazy
+    rev: 1e661c3087909adc5c1603b304848b02641facdc # frozen: v0.8.1
+    hooks:
+      - id: flake8-lazy
+        files: ^src/
   - repo: https://github.com/codespell-project/codespell
     rev: "2ccb47ff45ad361a21071a7eedda4c37e6ae8c5a" # frozen: v2.4.2
     hooks:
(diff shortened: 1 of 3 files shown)


layday (approved):

henryiii:
```
args: [--apply=list]
```

would make it autocorrect.

gaborbernat:
Not possible today, unfortunately. `flake8-lazy --apply` always emits double-quoted strings and ignores the project quote style, while this repo formats with `ruff` set to `quote-style = "single"`. So `args: [--apply=list]` ping-pongs with `ruff-format` — flake8-lazy rewrites the list to double quotes, ruff-format rewrites it back to single — and `pre-commit` never converges (it also single-lines lists under its own `--line-length`, default 88 vs our 120).

Filed henryiii/flake8-lazy#75 to have `--apply` respect the project's quote style (or add `--quote-style`, including a `preserve` mode). Once that lands I'll switch the hook to autocorrect; keeping it report-only until then.

PR #1092, opened 2026-06-06
🧪 test: compact test_parse_args with a typed make_kwargs helper

gaborbernat:
`test_parse_args` spelled out the full expected build kwargs as a dict literal for every case, so each case ran ~12 lines and the parametrize table was a few hundred lines of near-duplicate boilerplate.

This adds a typed `make_kwargs` helper (a `BuildKwargs` `TypedDict`, with a small `JSONValue` alias for `config_settings`) and collapses each case to a one-line `pytest.param(...)` with an `id`. No test behavior changes — same cli args, same expected calls — it's purely less boilerplate and proper types on the expected kwargs.

Pulled out on its own so the in-flight feature PRs that extend `test_parse_args` (#1080, #1081, #1083, #1084, #1085, #1086) can rebase onto it and add their cases as one-liners instead of growing the dict-literal table further.

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -15,7 +15,7 @@ import unittest.mock
 import venv
 
 from collections.abc import Callable
-from typing import Any
+from typing import Any, TypedDict
 
 import pytest
 import pytest_mock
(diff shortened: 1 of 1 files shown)


henryiii (approved):
Looks like a great simplification!

PR #1093, opened 2026-06-06
♻️ refactor(typing): ban explicit Any across src and tests

gaborbernat:
`build` was already typed under mypy `strict`, yet `Any`, `object`, and `# type: ignore` still appeared wherever a precise type looked awkward to recover: the decoded `--config-json` payloads, the `build-system` table, the parsed CLI namespace, and a scattering of test mocks. 🎯 Each of those is a place where the checker silently stopped helping, so a wrong argument shape or a renamed mock attribute could pass review unnoticed.

The recursive data really is recursive, so it is modelled that way: `TOMLValue` and `JSONValue` are covariant `Sequence`/`Mapping` aliases, and the `build-system` table is a `TypedDict` with `NotRequired` for `backend-path`. The argument object that `argparse` produces gets a dedicated `Namespace` subclass, which turns every `args.*` read into a concrete type instead of an implicit `Any`. On the test side the heterogeneous callables — sdist writers, tar member factories, captured subprocess runners, and the optional `pytest-subtests` shim — become `Protocol` types, and mocked attributes are read from the typed patch handles rather than off the real module, which clears the last `# type: ignore` comments.

To keep this from regressing, `disallow_any_explicit` is enabled and mypy now runs over both `src` and `tests`. ✨ An accidental `Any` fails the type gate directly. This is internal-only: no public API, runtime behaviour, or CLI surface changes.

### DIFF
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -45,7 +45,7 @@ import warnings
 from functools import partial
 from tarfile import TarError
 from tarfile import open as tar_open
-from typing import Any, NoReturn, TextIO
+from typing import NoReturn, TextIO, cast
 
 import pyproject_hooks
 
@@ -68,6 +68,25 @@ if TYPE_CHECKING:
     from collections.abc import Iterator, Mapping, Sequence
 
     from build._types import ConfigSettings, Distribution, StrPath, SubprocessRunner
+
+    # A decoded JSON value, as produced by ``--config-json``. Uses the covariant ``Sequence``/``Mapping``
+    # so the ``str | list[str]`` settings from ``--config-setting`` are also assignable to it.
+    JSONValue = str | int | float | bool | None | Sequence['JSONValue'] | Mapping[str, 'JSONValue']
+
+    class _Args(argparse.Namespace):
+        """The arguments :func:`main_parser` produces, typed for the rest of the module."""
+
+        srcdir: str
+        verbosity: int
+        outdir: str | None
+        distributions: list[Distribution] | None
+        metadata: bool
+        config_settings: list[str] | None
+        config_json: str | None
+        installer: _env.Installer | None
+        no_isolation: bool
+        dependency_constraints_txt: str | None
+        skip_dependency_check: bool
 
 
 _COLORS = {
@@ -242,16 +261,11 @@ def _handle_build_error() -> Iterator[None]:
             _cprint()
             _error(str(e))
 
-        if e.exc_info:
-            tb_lines = traceback.format_exception(
-                e.exc_info[0],
-                e.exc_info[1],
-                e.exc_info[2],
-                limit=-1,
-            )
+        if e.exc_info[0] is not None:
+            tb_lines = traceback.format_exception(e.exc_info[0], e.exc_info[1], e.exc_info[2], limit=-1)
             tb = ''.join(tb_lines)
         else:  # pragma: no cover
-            tb = traceback.format_exc(-1)  # type: ignore[unreachable]
+            tb = traceback.format_exc(limit=-1)
         _cprint('\n{dim}{}{reset}\n', tb.strip('\n'))
         _error(str(e))
     except Exception as e:  # pragma: no cover
@@ -409,7 +423,7 @@ def main_parser() -> argparse.ArgumentParser:
             self,
             option_strings: Sequence[str],
             dest: str,
-            default: Any = None,
+            default: int | None = None,
             help: str | None = None,
         ) -> None:
             super().__init__(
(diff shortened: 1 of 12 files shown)


src/build/__main__.py
```diff
@@ -69,6 +69,25 @@
 
     from build._types import ConfigSettings, Distribution, StrPath, SubprocessRunner
 
+    # A decoded JSON value, as produced by ``--config-json``. Uses the covariant ``Sequence``/``Mapping``
+    # so the ``str | list[str]`` settings from ``--config-setting`` are also assignable to it.
+    JSONValue = str | int | float | bool | None | Sequence['JSONValue'] | Mapping[str, 'JSONValue']
```
layday, line 74: I'm not sure we should be trying to encode the return type of `json.load` or `tomllib.load`, but both type aliases should  go in `build._types` regardless.

src/build/__main__.py
```diff
@@ -69,6 +69,25 @@
...
+
+    class _Args(argparse.Namespace):
```
layday, line 76:
```suggestion
    class _CliArgs(argparse.Namespace):
```

src/build/_builder.py
```diff
@@ -43,23 +41,46 @@
...
+
+    from collections.abc import Iterable, Iterator, Mapping, Sequence
+    from typing import TypedDict
 
     if sys.version_info < (3, 11):
-        from typing_extensions import Self
+        from typing_extensions import NotRequired, Self
     else:
-        from typing import Self
+        from typing import NotRequired, Self
```
layday, line 52: Unless ruff complains, this can be collapsed into a single `typing_extensions` import.

src/build/_builder.py
```diff
@@ -43,23 +41,46 @@
...
+        | datetime.time
+        | Sequence['TOMLValue']
+        | Mapping[str, 'TOMLValue']
+    )
+
+    # The validated ``[build-system]`` table.
+    BuildSystemTable = TypedDict(
+        'BuildSystemTable',
+        {'requires': list[str], 'build-backend': str, 'backend-path': NotRequired[list[str]]},
+    )
```
layday, line 74: Let's type this out as a class.

src/build/env.py
```diff
@@ -71,7 +71,7 @@ def make_extra_environ(self) -> Mapping[str, str] | None:
 
 
 def _has_dependency(
-    name: str, minimum_version_str: str | None = None, /, **distargs: object
+    name: str, minimum_version_str: str | None = None, /, **distargs: list[str]
```
layday, line 74: This is too specific.

src/build/_builder.py
```diff
@@ -94,51 +115,57 @@ def _read_pyproject_toml(path: StrPath) -> Mapping[str, Any]:
...
+    requires = build_system['requires']
+    if not isinstance(requires, list) or not all(isinstance(i, str) for i in requires):
         msg = '`requires` must be an array of strings'
         raise BuildSystemTableValidationError(msg)
 
-    match build_system_table:
-        case {'build-backend': str()}:
-            pass
-        case {'build-backend': _}:
+    table: BuildSystemTable = {
```
henryiii, line 140: Is it better to lose the match statement here for typing, or did a model just prefer older ways of doing things?

gaborbernat: Ah, sounds good to me either way!

henryiii (approved):

layday (commented):
I typed this out but never pressed submit it seems.

tests/test_main.py
```diff
@@ -26,6 +26,14 @@
 from build._compat import importlib as _importlib
 
 
+if TYPE_CHECKING:
+    from conftest import SubTests
+
+
+class CapturedRunner(Protocol):
```
layday, line 33: This is the same type as `builder.RunnerType`.

src/build/_builder.py
```diff
@@ -94,51 +115,57 @@ def _read_pyproject_toml(path: StrPath) -> Mapping[str, Any]:
...
+    requires = build_system['requires']
+    if not isinstance(requires, list) or not all(isinstance(i, str) for i in requires):
         msg = '`requires` must be an array of strings'
         raise BuildSystemTableValidationError(msg)
 
-    match build_system_table:
-        case {'build-backend': str()}:
-            pass
+    table: BuildSystemTable = {
+        'requires': [requirement for requirement in requires if isinstance(requirement, str)],
```
layday, line 141: This is redundant. We're verifying that all of the requirements are strings on line 136 above.  I assume this was only done to satisfy the type checker, presumably by a bot.

src/build/_builder.py
```diff
@@ -94,51 +115,57 @@ def _read_pyproject_toml(path: StrPath) -> Mapping[str, Any]:
...
             # If ``build-backend`` is missing, inject the legacy setuptools backend
             # but leave ``requires`` intact to emulate pip
-            build_system_table['build-backend'] = _DEFAULT_BACKEND['build-backend']
+            _find_typo(build_system, 'build-backend')
 
-    match build_system_table:
+    match build_system:
         case {'backend-path': list() as backend_path} if all(isinstance(i, str) for i in backend_path):
-            pass
+            table['backend-path'] = [path for path in backend_path if isinstance(path, str)]
```
layday, line 158: As above.

layday (commented):
Can you please open a new PR addressing these and my previous comments? I did make a start on it, but I don't think it's quite fair spending hours of my time correcting the output of a coding agent.

src/build/__main__.py
layday: This and the TOML one should be upstreamed to the `typeshed` if they correctly reflect the output of the parsers.

PR #1096, opened 2026-06-12
🐛 fix(cli): nest dependency version summary under its step

gaborbernat:
The resolved-versions summary introduced in #1086 printed with the `* ` step prefix, which misrepresents the CLI's output structure: the asterisk-prefixed lines enumerate build phases, and the summary is a result of the `Getting build dependencies` phase rather than a phase of its own. @abitrolly flagged this in post-merge review on #1086.

The summary now renders indented under the step that produced it, matching how other supplementary output nests:

```console
$ python -m build --wheel
* Creating isolated environment: venv+pip...
  - setuptools >= 42.0.0
* Getting build dependencies for wheel...
  Installed build dependency versions:
  - setuptools==82.0.1
* Building wheel...
```

The summary still prints at every verbosity level for isolated builds; only its placement changes. Since #1086 is unreleased, the changelog fragment was amended in place.


gaborbernat:
Closing per the discussion on #1086: layday is right that the summary isn't output of the get-requires step — the packages were installed by the "Installing packages" step(s), and the adjacency to "Getting build dependencies" is coincidental. Indenting it under whichever step happens to precede it misattributes the output, which is the same coincidental-ordering concern that made this a step during the #1086 review. Keeping the step as is.

PR #1101, opened 2026-06-13
chore: add lockfiles to gitignore

henryiii:
## Description

A few things added to the gitignore that show up, at least for me.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,14 @@ Dockerfile
 
 # Restore src/build in case user ignores 'build'
 !src/build
+
+# Lockfiles
+uv.lock
+*pylock*
+
+# System stuff
+.DS_Store
+
+# Symlink to AGENTS.md
+CLAUDE.md
+.claude/
--- a/.gitignore
+++ b/.gitignore
@@ -27,10 +27,3 @@ Dockerfile
 # Lockfiles
 uv.lock
 *pylock*
-
-# System stuff
-.DS_Store
-
-# Symlink to AGENTS.md
-CLAUDE.md
-.claude/


.gitignore
```diff
@@ -23,3 +23,14 @@ Dockerfile
...
+# Lockfiles
+uv.lock
+*pylock*
+
+# System stuff
+.DS_Store
+
+# Symlink to AGENTS.md
+CLAUDE.md
+.claude/
```
layday: I feel these belong in `.git/info/exclude` or your global `gitignore`.

henryiii: I wasn't sure about `.DS_Store`, I have a new-ish machine and have left it off the global for now, to see what projects don't ignore it. But most non-git tools that read this do ignore .DS_Store automatically too, like hatchling. So fine to drop.

```suggestion
# Symlink to AGENTS.md
CLAUDE.md
.claude/
```

For `CLAUDE.md`, in #1102 I'm proposing to keep an `AGENTS.md`, in which case we expect someone using Claude Code to symlink to it, so this file is expected and we don't want it committed. Less sure about `.claude` - claude sticks stuff like local settings in there, and if we ever add a `.skills`, another standard, you have to symlink that to `.claude/skills` (of course). It's not something you want globally ignored I think, since for a personal project someone might commit these files directly and not bother with following the standard.

Fine to move this to the #1102 PR. It will collide, but that's fine, can rebase. Also fine to drop it, and just add this to `.git/info/exclude` (everyone who uses will have to do the same).

henryiii: Update, since #1102 is out, this changes, I'd like to ignore AGENTS.md too.

henryiii: I've dropped everything but lock files. I can just add this to local excludes for now, happy to review a PR if someone wants more.

layday (approved):

PR #1103, opened 2026-06-13
chore: add check-dist

henryiii:
## Description

Adding check-dist. It reads the ignore list for common tools like flit, so adding the new tasks folder there makes it happy.


henryiii:
Ahh, forgot how to use my own hook. That's embarrassing.

layday (approved):

PR #1107, opened 2026-06-14
chore(deps): raise dependency floors to 3.10 minimums and drop pytest cap

henryiii:
Some of these are more technically correct, but the old version worked. Minor bumps to lower floors seem fine, though? Happy to revert some if someone prefers. I'll bump mypy in a followup.

## What

The dependency-group minimums were not revisited after the project moved to `requires-python >= 3.10`, so several floors point at releases that predate genuine 3.10 support (no cp310 wheels / never declared the classifier). This raises each to its first 3.10-supporting release and removes a cap that constrained the shared lockfile.

## Floor bumps

Verified against PyPI metadata and a real Python 3.10 interpreter:

| Package | Was | Now |
|---|---|---|
| `pytest` | 6.2.4 | 6.2.5 |
| `pytest-cov` | 2.12 | 3.0.0 |
| `pytest-mock` | 2 | 3.7.0 |
| `pytest-rerunfailures` | 9.1 | 10.2 |
| `pytest-xdist` | 1.34 | 2.4.0 |
| `wheel` | 0.36.0 | 0.37.0 |
| `setuptools_scm` | 6 | 6.4.2 |
| `filelock` | 3 | 3.4.0 |
| `coverage[toml]` | 5.1 | 6.0 |
| `diff_cover` | 3 | 6.4.2 |
| `typing-extensions` | 3.7.4.3 | 4.0.0 |
| `virtualenv` (extra) | 20.0.35 | 20.17 |

`tests/constraints.txt` synced to match: `setuptools_scm` 6.3.1→6.4.2 (6.3.1 itself predates 3.10), `wheel` 0.36.0→0.37.0.

## Cap removed

`pytest < 7.5` in the `typing` group pinned pytest across the entire universal lockfile even though the `test` group wants the latest. Replaced with `pytest >= 6.2.5`.

## Kept deliberately

- **`setuptools >= 56.0.0`** — confirmed it builds a wheel on real Python 3.10, so the floor is valid.
- **`mypy ~= 1.18.2`** — only pins the type-checker itself, not shared deps, so it does not constrain the lockfile. Intentional tool pin.

## Verification

- `uv lock --dry-run` resolves cleanly (pytest now floats to latest)
- `uv run --group mypy mypy` passes against the uncapped pytest
- Bumped constraints install together on a Python 3.10 venv
- `prek -a` passes


henryiii:
<details><summary>:robot: Reasons :robot:</summary>
<p>

Yes. Two categories: strict "first release that declared/shipped 3.10 support," and a few where I rounded up past that to avoid a known-bad early release — I'll flag those explicitly so you can decide.

**Strict first-3.10 release** (current floor predated it; bumped to exactly the first supporting version):

| Package                               | →   | Reason                                                                                                                                                               |
| ------------------------------------- | --- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `pytest` 6.2.4 → **6.2.5**            |     | 6.2.5 (2021-08-30) is the release that added the 3.10 classifier; 6.2.4 (Apr 2021) predates 3.10.                                                                    |
| `pytest-cov` 2.12 → **3.0.0**         |     | 3.0.0 (2021-10-04) first with 3.10 classifier; 2.12 (Jun 2021) predates.                                                                                             |
| `pytest-rerunfailures` 9.1 → **10.2** |     | 10.2 (2021-09-17) first with 3.10 classifier.                                                                                                                        |
| `pytest-xdist` 1.34 → **2.4.0**       |     | 2.4.0 (2021-09-21) is the first *real* release declaring 3.10 (the `0.0.0` PyPI junk upload skewed the naive sort).                                                  |
| `wheel` 0.36.0 → **0.37.0**           |     | 0.37.0 (2021-08-09) first with 3.10 classifier.                                                                                                                      |
| `coverage[toml]` 5.1 → **6.0**        |     | 6.0 (2021-10-03) is the first release officially supporting 3.10. (And `covdefaults 2.3` already requires `coverage>=6.2`, so the effective floor is higher anyway.) |
| `diff_cover` 3 → **6.4.2**            |     | 6.4.2 (2021-10-06) is both the first with the 3.10 classifier and the first release after 3.10.0 final.                                                              |

**Rounded up past the strict minimum** (judgment calls — tell me if you'd rather use the strict value):

| Package                                  | →   | Strict first-3.10 | Why I went higher                                                                                                                                                                                              |
| ---------------------------------------- | --- | ----------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `pytest-mock` 2 → **3.7.0**              |     | 3.7.0             | Already the strict value (3.7.0, 2022-01-28); no rounding.                                                                                                                                                     |
| `setuptools_scm` 6 → **6.4.2**           |     | 6.4.0             | 6.4.0 (2022-01) is the first with the 3.10 classifier; I used the 6.4.2 patch to skip the buggy 6.4.0/6.4.1. Could drop to 6.4.0 if you prefer.                                                                |
| `filelock` 3 → **3.4.0**                 |     | 3.3.1             | 3.1.0 declared 3.10 during the RC and 3.3.1 (2021-10-15) was first after 3.10 final; I picked 3.4.0 (Nov 2021) as a clean 3.10-era floor avoiding the 3.0–3.3 churn. Somewhat arbitrary — 3.3.1 is defensible. |
| `typing-extensions` 3.7.4.3 → **4.0.0**  |     | 3.10.0.0          | 3.10.0.0 (2021-05) first declared 3.10; I used the modern 4.0.0 line since 3.7.4.3 (2020) has known import issues on newer Pythons and nothing here needs a sub-4.0 version.                                   |
| `virtualenv` (extra) 20.0.35 → **20.17** |     | 20.6.0            | 20.6.0 is the strict first-3.10; I matched the project's own optional `virtualenv` extra floor (`>= 20.17`) for consistency rather than introduce a third virtualenv number.                                   |

**Kept (not bumped), with reasons:**
- `setuptools 56.0.0` — empirically built a wheel on a real 3.10 interpreter; the floor is genuinely valid, so I left it.
- `packaging 24.0`, `pip 22.3`, `tomli 1.1.0`, `covdefaults 2.3`, `importlib-metadata 4.6/5.1`, `uv 0.1.18`, `gitpython 3.1.44`, `bump-my-version 0.10` — all already at or above their first-3.10-supporting release.
- `mypy ~= 1.18.2` — kept the pin; it constrains only the type-checker, not shared lockfile deps.

The strict-first-3.10 values came from PyPI classifier/upload-date metadata; the verification (`uv lock`, mypy, installing the bumped constraints on a 3.10 venv) confirmed they resolve, but I did not run the full test suite at each individual floor. Want me to drop any of the four rounded-up ones to their strict minimum?

</p>
</details>

pyproject.toml
```diff
@@ -90,27 +90,27 @@ docs = [
...
+  "setuptools_scm >= 6.4.2",
   "pip >= 22.3",
   { include-group = "extra" },
 ]
 typing = [
   "importlib-metadata >= 5.1",
-  "pytest < 7.5",
-  "pytest-mock >= 2",
+  "pytest >= 6.2.5",
+  "pytest-mock >= 3.7.0",
```
layday: We can delete these two lines, I think? We need to install the `test` group to type check the test folder anyway.

henryiii: Reworked to use a shared group. (also pulled a change from #1106 and rewrote the changelog entry).

layday (approved):

PR #1108, opened 2026-06-14
chore: bump mypy, restrict pytest 9.1.0

henryiii:
## Description

Seems to be a simple bump with no changes for us (other than being faster).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -90,7 +90,7 @@ docs = [
   "towncrier >= 24.8",
 ]
 test-core = [
-  "pytest >= 6.2.5, !=9.1.0",
+  "pytest >= 6.2.5, != 9.1.0",
   "pytest-cov >= 3.0.0",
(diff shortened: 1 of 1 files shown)


henryiii:
Oh, maybe I should restrict pytest 9.1.0 too? As it will start breaking CI when it comes out.

pyproject.toml
```diff
@@ -90,7 +90,7 @@ docs = [
   "towncrier >= 24.8",
 ]
 test-core = [
-  "pytest >= 6.2.5",
+  "pytest >= 6.2.5, !=9.1.0",
```
henryiii:
```suggestion
  "pytest >= 6.2.5, != 9.1.0",
```

henryiii:
@layday, @gaborbernat after this I'm done with things I can think of for build right now.

gaborbernat (approved):

PR #1110, opened 2026-06-16
♻️ refactor(typing): address Any-ban review follow-ups

gaborbernat:
The type-safety work in #1093 closed a real gap: before it, a wrong argument shape or a renamed test attribute could slip through review because the checker had been told to look the other way in the awkward spots. Reviewing that change surfaced a handful of follow-ups worth settling before it becomes the baseline everyone builds on. Left alone they add quiet maintenance drag, and one of them is an annotation that overstates what the code actually accepts.

This PR clears that review feedback so the typing improvements rest on a clean footing. 🎯 The description of decoded JSON and TOML data had been copied into three files and was already drifting apart; it now lives in one place, so there is a single source of truth to keep correct. A helper that advertised it accepted far more than it ever receives now states exactly what it takes, which removes a trap for the next person who calls it. And where the code re-checked facts it had just proven, it now records that they hold instead of paying for the same validation twice.

One review suggestion was to push the JSON and TOML shape definitions upstream into Python's shared type stubs (typeshed). We looked into it and are deliberately keeping them inside `build`. 💡 The wider ecosystem has decided more than once that these parsers should stay loosely typed, because a stricter definition makes every project that reads JSON or TOML do extra work and breaks code that already exists. The JSON case was declined in python/typing#182 and again in python/typeshed#5911, with the most recent request closed in January 2026 (python/typeshed#15272); the TOML case was declined in python/typeshed#10031, whose trial run introduced errors into projects like `pytest` and `pylint`. Defining the shapes locally gives `build` the precision it wants exactly where this data is validated and consumed, without pushing that cost onto everyone downstream.

Nothing about how `build` is used changes. The public API, the command-line interface, and runtime behaviour are all untouched, so existing users and integrators see no difference. What improves is the maintenance cost and a type story that stays accurate enough to keep catching real mistakes as the project grows.

### DIFF
--- a/docs/changelog/1110.misc.rst
+++ b/docs/changelog/1110.misc.rst
@@ -1,0 +1,2 @@
+Consolidate the ``JSONValue``/``TOMLValue`` aliases in ``build._types`` and simplify the type annotations added in
+:issue:`1093` following review - by :user:`gaborbernat`
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -70,13 +70,9 @@ TYPE_CHECKING = False
 if TYPE_CHECKING:
     from collections.abc import Iterator, Mapping, Sequence
 
-    from build._types import ConfigSettings, Distribution, StrPath, SubprocessRunner
+    from build._types import ConfigSettings, Distribution, JSONValue, StrPath, SubprocessRunner
 
-    # A decoded JSON value, as produced by ``--config-json``. Uses the covariant ``Sequence``/``Mapping``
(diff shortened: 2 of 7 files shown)


gaborbernat:
@layday as requested the followup, let me know if you have any further questions here 🤔

gaborbernat:
@layday soft ping if you can, thanks!

layday:
Sorry, I haven't had time to look at this yet, but it's on my radar. Thanks!

PR #1113, opened 2026-06-24
Bump vulnerable dependency floors (virtualenv, wheel)

gaborbernat:
Raise the declared minimum versions to the first releases that resolve their published advisories, clearing build's open Dependabot alerts. This also raises `tests/constraints.txt` (the minimum-version CI matrix), so it **drops support for the older pip/virtualenv/wheel releases** build previously tested against.

Scope of use of each bumped dependency:

- **pip** `22.3` → **`26.1`** — *test-only* (`[dependency-groups].test`). At runtime build invokes whatever pip already exists in the target/bootstrap environment; it does not declare pip as an install requirement, so this floor only governs build's own test matrix. pip 26.1 requires Python ≥ 3.10, matching `requires-python`.
- **wheel** `0.37.0` → **`0.38.1`** — *test-only* (`[dependency-groups].test`). ReDoS fix.
- **virtualenv** `20.17`/`20.31` → **`20.36.1`** — *optional runtime* (the `build[virtualenv]` isolation backend in `src/build/env.py`) plus the dev `extra` group. Fixes the activation-script command injection and directory-creation TOCTOU. 20.36.1 supports every Python build targets.

`tests/constraints.txt` is updated to the new minimums so the `min` job resolves. The `< 20.31` `--no-wheel` compatibility branch in `env.py` is retained as a defensive fallback (still covered by `test_virtualenv_no_wheel_flag`) for users who install a below-floor virtualenv; happy to drop it if you'd prefer.

Advisories cleared: virtualenv (cmd-injection + TOCTOU), wheel (ReDoS), pip (path traversal, command injection, tar/zip confusion, untrusted control sphere, symlink extraction).

### DIFF
--- a/docs/changelog/1113.misc.rst
+++ b/docs/changelog/1113.misc.rst
@@ -1,0 +1,1 @@
+Raise vulnerable dependency floors (pip, virtualenv, wheel) to releases that fix their advisories.
--- a/docs/changelog/1113.misc.rst
+++ b/docs/changelog/1113.misc.rst
@@ -1,1 +1,1 @@
-Raise vulnerable dependency floors (pip, virtualenv, wheel) to releases that fix their advisories.
+Raise vulnerable dependency floors (virtualenv, wheel) to releases that fix their advisories.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -103,7 +103,7 @@ test = [
   'setuptools >= 56.0.0; python_version == "3.11"',
   'setuptools >= 67.8.0; python_version >= "3.12"',
   "setuptools_scm >= 6.4.2",
-  "pip >= 26.1",
+  "pip >= 22.3",
   { include-group = "test-core" },
   { include-group = "extra" },
 ]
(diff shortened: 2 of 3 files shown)


layday:
I don't mind the other two - virtualenv because it's a direct dependency and v20.36.1 is from Jan, and wheel because it's only used for testing - but pip 26.1 is very fresh.  Are we saying that we no longer support earlier versions of pip, and if so, should we also warn about it in `IsolatedEnv`? On the other hand, if this is just about getting rid of the dependabot alert, I think being able to test with the earliest supported version of pip is probably more practical than placating dependabot.

gaborbernat:
pip is also only used in testing here 🤔

layday:
pip is an undeclared dependency of the default isolated env.

gaborbernat:
Good point — dropped the pip bump. pip in the test group only drives `tests/constraints.txt` (the minimum-version matrix), and pinning it to 26.1 would have lost CI coverage of the older pip versions the default isolated env still supports at runtime. Reverted pip to 22.3 (matching #1074) and kept the virtualenv and wheel floors. I'll dismiss the pip Dependabot alert as test-only for now; raising the actual runtime floor + an `IsolatedEnv` warning can be a separate change if we want to go there.

layday (approved):

Issue #1116, opened 2026-07-01
build review from Claude Fable 5


henryiii:
Fable is back, so this is what I wanted to do with #1097. Following https://iscinumpy.dev/post/claude-code-reviews/.

:robot: _AI text below_ :robot:

Review complete. I read all 12 source modules (~2,400 lines) and verified the suspicious findings empirically. Overall the codebase is in very good shape — already modern (match statements, lazy imports, `functools.cache`, protocols), so most findings are small. Two are real bugs worth fixing.

## Bugs

**1. `_fs_supports_symlink` interpolates the file object, not its name — `src/build/env.py:504`**
`dest = f'{tmp_file}-b'` formats the `NamedTemporaryFile` object's repr (verified: `"<_TemporaryFileWrapper file=<_io.BufferedRandom name='...'>>-b"`). This only runs on Windows, where `<`, `>`, and quotes are invalid filename characters, so `os.symlink` always raises `OSError` and the function always returns `False` — symlink-based venv creation is silently never used on Windows even when Developer Mode enables it. Fix: `f'{tmp_file.name}-b'`.

**2. Wheel filename regex misparses build tags — `src/build/_util.py:25`**
Verified: `foo-1.0-1-py3-none-any.whl` parses as distribution=`foo-1.0`, version=`1`. In `metadata_path`'s fallback path (`_builder.py:379`), this constructs `foo-1.0-1.dist-info` instead of `foo-1.0.dist-info`, so `extractall` extracts nothing and a nonexistent path is returned with no error. Since `packaging >= 24.0` is already a dependency, `packaging.utils.parse_wheel_filename` handles this correctly (verified) — or constrain the regex (per spec, a build tag starts with a digit: `(-(?P<build_tag>\d[^-]*))?` with non-greedy distribution/version).

**3. Warnings print to stdout, corrupting `--metadata` JSON output — `src/build/__main__.py:134`**
`_showwarning` calls `_cprint(..., file)` with `file=None`, which means stdout — unlike the default `warnings.showwarning` (stderr) and unlike every other diagnostic in the CLI. With `--metadata`, stdout is machine-readable JSON; a `TypoWarning` or the valueless `--config-setting` warning lands in the middle of it and breaks `json.loads` consumers. Should default to `sys.stderr`.

**4. `os.linesep.join` in text-mode files double-translates on Windows — `src/build/env.py:417,426,479`**
[...]

PR #1117, opened 2026-07-01
chore: default --installer to pip in argparse

henryiii:
Maybe it was this way because we were thinking to add a smart default that would use either? That never happened, so this seem fine and simpler to check.

## Summary
- `--installer` had no `default`, so `args.installer` was `None` at runtime even though it's typed as `_env.Installer` (`Literal['pip', 'uv']`) everywhere downstream. It only worked because callers just checked `== 'uv'`.
- Set `default='pip'` on the argparse option (matching the existing help text "defaults to pip") and tighten the `_Args` stub annotation from `_env.Installer | None` to `_env.Installer`.
- Updated `tests/test_main.py` assertions/defaults that previously expected `installer` to be `None`.

### DIFF
--- a/docs/changelog/1117.misc.rst
+++ b/docs/changelog/1117.misc.rst
@@ -1,0 +1,1 @@
+The ``--installer`` CLI option now has an explicit ``pip`` default instead of relying on ``None``.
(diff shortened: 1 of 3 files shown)


layday (approved):

gaborbernat (approved):
LGTM

PR #1118, opened 2026-07-01
fix: build correct dest path in Windows symlink support probe

henryiii:
I'm on macOS, so couldn't run the regression test. If there's any question about this, I can revert the fix and see if the CI fails.

## Summary

`_fs_supports_symlink()` in `src/build/env.py` builds a destination path for a symlink probe on Windows with:

```python
dest = f'{tmp_file}-b'
```

This interpolates the `NamedTemporaryFile` object's `repr()` (e.g. `<_TemporaryFileWrapper file=<_io.BufferedRandom name='...'>>`) instead of its `.name` attribute. Since `<` and `>` are invalid characters in Windows filenames, `os.symlink()` always raises `OSError`, so the function always returns `False` - even on systems where Windows Developer Mode has symlinks enabled. This silently prevents symlink-based venv creation from ever being used on Windows.

Fix: use `tmp_file.name` instead of `tmp_file`.

### DIFF
--- a/docs/changelog/1118.bugfix.rst
+++ b/docs/changelog/1118.bugfix.rst
@@ -1,0 +1,2 @@
+Fix the Windows symlink support probe always returning ``False`` due to a stale object interpolated into the destination
+path - by :user:`henryiii`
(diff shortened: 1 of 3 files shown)


gaborbernat (approved):

layday (approved):

gaborbernat (approved):
LGTM

PR #1119, opened 2026-07-01
fix: derive dist-info directory from wheel contents, not filename

henryiii:
I think fixing the regex is also an option. But this seems fine and it looks like we already do it elsewhere.

## Summary
- `ProjectBuilder.metadata_path`'s fallback path (used when a backend lacks `prepare_metadata_for_build_wheel`) derived the wheel's dist-info directory name by parsing the wheel *filename* with a hand-rolled regex. That regex's greedy groups misparse build-tagged filenames — e.g. `foo-1.0-1-py3-none-any.whl` was read as distribution=`foo-1.0`, version=`1`, silently dropping the build tag — producing a dist-info name that doesn't exist in the wheel, so metadata extraction found nothing and returned a path that doesn't exist.
- Fixed by inspecting the built wheel's contents to find the single `*.dist-info/METADATA` entry instead of guessing from the filename, mirroring the existing approach in `_wheel_metadata` (`__main__.py`).
- Removed `_util.py`'s now-unused `parse_wheel_filename` / `_WHEEL_FILENAME_REGEX` (no remaining callers in `src/` or `tests/`).

### DIFF
--- a/src/build/_builder.py
+++ b/src/build/_builder.py
@@ -35,7 +35,7 @@ from ._exceptions import (
     BuildSystemTableValidationError,
     TypoWarning,
 )
-from ._util import check_dependency, parse_wheel_filename
+from ._util import check_dependency
 
 
 TYPE_CHECKING = False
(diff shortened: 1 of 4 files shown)


gaborbernat (approved):

src/build/_builder.py
```diff
@@ -372,16 +372,25 @@ def metadata_path(self, output_directory: StrPath) -> str:
 
         # fallback to build_wheel hook
         wheel = self.build('wheel', output_directory)
-        match = parse_wheel_filename(os.path.basename(wheel))
-        if not match:
+        try:
+            with zipfile.ZipFile(wheel) as w:
```
layday, line 376: We're opening the zip file twice - can we keep it open if no error is raised?

layday (changes requested):

gaborbernat (approved):
LGTM

src/build/_builder.py
henryiii: :robot: _AI text below_ :robot:

Good point — fixed in #1126, which opens the wheel once and keeps only the open itself inside the `try`.

henryiii: Ahh, I was approving line by line to catch this, however it responds to comments must be in my auto-approve list. 😠

henryiii: Ah, it is, it uses `gh api`, which I added because there's read only stuff that's useful from that.

PR #1120, opened 2026-07-01
fix: use LF instead of os.linesep when writing pip/uv requirements files

henryiii:
I'm not on Windows, but this seems correct. You aren't supposed to use linesep in text mode.

## Summary

- `src/build/env.py` writes the pip requirements file, the pip constraints file, and the uv constraints file via `tempfile.NamedTemporaryFile('w', ...)` (text mode, default `newline=None`) and `os.linesep.join(...)`.
- Text mode already translates every `\n` written to `os.linesep` on output. On Windows, `os.linesep` is `'\r\n'`, so joining with it first inserts literal `\r\n` between lines, and the text-mode translation then turns each of *those* `\n` into `\r\n` as well, producing `\r\r\n` line endings.
- pip and uv happen to tolerate this today because they strip line whitespace when parsing requirements/constraints files, so no user-visible breakage — but it's incorrect and fragile.
- Fix: join with `'\n'` in all three places and let the text-mode file object perform the OS-appropriate newline translation on write.

### DIFF
--- a/docs/changelog/1120.bugfix.rst
+++ b/docs/changelog/1120.bugfix.rst
@@ -1,0 +1,2 @@
+Write pip/uv requirements and constraints files with ``\n`` instead of ``os.linesep``, avoiding doubled ``\r\r\n`` line
+endings on Windows - by :user:`henryiii`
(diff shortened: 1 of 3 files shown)


gaborbernat (approved):

layday (approved):

gaborbernat (approved):
LGTM

PR #1121, opened 2026-07-01
fix: minor robustness fixes from code review

henryiii:
I need to edit this a bit, leaving it in draft until I do.

A batch of minor robustness fixes found during code review.

1. **`env._get_minimum_pip_version_str`**: parse the macOS release with `split(".", 1)[0]` instead of slicing at `find(".")`. The old code truncated the last digit of a dot-less release (e.g. `"15"` became `"1"`) and raised `ValueError` on an empty release. An empty release now falls back to the generic minimum.
2. **`_builder._parse_build_system_table`**: return a fresh copy of the default build-system table (including a fresh `requires` list) so a caller mutating the result cannot corrupt the shared module-level default for every later builder in the process.
3. **`env.DefaultIsolatedEnv.__enter__`**: assign `self._path` before calling `os.path.realpath`, so a failure there no longer masks the original exception with an `AttributeError` from `__exit__` and no longer leaks the `mkdtemp` directory.
4. **`_builder.metadata_path`**: raise `BuildException` (not a bare `ValueError`) for an invalid wheel, so the CLI renders it nicely like every other failure in this package.
5. **`_builder._read_pyproject_toml`**: remove stray trailing spaces from two error messages.

Regression tests were added for items 1-4 (test-first, confirmed failing before the fixes).

### DIFF
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -131,6 +131,8 @@ class DefaultIsolatedEnv(IsolatedEnv):
                         raise BuildException(msg)
             os.makedirs(path, exist_ok=True)
 
+        # Set early so ``__exit__`` can clean up even if the calls below fail.
+        self._path = path
         try:
             # Call ``realpath`` to prevent spurious warning from being emitted
             # that the venv location has changed on Windows for the venv impl.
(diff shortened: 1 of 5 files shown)


src/build/env.py
```diff
@@ -132,6 +132,9 @@ def __enter__(self) -> Self:
             os.makedirs(path, exist_ok=True)
 
         try:
+            # Set early so ``__exit__`` can clean up even if the calls below fail.
+            self._path = path
+
```
henryiii:
```suggestion
        # Set early so ``__exit__`` can clean up even if the calls below fail.
        self._path = path
        try:
```

gaborbernat (approved):

layday (approved):

PR #1123, opened 2026-07-01
fix: decide color support per output stream, not solely from stdout

henryiii:
I did a bit of cleanup (see last commit), I think this is ready.

## Summary

- `_init_colors` decided color support solely from `sys.stdout.isatty()`, but nearly all colored output (log lines, warnings, TIP hints, tracebacks, `_error`) is written to **stderr**; only the final "Successfully built ..." message goes to stdout. This meant `python -m build > log` needlessly disabled colors on a tty stderr, and `python -m build 2> log` wrote raw ANSI escapes into the log while stdout was a tty.
- `_styles` now holds a `{'stdout': ..., 'stderr': ...}` mapping instead of a single style set, and `_cprint(fmt, msg, file)` picks the style set for its target stream (`file or sys.stdout`).
- `NO_COLOR` still disables color on both streams (keeping the existing warning when both `NO_COLOR` and `FORCE_COLOR` are set), and `FORCE_COLOR` still enables color on both streams. Otherwise each stream gets color independently based on its own `.isatty()`.
- Updated the direct `_styles.get()` use in `main()`'s success message to use the stdout styles.

## Test plan

- [x] Added a regression test (`test_colors`, extended; `test_cprint_selects_style_per_stream`, new) covering stdout-tty/stderr-non-tty and the inverse, confirmed it fails on the old single-stream logic and passes after the fix.
- [x] `uv run --group test pytest` (285 passed)
- [x] `uv run --group mypy mypy` (clean)
- [x] `prek -a --quiet` (clean)

### DIFF
--- a/docs/changelog/1123.bugfix.rst
+++ b/docs/changelog/1123.bugfix.rst
@@ -1,0 +1,2 @@
+Decide color support independently for stdout and stderr instead of only checking ``stdout.isatty()``, so redirecting
+one stream no longer disables or leaks ANSI colors on the other - by :user:`henryiii`
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -108,23 +108,25 @@ _NO_COLORS = dict.fromkeys(_COLORS, '')
 
 
 _DEFAULT_STYLES = {'stdout': _COLORS, 'stderr': _COLORS}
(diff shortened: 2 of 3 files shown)


gaborbernat (approved):

layday (approved):

gaborbernat (approved):
LGTM

PR #1124, opened 2026-07-01
refactor: simplify code paths found in review

henryiii:

Behavior-preserving simplifications spotted in review (except item 7, a small behavior improvement noted below):

1. `_builder.py` `_parse_build_system_table`: `requires` and `backend-path` are already validated as lists of strings, so the redundant re-filtering comprehensions become plain `list(...)` copies.
2. `env.py` `_has_virtualenv`: drop the local `from packaging.requirements import Requirement` — it is already imported at module top.
3. `env.py` `_has_dependency`: use the module-level `importlib_metadata` alias instead of re-importing `._compat.importlib` (same object).
4. `_builder.py` `_call_backend`: replace the exists/isdir/makedirs branches (and their TOCTOU window) with a single `os.makedirs(outdir, exist_ok=True)` guarded against `FileExistsError`/`NotADirectoryError`, raising the same `BuildException`.
5. `_compat/tarfile.py`: hoist the duplicated Python-version predicate into a single `_HAS_DATA_FILTER` bool and remove the now-dead `TarFile` subclass (its `extraction_filter` default was never used because `safe_extractall` always passes `filter='data'`). Only `safe_extractall` is exported; `__main__.py` now uses the plain `tarfile.open` helper it already imports.
6. `__main__.py` `_bootstrap_build_env`: deduplicate the `partial(...)`/runner construction that was repeated in both branches into a shared `builder_kwargs`.
7. `__main__.py` `build_package_via_sdist`: reuse `_validate_sdist_archive(sdist)` to derive the top-level directory instead of assuming `sdist_name.removesuffix('.tar.gz')`. **Slight behavior improvement:** a nonconforming backend-produced sdist (e.g. missing PKG-INFO) is now rejected clearly up front instead of failing later with a confusing "Source ... is not a directory" (or a downstream backend error). The `test_build_package_via_sdist_incomplete_sdist` test was updated to expect this clearer error.

All existing tests pass (`282 passed`), mypy and pre-commit are clean.

### DIFF
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -76,11 +76,6 @@ if TYPE_CHECKING:
     # so the ``str | list[str]`` settings from ``--config-setting`` are also assignable to it.
     JSONValue = str | int | float | bool | None | Sequence['JSONValue'] | Mapping[str, 'JSONValue']
 
-    class _BuilderKwargs(TypedDict, total=False):
-        """Optional keyword arguments shared by both ``ProjectBuilder`` constructors."""
-
-        runner: SubprocessRunner
-
     class _Args(argparse.Namespace):
         """The arguments :func:`main_parser` produces, typed for the rest of the module."""
 
@@ -209,10 +204,10 @@ def _bootstrap_build_env(
     env_dir: str | None = None,
     runner: SubprocessRunner | None = None,
 ) -> Iterator[ProjectBuilder]:
-    builder_kwargs: _BuilderKwargs = {'runner': runner} if runner else {}
+    runner = runner or pyproject_hooks.default_subprocess_runner
     if isolation:
         with DefaultIsolatedEnv(installer=installer, path=env_dir) as env:
-            builder = ProjectBuilder.from_isolated_env(env, srcdir, **builder_kwargs)
+            builder = ProjectBuilder.from_isolated_env(env, srcdir, runner=runner)
 
             install = env.install
             if dependency_constraints_txt:
@@ -230,7 +225,7 @@ def _bootstrap_build_env(
             yield builder
 
     else:
-        builder = ProjectBuilder(srcdir, **builder_kwargs)
+        builder = ProjectBuilder(srcdir, runner=runner)
 
         if not skip_dependency_check and (missing := builder.check_dependencies(distribution, config_settings)):
             _cprint()
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -59,7 +59,7 @@ import build
 import build.env as _env
 
 from build import ProjectBuilder, _ctx
-from build._compat.tarfile import TarFile, safe_extractall
+from build._compat.tarfile import safe_extractall
 from build._exceptions import BuildBackendException, BuildException, FailedProcessError
 from build._util import format_unmet_dependencies
 from build.env import DefaultIsolatedEnv
@@ -75,6 +75,11 @@ if TYPE_CHECKING:
     # A decoded JSON value, as produced by ``--config-json``. Uses the covariant ``Sequence``/``Mapping``
     # so the ``str | list[str]`` settings from ``--config-setting`` are also assignable to it.
     JSONValue = str | int | float | bool | None | Sequence['JSONValue'] | Mapping[str, 'JSONValue']
+
+    class _BuilderKwargs(TypedDict, total=False):
+        """Optional keyword arguments shared by both ``ProjectBuilder`` constructors."""
+
+        runner: SubprocessRunner
 
     class _Args(argparse.Namespace):
         """The arguments :func:`main_parser` produces, typed for the rest of the module."""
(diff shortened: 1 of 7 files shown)


src/build/__main__.py
```diff
@@ -76,6 +76,11 @@
     # so the ``str | list[str]`` settings from ``--config-setting`` are also assignable to it.
     JSONValue = str | int | float | bool | None | Sequence['JSONValue'] | Mapping[str, 'JSONValue']
 
+    class _BuilderKwargs(TypedDict, total=False):
+        """Optional keyword arguments shared by both ``ProjectBuilder`` constructors."""
+
+        runner: SubprocessRunner
```
layday: Replicating the builder's parameters in `TypedDict` format looks like code smell to me. What's the benefit here? The type checker should be able to map `partial` args to the builder's signature, and if it doesn't, that's just unfortunate.

src/build/_builder.py
```diff
@@ -138,7 +138,7 @@ def _parse_build_system_table(pyproject_toml: Mapping[str, TOMLValue]) -> BuildS
         raise BuildSystemTableValidationError(msg)
 
     table: BuildSystemTable = {
-        'requires': [requirement for requirement in requires if isinstance(requirement, str)],
+        'requires': list(requires),
```
layday: This is already a list and is validated on line 136 above.

henryiii:
```suggestion
        'requires': requires,
```

src/build/__main__.py
henryiii:
```
type: commands[0]> mypy
src/build/__main__.py:228: error: Argument 2 to "ProjectBuilder" has incompatible type "**dict[str, SubprocessRunner]"; expected "str"  [arg-type]
Found 1 error in 1 file (checked 24 source files)
type: exit 1 (1.40 seconds) /Users/henryfs/git/pypa/build/.claude/worktrees/agent-a94ebc74b954f1e58> mypy pid=74554
.pkg: _exit> python /Users/henryfs/.local/share/uv/tools/tox/lib/python3.14/site-packages/pyproject_api/_backend.py True flit_core.buildapi
  type: FAIL code 1 (2.70=setup[1.31]+cmd[1.40] seconds)
  evaluation failed :( (2.74 seconds)
```

src/build/_builder.py
```diff
@@ -392,12 +392,11 @@ def _call_backend(
...
         callback = getattr(self._hook, hook_name)
 
-        if os.path.exists(outdir):
-            if not os.path.isdir(outdir):
-                msg = f"Build path '{outdir}' exists and is not a directory"
-                raise BuildException(msg)
-        else:
+        try:
             os.makedirs(outdir, exist_ok=True)
+        except (FileExistsError, NotADirectoryError):
```
layday: These are not exactly equivalent.  An intermediate folder could also turn out to be a file, in which case build path doesn't exist at all and the error message would be misleading.

Does `os.makedirs` ever raise `NotADirectoryError`?

henryiii: On linux/unix, yes, if the intermediate path component is a file. On windows, I'm not sure it's guaranteed. `OSError` would be safer but might over-catch.

Fine to revert this, usually the rule is it's best to not check file paths, and instead catch errors, but the alternative is really basically doing that too, so it's not better.

src/build/_builder.py
henryiii: (the list would make sure it's a copy, which I've been bit by before in both packaging and schemas, but we should not be mutating this, so it should be fine)

src/build/_builder.py
henryiii: Ahh, I can't read a diff, the `os.makedirs` was there before, it's white. The rule is applicable here then, since it's possible the file system gets modified between the check and the run, causing the exception to be missed, and surfacing the exception from `os.makedirs` instead.

src/build/__main__.py
henryiii: And this is used twice (different functions), so you can't nicely use a partial.

henryiii: We could allow None to select the default runner in Builder/.from_isolated_env. Or we could pass pyproject_hooks.default_subprocess_runner here (which would be a smaller change).

src/build/_builder.py
henryiii: I've added a test, let's see what happens on Windows.

src/build/__main__.py
henryiii: I've gone with the simpler change, let me know if you'd rather teach these to accept None.

src/build/_builder.py
henryiii: (The old code did leak an exception when there was a file further up the path instead of a directory - I didn't customize that error message, since it's sort of correct, though I can, assuming it works the same way on Windows)

henryiii: FileNotFoundError on Windows

src/build/_builder.py
```diff
@@ -155,7 +155,7 @@ def _parse_build_system_table(pyproject_toml: Mapping[str, TOMLValue]) -> BuildS
 
     match build_system:
         case {'backend-path': list() as backend_path} if all(isinstance(i, str) for i in backend_path):
-            table['backend-path'] = [path for path in backend_path if isinstance(path, str)]
+            table['backend-path'] = list(backend_path)
```
layday:
```suggestion
            table['backend-path'] = backend_path
```

henryiii: Oh, thanks, I think I forgot to pull and then it got lost in the rebase.

henryiii: Oh, unless this is a different spot

layday (approved):

PR #1125, opened 2026-07-01
tests: check send CLI warnings to stderr instead of stdout

henryiii:

## Summary
- `_showwarning` forwarded the (normally `None`) `file` argument straight to `_cprint`, which defaults `print()`'s `file=` to stdout, so warnings landed on stdout while every other CLI diagnostic (`_error`, the logger, TIPs) goes to stderr.
- This corrupts machine-readable output: a `TypoWarning` or the valueless `--config-setting` warning printed in the middle of `--metadata`'s JSON breaks `json.loads` for consumers.
- Fix: default the destination to `sys.stderr` when `file` is `None`.

## Test plan
- Added `test_showwarning_writes_to_stderr`, a direct unit test confirming `_showwarning` writes to stderr and not stdout.
- Added `test_metadata_json_output_not_corrupted_by_warning`, an end-to-end regression test that reproduces the exact JSON-corruption bug (`--metadata` + a bare `--config-setting`) and confirms it no longer occurs.
- Both tests fail with the pre-fix code (the second with `json.decoder.JSONDecodeError`) and pass after the fix.
- [x] `uv run --group test pytest`
- [x] `uv run --group mypy mypy`
- [x] `prek -a --quiet`

### DIFF
--- a/docs/changelog/1125.bugfix.rst
+++ b/docs/changelog/1125.bugfix.rst
@@ -1,0 +1,2 @@
+Send CLI warnings to stderr instead of stdout, so they no longer corrupt ``--metadata``'s JSON output - by
+:user:`henryiii`
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -757,6 +757,34 @@ def test_metadata_json_output(
     assert metadata['version'] == '1.0.0'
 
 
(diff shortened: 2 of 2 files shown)


henryiii:
I thought this was done already in #1111.

henryiii:
Ahh, this was from an old branch, it is in main. The test is a bit stronger, will reopen if people want the extra test.

PR #1132, opened 2026-07-15
🐛 fix(release): semver auto bump, changelog reflow, and roll back 1.5.1

gaborbernat:
`build` 1.5.1 went out as a patch even though it added the `project_wheel_metadata` deprecation and several features, which broke scikit-build-core when the new deprecation warning turned into a test error. 🐛 `resolve_version` mapped `--version auto`, the pre-release workflow's default, straight to a patch, ignoring the pending changelog fragments.

`auto` now derives the bump from the fragments: a `removal` is breaking and bumps major, a `feature` or `deprecation` bumps minor, and anything else bumps patch. towncrier's `removal` type used to file deprecations and removals under one heading, so it could not tell a backward-compatible warning apart from a break. This splits it into `removal` (Removals, major) and a new `deprecation` (Deprecations, minor) that map to different bumps.

That release also committed a `CHANGELOG.rst` that fails the repo's own `docstrfmt -l 120` hook. towncrier appends `` (:issue:`N`) `` past the width docstrfmt already budgeted for, so its raw output can run over 120 columns, and the release job's isolated docstrfmt hook passed those lines instead of rewrapping them. ✅ The release now reflows the built `CHANGELOG.rst` with a pinned `docstrfmt` before it commits, so a clean changelog no longer depends on that hook's environment.

The second commit rolls back the 1.5.1 release so the next run re-cuts it under one version. It restores the 28 consumed fragments, drops the 1.5.1 `CHANGELOG.rst` section, and resets `__version__` to 1.5.0. Fragments 557 and 640 move from `removal` to `deprecation`, since they add warnings rather than remove anything, so `auto` resolves the pending set to 1.6.0. A dry run of the release against this state produces 1.6.0 with a changelog the pinned docstrfmt hook accepts.

Releases also gain a review gate. The `release` environment now lists the three maintainers as required reviewers with self-review disabled, so a run waits for a maintainer other than the one who triggered it. One approval is enough given three active maintainers, which catches a wrong bump without turning any single person into a release bottleneck. That protection rule is a repo setting applied outside this diff; `docs/development/release.rst` describes it.


henryiii (approved):
I think this is fine (the huge number of files scared me, but it turns out they are just the changelog fragments. :) Though I'd rather remove the auto option entirely in a followup, it's simpler and a maintainer pushing a release should know if it's major/minor/patch. It can still be suggested locally, but I'd rather not have it as a button in the GH CI.

layday (approved):

PR #1135, opened 2026-07-16
🔧 build(type): replace mypy with ty

gaborbernat:
Replaces mypy with ty. #1136 replaces mypy with pyrefly instead, from the same starting point. The two are alternatives; pick one and close the other.

## Why consider it

ty is fast and actively developed by the team behind ruff and uv, and a single vendor across our lint and type tooling has some appeal. That is the whole case. It is not a response to a problem we have: the type check takes 22 seconds today and 15 seconds here, on a job that runs alongside the full test matrix and is mostly setup. Seven seconds is not a reason to switch.

## What we gain

Checking now covers the release tooling, which the mypy setup never looked at. That immediately turned up a real defect in the release script that has been sitting there unnoticed. #1136 finds the same one, so this is an argument for widening coverage rather than for ty.

## What it costs

ty has no strict mode. This PR turns on every rule ty has, which is the strictest it can be configured, and that still enforces less than the mypy setup it replaces. Some of what we enforce today has no ty equivalent at all, including the project's deliberate ban on escape-hatch typing.

Three ty defects force changes to shipping code rather than to configuration. Two are cosmetic but permanent until fixed upstream: dead guards added to five modules, and one suppression. The third is a genuine loss of safety. A type describing parsed configuration has to be weakened for ty to handle it, and once weakened, mypy catches a real error in that area that ty cannot see. All three are filed upstream and linked from the code. pyrefly has none of them.

ty is pre-1.0 and states that its diagnostics may change between any two releases. We therefore have to pin it, and that pin is load-bearing for CI staying green. On the independent conformance suite maintained by the Python typing council, ty scores below mypy; pyrefly scores roughly twenty points above both.

## Recommendation

Even configured as strictly as it goes, this checks less than what we have and less than #1136, and it costs working code to get there. Staying on mypy is defensible. #1136 is the better trade. I would close this one.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -115,8 +115,9 @@ typing = [
   { include-group = "test-core" },
   { include-group = "extra" },
 ]
-mypy = [
-  "mypy ~= 2.1.0",
+type = [
+  "ty>=0.0.59",
+  "gitpython >= 3.1.44",
   { include-group = "typing" },
 ]
 release = [
@@ -128,7 +129,7 @@ release = [
 dev = [
   "flit-core",
   { include-group = "test" },
-  { include-group = "mypy" },
+  { include-group = "type" },
 ]
 
 [tool.flit.sdist]
@@ -144,6 +145,8 @@ exclude = [
 
 [tool.uv]
 exclude-newer = "7 days"
+# ty releases lag the 7-day global cutoff; allow the pinned version through
+exclude-newer-package = { ty = "2026-07-13T00:00:00Z" }
 
 
 [tool.coverage]
@@ -190,20 +193,8 @@ filterwarnings = [
   "ignore:os.path.commonprefix:DeprecationWarning" # https://github.com/pypa/pyproject-hooks/pull/222
 ]
 
-[tool.mypy]
-files = ["src", "tests"]
-exclude = ["tests/packages"]
-python_version = "3.10"
-strict = true
-disallow_any_explicit = true
-enable_error_code = ["ignore-without-code", "truthy-bool", "redundant-expr"]
-warn_unreachable = true
-
-[[tool.mypy.overrides]]
-module = [
-  "virtualenv", # Optional dependency
-]
-ignore_missing_imports = true
+[tool.ty]
+environment.python-version = "3.14"
 
 [tool.ruff]
 exclude = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -117,7 +117,7 @@ typing = [
 ]
 type = [
   "ty>=0.0.59",
-  "gitpython >= 3.1.44",
+  "gitpython >= 3.1.44",  # ty checks the whole tree, including tasks/release.py
   { include-group = "typing" },
 ]
 release = [
@@ -194,7 +194,15 @@ filterwarnings = [
 ]
 
 [tool.ty]
-environment.python-version = "3.14"
+environment.python-version = "3.10"
+
+[tool.ty.src]
+exclude = ["tests/packages"]
+
+[tool.ty.rules]
+all = "error"
+# `@override` needs typing_extensions before 3.12; build ships no runtime deps for typing
+missing-override-decorator = "ignore"
 
 [tool.ruff]
 exclude = [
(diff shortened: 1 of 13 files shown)


henryiii:
`__package__` is deprecated. `__spec__.parent` is correct. ty is not understanding that this can't be None in this case.

Did we want to try pyrefly? It does better than ty when I've compared them. ty is heavily focused on fast recompute for LSPs.

Are we having a speed problem with mypy? It's gotten a _lot_ faster recently, just first run is still a bit slow.

henryiii:
Also, this is much less strict, I believe, isn't ty lenient by default, just with different config? (pyrefly is much closer to mypy in config, so not sure about ty).

IMO, this is simply too many regressions to switch.

src/build/__init__.py
```diff
@@ -5,6 +5,11 @@
 from __future__ import annotations
 
 
+# ty types module-scope `__spec__` as `ModuleSpec | None`; narrow it so `__spec__.parent`
+# resolves. https://github.com/astral-sh/ty/issues/4017
+if __spec__ is None:  # pragma: no cover
+    raise RuntimeError
```
henryiii, line 11: This is really ugly, and I'd like it resolved before considering ty seriously.

src/build/__main__.py
```diff
@@ -177,7 +177,7 @@ def log(message: str, *, kind: tuple[str, ...] | None = None) -> None:
 
 
 def _setup_cli(*, verbosity: int) -> None:
-    warnings.showwarning = _showwarning
+    warnings.showwarning = _showwarning  # ty: ignore[invalid-assignment]  # https://github.com/astral-sh/ty/issues/3962
```
henryiii, line 180: Annoying, but not a blocker.

src/build/__main__.py
```diff
@@ -834,14 +834,15 @@ def _select_build(parser: argparse.ArgumentParser, args: _Args, *, sdist_input:
             'cannot build a source distribution from a source distribution; '
             'pass --wheel to build a wheel from the sdist (see https://github.com/pypa/build/issues/311)'
         )
+    wheel_only: list[Distribution] = ['wheel']
```
henryiii, line 837: mypy inferring this was better.

src/build/_ctx.py
```diff
@@ -30,7 +35,7 @@ def _log_default(message: str, *, kind: tuple[str, ...] | None = None) -> None:
     _default_logger.log(logging.INFO, message, stacklevel=2)
 
 
-LOGGER = contextvars.ContextVar('LOGGER', default=_log_default)
+LOGGER: contextvars.ContextVar[Logger] = contextvars.ContextVar('LOGGER', default=_log_default)
```
henryiii, line 38: mypy inferring this was better.

tasks/release.py
```diff
@@ -92,7 +92,7 @@ def tag_release_commit(release_commit: Commit, repo: Repo, version: Version) ->
     existing_tags = [x.name for x in repo.tags]
     if tag_name in existing_tags:
         print(f'delete existing tag {tag_name}')
-        repo.delete_tag(tag_name)
+        repo.delete_tag(repo.tags[tag_name])
```
henryiii, line 95: I think this is a fix? Nice if so, but we can put the fix in separately.

pyproject.toml
```diff
@@ -115,8 +115,9 @@ typing = [
   { include-group = "test-core" },
   { include-group = "extra" },
 ]
-mypy = [
-  "mypy ~= 2.1.0",
+type = [
+  "ty>=0.0.59",
+  "gitpython >= 3.1.44",  # ty checks the whole tree, including tasks/release.py
```
henryiii, line 120: We can have mypy check the whole tree too.

src/build/_builder.py
```diff
@@ -63,8 +71,8 @@
         | datetime.datetime
         | datetime.date
         | datetime.time
-        | Sequence['TOMLValue']
-        | Mapping[str, 'TOMLValue']
+        | list['TOMLValue']
+        | dict[str, 'TOMLValue']
```
henryiii, line 75: I'm not sure here. The previous one was intentional (according to the comment). The ty issue says this was a true positive, but they are also unsure about closing the issue.

gaborbernat:
Superseded by #1159, which keeps mypy and runs ty and pyrefly alongside it. The review feedback from here is folded in there: the runtime `__spec__` guards are replaced with inline suppressions linking the upstream issues.

PR #1136, opened 2026-07-16
🔧 build(type): replace mypy with pyrefly

gaborbernat:
Replaces mypy with pyrefly. #1135 replaces mypy with ty instead, from the same starting point. The two are alternatives; pick one and close the other. @henryiii asked whether pyrefly does better than ty, so this is the comparison with the code attached.

## Why consider it

Two reasons, neither of them speed. Our current setup is slow to nobody: the type check takes 22 seconds today and 20 here.

The first is coverage. pyrefly enforces more than the mypy configuration it replaces, and more than #1135 can. Every rule is switched on, and everything it flagged was worth fixing.

The second is maturity. pyrefly is production ready as of its 1.0 and is the default type checker at Meta. On the independent conformance suite maintained by the Python typing council it scores roughly twenty points above both mypy and ty, and ty scores below mypy. Migration was mechanical: pyrefly read our mypy configuration and translated it without help, including the per-module exceptions.

## What it found

Four classes of problem mypy had been letting through.

A real defect in the release tooling, which the mypy setup never checked. #1135 finds this one too.

An unchecked assignment in the configuration parser, where a runtime validation convinced mypy of something it had not actually established. This is the interesting one: #1135 hits the same wall and has to resolve it by weakening the type, because ty cannot handle the accurate version. Weakening it loses real safety. Here the type stays accurate and the check gets stronger.

Five uses of a decorator form the standard library has deprecated. Neither mypy nor ty reports these; ty has the rule but does not apply it in this case, which is filed upstream.

Four attributes set outside their constructor, now declared properly. Nothing changes at runtime. Neither mypy nor ty has this check.

## What it costs

Two suppressions, both temporary and both tracked upstream. One is a genuine bug in a third-party package that mypy was papering over; #1135 carries the same one, and I have opened the fix upstream. The other is a pyrefly false positive already fixed upstream but landing after the current release, so it clears itself on the next version.

Nothing else. No changes to shipping code to satisfy the checker, no version pin holding CI together.

## Recommendation

This enforces more than mypy, more than #1135, and costs nothing in working code to get there. I would merge this and close #1135.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -115,8 +115,9 @@ typing = [
   { include-group = "test-core" },
   { include-group = "extra" },
 ]
-mypy = [
-  "mypy ~= 2.1.0",
+type = [
+  "pyrefly >= 1.1.1",
+  "gitpython >= 3.1.44",  # pyrefly checks the whole tree, including tasks/release.py
   { include-group = "typing" },
 ]
 release = [
@@ -128,7 +129,7 @@ release = [
 dev = [
   "flit-core",
   { include-group = "test" },
-  { include-group = "mypy" },
+  { include-group = "type" },
 ]
 
 [tool.flit.sdist]
@@ -190,20 +191,17 @@ filterwarnings = [
   "ignore:os.path.commonprefix:DeprecationWarning" # https://github.com/pypa/pyproject-hooks/pull/222
 ]
 
-[tool.mypy]
-files = ["src", "tests"]
-exclude = ["tests/packages"]
-python_version = "3.10"
-strict = true
-disallow_any_explicit = true
-enable_error_code = ["ignore-without-code", "truthy-bool", "redundant-expr"]
-warn_unreachable = true
+[tool.pyrefly]
+project-includes = ["src", "tests", "tasks"]
+project-excludes = ["**/tests/packages*"]
+search-path = ["src", "tests"]
+python-version = "3.10"
+preset = "strict"
+ignore-missing-imports = ["virtualenv"]  # optional dependency
 
-[[tool.mypy.overrides]]
-module = [
-  "virtualenv", # Optional dependency
-]
-ignore_missing_imports = true
+[tool.pyrefly.errors]
+# `@override` needs typing_extensions before 3.12; build takes no runtime deps for typing
+missing-override-decorator = false
 
 [tool.ruff]
 exclude = [
@@ -231,7 +229,7 @@ ignore = [
   "PLR2004", # Magic values
   "PTH",     # Too many change for now
   "Q",       # Using formatter for quotes
-  "S101",    # Assert is used by mypy and pytest
+  "S101",    # Assert is used by the type checker and pytest
   "SLF001",  # Private members are okay to access
 ]
 
@@ -260,7 +258,7 @@ ignore = [
 ]
 
 [tool.repo-review]
-ignore = ["PC140"]  # mypy in tox
+ignore = ["PC140"]  # type checker in tox
 
 
 [tool.towncrier]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -117,6 +117,7 @@ typing = [
 ]
 type = [
   "pyrefly >= 1.1.1",
+  "colorama",  # win32-only runtime dep; installed so its stubs resolve to source
   "gitpython >= 3.1.44",  # pyrefly checks the whole tree, including tasks/release.py
   { include-group = "typing" },
 ]
@@ -196,7 +197,7 @@ project-includes = ["src", "tests", "tasks"]
 project-excludes = ["**/tests/packages*"]
 search-path = ["src", "tests"]
 python-version = "3.10"
-preset = "strict"
+preset = "all"
 ignore-missing-imports = ["virtualenv"]  # optional dependency
 
 [tool.pyrefly.errors]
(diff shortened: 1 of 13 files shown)


src/build/__main__.py
```diff
@@ -215,7 +215,7 @@ def _bootstrap_build_env(
     installer: _env.Installer,
     env_dir: str | None = None,
     runner: SubprocessRunner | None = None,
-) -> Iterator[ProjectBuilder]:
+) -> Generator[ProjectBuilder]:
```
henryiii, line 218: Good, but I'd not use the one-arg version of Generator, it requires a newer Python at runtime. Can be done anytime, though, could be a PR.

src/build/_builder.py
```diff
@@ -90,6 +90,10 @@ def _find_typo(dictionary: Iterable[str], expected: str) -> None:
             )
 
 
+def _is_str_list(value: TOMLValue) -> TypeGuard[list[str]]:
+    return isinstance(value, list) and all(isinstance(i, str) for i in value)
```
henryiii, line 94: mypy feels a little smarter here, but this isn't a bad workaround.

src/build/env.py
```diff
@@ -105,6 +105,10 @@ class DefaultIsolatedEnv(IsolatedEnv):
     Isolated environment which supports several different underlying implementations.
     """
 
+    # both are set in ``__enter__`` and only valid inside the context
+    _path: str
+    _env_backend: _EnvBackend
```
henryiii, line 110: Okay. Not a great design, though, but correctly typed to match. (Same below)

tests/conftest.py
```diff
@@ -82,7 +82,7 @@ def _xfail_isolated_strict(item: pytest.Item, *, is_integration_file: bool) -> b
 
 def is_integration(item: pytest.Item) -> bool:
     # item.location is typically a (path, lineno) tuple; cast to str for typing
-    return os.path.basename(str(item.location[0])) == 'test_integration.py'
+    return os.path.basename(item.location[0]) == 'test_integration.py'
```
henryiii, line 85: Interesting, this is a 3.6 holdout, I expect. Should not have triggered anything in a type checker, and should be possible to make regardless of type checker.

tests/test_env.py
```diff
@@ -132,6 +133,7 @@ def test_venv_executable_missing_post_creation(
 
 
 @typing.no_type_check
+# pyrefly: ignore[unannotated-return]  # https://github.com/facebook/pyrefly/issues/3561
```
henryiii, line 136: I think a fix went in for this 13 hours ago (unreleased)

tests/test_projectbuilder.py
henryiii, line 1: Several pyrefly issues in this file, nothing too bad, though (and it's a test file).

henryiii (commented):
Overall, much better than ty (though ty really is mostly bad from two issues). If we had to go with one now, this is the better one, though I'm fine to wait and watch longer.

pyproject.toml
```diff
@@ -115,8 +115,10 @@ typing = [
   { include-group = "test-core" },
   { include-group = "extra" },
 ]
-mypy = [
-  "mypy ~= 2.1.0",
+type = [
+  "pyrefly >= 1.1.1",
+  "colorama",  # win32-only runtime dep; installed so its stubs resolve to source
+  "gitpython >= 3.1.44",  # pyrefly checks the whole tree, including tasks/release.py
```
henryiii, line 121: This should be in typing.

pyproject.toml
```diff
@@ -115,8 +115,10 @@ typing = [
   { include-group = "test-core" },
   { include-group = "extra" },
 ]
-mypy = [
-  "mypy ~= 2.1.0",
+type = [
+  "pyrefly >= 1.1.1",
+  "colorama",  # win32-only runtime dep; installed so its stubs resolve to source
```
henryiii, line 120: I don't know why this is here when we already have types-colorama in typing?

tests/test_env.py
```diff
@@ -44,7 +44,8 @@ def test_make_extra_environ_overrides_pythonpath() -> None:
 
 def test_installed_versions(mocker: pytest_mock.MockerFixture) -> None:
     env = build.env.DefaultIsolatedEnv()
-    env._env_backend = SimpleNamespace(purelib='/purelib')
+    # _env_backend is created lazily in __enter__, so it is absent on a fresh env
+    mocker.patch.object(env, '_env_backend', SimpleNamespace(purelib='/purelib'), create=True)
```
henryiii, line 48: But I thought the declaration was moved to the class above?

tests/test_util.py
```diff
@@ -52,7 +52,7 @@ def test_with_get_requires(package_test_metadata: str) -> None:
 
     # Setuptools < v69.0.3 (https://github.com/pypa/setuptools/pull/4159) normalized this to dashes
     assert metadata['name'].replace('-', '_') == 'test_metadata'
-    assert str(metadata['version']) == '1.0.0'
+    assert metadata['version'] == '1.0.0'
```
henryiii, line 55: Does pyrefly warn when you do `str(str())`? I'm guessing it must, since there were several of these.

gaborbernat:
Superseded by #1159, which keeps mypy and runs pyrefly and ty alongside it. The review feedback from here is folded in there: the dependencies stay in the typing group and the mypy configuration is untouched.

PR #1138, opened 2026-07-25
feat: add wheel_metadata helper

### LINKED ISSUE #1131: Deprecation of project_wheel_metadata points only at the CLI; the promised wheel_metadata() (from #1084) never shipped
### Summary

`build.util.project_wheel_metadata` was deprecated in 1.5.1, but the deprecation points **only** at the `python -m build --metadata` CLI. The programmatic replacement described in #1084 — `build.util.wheel_metadata()` with a `check_dependencies` flag — does not exist in 1.5.1 or on `main`. The merged diff for #1084 ([`cd56622`](https://github.com/pypa/build/commit/cd56622ee63b97c272941a286f4e8d96919cfc4f)) added the deprecation warning but not the new function.

Net effect: consumers of a supported, in-process Python API are being told to migrate to a subprocess, with no supported programmatic path offered.

### Why this matters

`project_wheel_metadata` returns a live `importlib.metadata.PackageMetadata` object. `python -m build --metadata` instead emits a JSON core-metadata mapping (snake_case keys, `project_urls` as a mapping) over stdout. For a library that reads metadata programmatically, the recommended migration therefore means: spawn a child interpreter, capture stdout, parse JSON, and rewrite every consumer that relied on the `Message` interface — losing the object and the `runner` customization point.

To make the trade-off concrete, I put up two alternative PRs downstream:

- jaraco/jaraco.packaging#26 — reimplements the read in-process using the still-public `ProjectBuilder` / `DefaultIsolatedEnv` / `metadata_path()` (i.e. what `build.util._project_wheel_metadata` still does internally). Clean; no consumer changes.
- jaraco/jaraco.packaging#27 — follows the CLI recommendation literally. Breaking return-type change and a rewrite of every consumer, shown side-by-side.

The fact that the tidy fix (#26) just re-inlines `build`'s own non-deprecated internals suggests the capability should stay public.

### Request

One of:

1. Ship the `wheel_metadata()` described in #1084 (the deprecation message already implies it exists), or
2. Keep a supported in-process entry point (un-deprecate, or expose the `_project_wheel_metadata` core), and only then point the deprecation at it.

Until a supported programmatic replacement lands, the deprecation message recommending a CLI leaves library consumers without a like-for-like path. Refs #557, #619, #1084.

deepakganesh78:
## Summary

- add the promised public `wheel_metadata` helper with optional dependency validation
- retain the deprecated wrapper behavior without introducing dependency checks
- update tests, deprecation guidance, and changelog fragments

## Validation

- `tox -e type,path,3.13`
- `proselint`
- Sphinx documentation build
- `towncrier --draft`
- `pre-commit run check-sdist --all-files`

Closes #1131

### DIFF
--- a/docs/changelog/1131.feature.rst
+++ b/docs/changelog/1131.feature.rst
@@ -1,1 +1,2 @@
-Add :func:`build.util.wheel_metadata` as a supported in-process metadata API with dependency checks for non-isolated builds - by :user:`deepakganesh78`
+Add :func:`build.util.wheel_metadata` as a supported in-process metadata API with dependency checks for non-isolated
+builds - by :user:`deepakganesh78`
--- a/docs/changelog/1131.feature.rst
+++ b/docs/changelog/1131.feature.rst
@@ -1,0 +1,1 @@
+Add :func:`build.util.wheel_metadata` as a supported in-process metadata API with dependency checks for non-isolated builds - by :user:`deepakganesh78`
--- a/docs/changelog/557.deprecation.rst
+++ b/docs/changelog/557.deprecation.rst
@@ -1,1 +1,1 @@
-Deprecate :func:`build.util.project_wheel_metadata`; use ``python -m build --metadata`` instead - by :user:`gaborbernat`
+Deprecate :func:`build.util.project_wheel_metadata`; prefer :func:`build.util.wheel_metadata` - by :user:`gaborbernat`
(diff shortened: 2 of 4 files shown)


henryiii:
Could you change the new one to return RawMetadata?

layday:
I'd prefer to wait with this - let's continue discussion at #1131.

layday:
I have a new helper API in the works, so I'm gonna go ahead and close this.

PR #1140, opened 2026-07-30
fix: hash verification in --dependency-constraints-txt could be silently skipped

manfred-kaiser:
## Problem

`--dependency-constraints-txt` reads the file with:

```python
constraints=set(map(str.strip, dependency_constraints_file))
```

This treats every *physical* line as one independent constraint. It does not join
backslash-continued lines, so a file like the one `pip-compile --generate-hashes`
produces:

```
hatchling==1.31.0 \
    --hash=sha256:6b48ad...
    --hash=sha256:aac80b...
```

is split into three separate set elements: the requirement and each `--hash` line on
its own. When these get written back out to a temporary file for pip/uv
(`'\n'.join(constraints)`), a `set` has no guaranteed iteration order, so the
requirement and its hashes may or may not end up adjacent again.

## Impact

This is silent, not just cosmetic. When the continuation breaks apart, pip's
constraint parser reads the orphaned `--hash` line as unattached ("has `--hash` but
no requirement, and will be ignored") and installs the requirement **without
checking any hash at all** — no warning surfaces to the `build` output.

Verified directly against a single-dependency constraints file
(`uv_build==0.12.0 --hash=sha256:...`) with a deliberately wrong hash, same file,
same command, 8 runs back to back:

```
$ python -m build --dependency-constraints-txt requirements-build.txt -o dist .

run 1: build failed, wrong hash detected
run 2: build succeeded  ← wrong hash silently accepted
run 3: build succeeded  ← wrong hash silently accepted
run 4: build failed, wrong hash detected
run 5: build succeeded  ← wrong hash silently accepted
run 6: build succeeded  ← wrong hash silently accepted
run 7: build failed, wrong hash detected
run 8: build succeeded  ← wrong hash silently accepted
```

5 of 8 runs accept the tampered hash with no error and no indication in the output
that anything was skipped. With `PYTHONHASHSEED` fixed, the outcome becomes 100%
deterministic per seed (5/5 identical for a given seed) — different seeds give
different, opposite outcomes. This isolates the cause to `set` iteration order, not
to anything timing- or network-related.

- Against a larger, six-package constraints file (a `hatchling` build tree with
  multiple multi-hash entries), the failure was consistent rather than
  intermittent, but the *symptom* varied between runs: sometimes a hash got
  attributed to the wrong package, sometimes two unrelated requirement lines were
  concatenated into one invalid line (e.g. `packaging==26.2 hatchling==1.31.0
  pluggy==1.6.0`).

So depending on the file and the hash seed, this flag can either crash with a
confusing parse error, or silently skip hash verification entirely while reporting
success.

## Fix

Join continuation lines into complete logical lines before deduplicating, so every
set element is a self-contained, order-independent constraint (comments and blank
lines are also filtered out explicitly, matching how the file is meant to be read).

## Test plan

- `tests/test_main_helpers.py`: three new tests —
  `test_parse_constraints_txt_single_line`,
  `test_parse_constraints_txt_ignores_comments_and_blank_lines`, and
  `test_parse_constraints_txt_joins_backslash_continuations` (the regression test,
  using a real `pip-compile --generate-hashes`-style fixture).
- Full existing suite (`test_main.py`, `test_env.py`, `test_main_helpers.py`) passes
  unchanged: 211 passed, 2 skipped (platform-specific).

### DIFF
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -204,36 +204,6 @@ def _error(msg: str, code: int = 1) -> NoReturn:  # pragma: no cover
     raise SystemExit(code)
 
 
-def _parse_constraints_txt(path: os.PathLike[str] | str) -> set[str]:
-    """
-    Parse a pip/uv constraints file into a set of constraint lines.
-
-    Requirement files support backslash line continuations (as produced by, e.g.,
-    ``pip-compile --generate-hashes``, where a package's ``--hash`` options are
-    wrapped onto continuation lines). These must be joined back into a single
-    logical line *before* being placed in a set: splitting on physical lines and
-    only later rejoining with ``'\\n'.join()`` loses the continuation marker's
-    positional relationship, and because sets do not preserve insertion order,
-    the rejoined file can scramble unrelated requirement and hash lines together,
-    or separate a hash from its requirement so it stops applying, silently and
-    unpredictably (depending on the interpreter's hash seed for that run).
-    """
-    constraints: list[str] = []
-    logical_line = ''
-    with open(path, encoding='utf-8') as constraints_file:
-        for raw_line in constraints_file:
-            line = raw_line.strip()
-            if logical_line:
-                line = f'{logical_line} {line}'
-                logical_line = ''
-            if line.endswith('\\'):
-                logical_line = line[:-1].strip()
-                continue
-            if line and not line.startswith('#'):
-                constraints.append(line)
-    return set(constraints)
-
-
 @contextlib.contextmanager
 def _bootstrap_build_env(
     isolation: bool,
@@ -253,7 +223,8 @@ def _bootstrap_build_env(
 
             install = env.install
             if dependency_constraints_txt:
-                install = partial(install, constraints=_parse_constraints_txt(dependency_constraints_txt))
+                with open(dependency_constraints_txt, encoding='utf-8') as dependency_constraints_file:
+                    install = partial(install, constraints=set(map(str.strip, dependency_constraints_file)))
 
             # first install the build dependencies
             install(builder.build_system_requires, _fresh=True)
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -206,9 +206,17 @@ def _error(msg: str, code: int = 1) -> NoReturn:  # pragma: no cover
 
 def _parse_constraints_txt(path: os.PathLike[str] | str) -> set[str]:
     """
-    Join backslash-continued lines before deduplicating, so a hashed requirement (e.g. from
-    ``pip-compile --generate-hashes``) can't have its ``--hash`` options split from its requirement line and lost
-    when the set is later reassembled, which would silently skip that hash check depending on the hash seed.
+    Parse a pip/uv constraints file into a set of constraint lines.
+
+    Requirement files support backslash line continuations (as produced by, e.g.,
+    ``pip-compile --generate-hashes``, where a package's ``--hash`` options are
+    wrapped onto continuation lines). These must be joined back into a single
+    logical line *before* being placed in a set: splitting on physical lines and
+    only later rejoining with ``'\\n'.join()`` loses the continuation marker's
+    positional relationship, and because sets do not preserve insertion order,
+    the rejoined file can scramble unrelated requirement and hash lines together,
+    or separate a hash from its requirement so it stops applying, silently and
+    unpredictably (depending on the interpreter's hash seed for that run).
     """
     constraints: list[str] = []
     logical_line = ''
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -224,7 +224,13 @@ def _bootstrap_build_env(
             install = env.install
             if dependency_constraints_txt:
                 with open(dependency_constraints_txt, encoding='utf-8') as dependency_constraints_file:
-                    install = partial(install, constraints=set(map(str.strip, dependency_constraints_file)))
+                    constraints_text = dependency_constraints_file.read()
+                if constraints_text.strip():
+                    # Passed through as a single element so the installer backends' `'\n'.join(constraints)` reproduces
+                    # the file byte-for-byte, instead of re-parsing it into individual lines (which can split a
+                    # requirement from its `--hash` continuation lines, e.g. from `pip-compile --generate-hashes`,
+                    # and silently drop the hash check).
+                    install = partial(install, constraints=(constraints_text,))
 
             # first install the build dependencies
             install(builder.build_system_requires, _fresh=True)
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -226,10 +226,11 @@ def _bootstrap_build_env(
                 with open(dependency_constraints_txt, encoding='utf-8') as dependency_constraints_file:
                     constraints_text = dependency_constraints_file.read()
                 if constraints_text.strip():
-                    # Passed through as a single element so the installer backends' `'\n'.join(constraints)` reproduces
-                    # the file byte-for-byte, instead of re-parsing it into individual lines (which can split a
-                    # requirement from its `--hash` continuation lines, e.g. from `pip-compile --generate-hashes`,
-                    # and silently drop the hash check).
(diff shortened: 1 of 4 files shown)


tests/test_main_helpers.py
```diff
@@ -1,8 +1,10 @@
 from __future__ import annotations
 
+import pathlib
+
 import pytest
 
-from build.__main__ import _natural_language_list
+from build.__main__ import _natural_language_list, _parse_constraints_txt
```
gaborbernat: Generally avoid testing private methods, add tests via the public APIs 🤔 to test the behaviour not the implementation detail.

src/build/__main__.py
```diff
@@ -204,6 +204,36 @@ def _error(msg: str, code: int = 1) -> NoReturn:  # pragma: no cover
     raise SystemExit(code)
 
 
+def _parse_constraints_txt(path: os.PathLike[str] | str) -> set[str]:
```
gaborbernat: Can you align docstring to 120 chars. Also the docstring should describe the why not the what, code is self documenting.

tests/test_main_helpers.py
manfred-kaiser: Good point, done: replaced the private-function tests with one via the public `build_package` API (mirrors `test_build_package_with_constraints` above it, asserting on the constraints passed to `env.install`).

src/build/__main__.py
manfred-kaiser: Reworded and wrapped to 120 columns, focused on why the join is needed rather than restating the code.

manfred-kaiser:
The two failing jobs (macOS pypy3.10/min, Windows 3.10/min) are unrelated to this change — both are transient network issues in `test_integration.py` building external packages: a `RemoteDisconnected` on macOS (Solaar) and a PyPI index serving an unexpected `Content-Type` on Windows (setuptools resolution). All jobs touching the changed code (`test_main.py`, `test_main_helpers.py`) passed. Happy to have a maintainer re-run them if needed — I don't have permission to trigger reruns on this repo myself.

layday:
These aren't valid dependency specifiers, really - `requirements.txt` is its own minilang.  We shouldn't be treating them the same way we do `pyproject.toml` dependencies or `get_requires_for_*`. I think the safest thing to do here is forward the constraints file to the installer as is. pip and uv are better equipped to parse it.

manfred-kaiser:
**Updated approach**

Instead of re-parsing `--dependency-constraints-txt` into individual lines (which is what broke backslash-continued `--hash` options in the first place), this now reads the file once and passes its content through as a single, untouched unit — the installer backends' `'\n'.join(constraints)` then reproduces it byte-for-byte. No line splitting, no dedup, no reassembly, so the bug class is gone structurally rather than patched around. Net diff is smaller than the previous fix, and only touches `_bootstrap_build_env`.

**On pip/uv and hash verification**

Testing this end-to-end (real `pip-compile --generate-hashes` file, real package, correct vs. tampered hash) surfaced something worth flagging so it isn't mistaken for a regression here: whether a `--hash` sourced only from a `-c` constraints file (not repeated on the `-r` requirement) actually gets enforced depends on the installer version — both pip and uv have had the exact same class of resolver gap, just fixed at different times:

- **uv < 0.4.6**: didn't respect hashes from constraints files. Fixed in astral-sh/uv#7093 (closing astral-sh/uv#7089), released 2024-09-05.
- **pip < 26.1**: same gap, opposite failure mode — a `--hash` in `-c` alone is never picked up, so pip aborts the build outright, correct or tampered hash alike (a hard failure, not a silent pass). Fixed in pypa/pip#13887 (closing pypa/pip#9243, tracked since pypa/pip#8792/#9020), merged 2026-04-17.

Both are pre-existing installer bugs, not something introduced or fixed by this PR — `build` passes the constraints file through unchanged either way; what a given installer version does with a hash sourced from a constraint is outside `build`'s control.

**Practically:** if you rely on `--dependency-constraints-txt` for hash verification, you need **pip ≥ 26.1** or **uv ≥ 0.4.6**. Below those versions the hash is silently not enforced (uv, pre-#7093) or the build refuses to proceed at all (pip, pre-#13887) — either way, not a `build` issue, and not something `build` can or should detect or work around on older installers.

src/build/__main__.py
```diff
@@ -224,7 +224,13 @@ def _bootstrap_build_env(
             install = env.install
             if dependency_constraints_txt:
                 with open(dependency_constraints_txt, encoding='utf-8') as dependency_constraints_file:
-                    install = partial(install, constraints=set(map(str.strip, dependency_constraints_file)))
+                    constraints_text = dependency_constraints_file.read()
+                if constraints_text.strip():
+                    # Passed through as a single element so the installer backends' `'\n'.join(constraints)` reproduces
```
layday: Not sure what "`'\n'.join(constraints)` reproduces the file byte-for-byte" is referring to here. Who's doing the joining?

src/build/__main__.py
```diff
@@ -224,7 +224,13 @@ def _bootstrap_build_env(
             install = env.install
             if dependency_constraints_txt:
                 with open(dependency_constraints_txt, encoding='utf-8') as dependency_constraints_file:
-                    install = partial(install, constraints=set(map(str.strip, dependency_constraints_file)))
+                    constraints_text = dependency_constraints_file.read()
```
layday, line 227: Can we just forward the path? Don't think we need to be reading the file at all.

src/build/__main__.py
```diff
@@ -224,7 +224,13 @@ def _bootstrap_build_env(
...
+                if constraints_text.strip():
+                    # Passed through as a single element so the installer backends' `'\n'.join(constraints)` reproduces
+                    # the file byte-for-byte, instead of re-parsing it into individual lines (which can split a
+                    # requirement from its `--hash` continuation lines, e.g. from `pip-compile --generate-hashes`,
+                    # and silently drop the hash check).
+                    install = partial(install, constraints=(constraints_text,))
```
layday, line 234: This type checks but only because Python doesn't have a type for "collection of strings which isn't itself a string" - it isn't how the parameter was meant to be used.

src/build/__main__.py
manfred-kaiser: These are two spots in `env.py`s installer backends, which write the temp constraints file passed to pip/uv via `-c`:
- https://github.com/pypa/build/blob/fb5e5a0f69b2e0fbb2c58164632cb80186a180c2/src/build/env.py#L421-L425 (`_PipInstaller.install_dependencies`)
- https://github.com/pypa/build/blob/fb5e5a0f69b2e0fbb2c58164632cb80186a180c2/src/build/env.py#L474-L478 (`_UvInstaller.install_dependencies`)

Since this PR passes a 1-tuple `(constraints_text,)`, `'\n'.join(...)` is a no-op there and the original file content reaches the temp file unchanged. Will reword the comment here so it does not reference code the reader cannot see from this file.

src/build/__main__.py
manfred-kaiser: I would like to, but `install()`/`install_dependencies()` currently have a public signature of `constraints: Collection[str]`. Forwarding a path instead means changing that signature (new type, likely a new parameter), which is a change to `build`s public Python API — anyone calling `DefaultIsolatedEnv.install()` directly is affected. That pushes this from a patch release into minor/major territory. Given the bug itself (silent hash-check bypass) is patch-worthy on its own, I would rather land the minimal fix now and treat "forward the path, stop reading the file" as a separate follow-up API change if maintainers want it.

src/build/__main__.py
manfred-kaiser: Agreed, a 1-tuple standing in for "opaque blob" is not how the parameter was meant to be used. Same tradeoff as above: it is the cost of staying inside the existing public signature for a patch fix. Happy to open a follow-up issue/PR proposing a cleaner API (e.g. a dedicated `constraints_file` path parameter) for a minor release, separate from this fix.

src/build/__main__.py
layday: Fair enough - we can change it in a follow-up.  The next release might contain one other breaking change, so it'd be a good time to get this one in too.

layday (approved):

PR #1144, opened 2026-08-12
tests: mark test_main_sdist_input_end_to_end with pytest.mark.network

frenzymadness:
## Description

`test_main_sdist_input_end_to_end` tries to install `setuptools` from PyPI, and that does not work in offline environments.

## Changelog

Micro change; I believe a changelog entry is not necessary.

## Checklist


henryiii:
This seems to be the only one missing.

henryiii (approved):

PR #1102, opened 2026-06-13
chore: add an AGENTS file

henryiii:
## Description

This is an AGENTS.md file written originally by GLM-5.1 (I like it's short, consice output), hand edited a bit for things like prek, then symlinked to CLAUDE.md and `/init` in Claude Code used to touch it up a bit (mostly for a bit that has been added since writing it).

If other people have things in AGENTS.md or CLAUDE.md that aren't there, and feel like they are worthwile, happy to take edits. I've made the branch in 

An alternative is to ignore this file and expect users to always generate it. Another variation would be to commit the symlink to CLAUDE.md, but I'd rather not have an advertisement file here just because Claude is the only harness not to support the standard. ;)

Assisted-by: ClaudeCode:claude-opus-4.8
Asssisted-by: OpenCode:GLM-5.1

PR #1146, opened 2026-08-12
📄 docs: publish llms.txt from the docs build

gaborbernat:
The [sphinx-llm](https://github.com/NVIDIA/sphinx-llm) extension emits `llms.txt` ([llmstxt.org](https://llmstxt.org/)), `llms-full.txt`, and a markdown twin of each page during the HTML build; Read the Docs serves `llms.txt` from the docs domain root ([announcement](https://about.readthedocs.com/blog/2026/02/llms-txt-support/)). Plain markdown lets agents read the docs without rendering themed, JavaScript-dependent HTML. `markdown_http_base` keeps the sitemap links absolute so they resolve from the domain root.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,15 +1,14 @@
+import os
+
 # Configuration file for the Sphinx documentation builder.
 #
 # This file only contains a selection of the most common options. For a full
 # list see the documentation:
 # https://www.sphinx-doc.org/en/master/usage/configuration.html
-
 # -- Path setup --------------------------------------------------------------
-
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
-
 import warnings
 
 import build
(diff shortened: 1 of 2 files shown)


layday:
#1102 is probably relevant.  I am very much against the use of LLMs personally, so I won't be supporting this addition.

henryiii:
I'm not sure how much this might help LLMs, generally I'd expect them to run `--help` over looking this up, and I'm not sure anything is reliably loading it. I'm generally in favor of the addition, +0.5.

This is a much less visible of an addition than a file in the root of the repo (it's not visible in the repo, or on the docs unless you load the file directly). I don't want to override @layday's wishes if this is still a strong -1.

gaborbernat:
The help message is not necessarily containing all the details I had in my experience of a few cases where they try to look up the official documentation. They just realize that it cannot parse it because it requires JavaScript and ends up falling back on GitHub code read. I think there is some benefit and very little harm in adding it.

layday:
I am -1 on this.

PR #1152, opened 2026-08-13
chore: Generator is more accurate than Iterator

henryiii:
## Description

A function with a yield is a generator. Typeshed refuses to correctly type contextmanager, but it only works on Generator, not Iterator, in practice (it requires `.throw()`). I guess pyrefly patches this or doesn't use typeshed? Anyway, this is better (I'd love for this to be fixed some day, it's actually bit real code bases), so let's type it right here.

No AI assistence used, but based loosly on #1136, which probably did have AI assistence.

## Changelog

No changelog needed.

## Checklist

### DIFF
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -68,7 +68,7 @@ from build.env import DefaultIsolatedEnv
 TYPE_CHECKING = False
 
 if TYPE_CHECKING:
-    from collections.abc import Iterator, Mapping, Sequence
+    from collections.abc import Generator, Mapping, Sequence
 
     from build._types import ConfigSettings, Distribution, JSONValue, StrPath, SubprocessRunner
 
@@ -212,7 +212,7 @@ def _bootstrap_build_env(
     installer: _env.Installer,
     env_dir: str | None = None,
     runner: SubprocessRunner | None = None,
-) -> Iterator[ProjectBuilder]:
+) -> Generator[ProjectBuilder, None, None]:
     runner = runner or pyproject_hooks.default_subprocess_runner
     if isolation:
         with DefaultIsolatedEnv(installer=installer, path=env_dir) as env:
@@ -285,7 +285,7 @@ def _build(
 
 
 @contextlib.contextmanager
-def _handle_build_error(*, env_dir: str | None, sdist_extract_dir: StrPath | None) -> Iterator[None]:
+def _handle_build_error(*, env_dir: str | None, sdist_extract_dir: StrPath | None) -> Generator[None, None, None]:
     try:
         yield
     except (BuildException, FailedProcessError) as e:
@@ -883,7 +883,7 @@ def _validate_sdist_archive(archive: StrPath) -> str:
 
 
 @contextlib.contextmanager
-def _extract_sdist(archive: StrPath, top_level: str, *, extract_dir: StrPath | None = None) -> Iterator[str]:
+def _extract_sdist(archive: StrPath, top_level: str, *, extract_dir: StrPath | None = None) -> Generator[str, None, None]:
     if extract_dir is None:
         tmp_dir = tempfile.mkdtemp(prefix='build-via-sdist-')
         try:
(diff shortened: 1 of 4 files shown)


gaborbernat (approved):
I thought, according to the specification, they are equivalent, but I am not against it in general.

henryiii:
A `Generator` also has `.send()`. `@contextlib.contextmanager` actually requires `Generator` (`self.gen.throw(...)` is called in `__exit__`), but it typed `Iterator`, because "everyone types it that way". Which isn't exactly wrong as you can see. :)

henryiii:
(Also, it was `.throw()`, not `.send()`, corrected above. Same conclusion, though, it's `Generator` only: https://docs.python.org/3/library/collections.abc.html#collections.abc.Generator)

layday:
You should be able to shorten this to omit the second and third parameters; they have defaults.

henryiii:
I left them off because defaults were added in 3.13 / typing extensions. I don't know which version are stringified annotations are valid at runtime in, maybe it's already 3.13 or higher.

See https://github.com/pypa/build/pull/1136#discussion_r3771728936

If you think it's better with them removed, I can remove them.

layday:
The future import takes precedence, though. The short form should never error.

henryiii:
Okay, there's not a way to see the types at runtime this anyway (the TYPE_CHECKING breaks it).

```PyCon
uv run --python 3.15 --with typeguard python
>>> import typeguard
>>> typeguard.install_import_hook("build")
>>> import build._builder
>>> b = build._builder.ProjectBuilder(".")
  File "<python-input-5>", line 1, in <module>
    b = build._builder.ProjectBuilder(".")
  File "/Users/henryfs/git/pypa/build/src/build/_builder.py", line 184, in __init__
    source_dir: StrPath,
                ^^^^^^^
NameError: name 'StrPath' is not defined
```

```PyCon
>>> import typing
>>> import build._builder
>>> b = build._builder.ProjectBuilder(".")
>>> typing.get_type_hints(build._builder.ProjectBuilder.__init__)
  File "<python-input-5>", line 1, in <module>
    typing.get_type_hints(build._builder.ProjectBuilder.__init__)
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/henryfs/.local/share/uv/python/cpython-3.15.0b4-macos-aarch64-none/lib/python3.15/typing.py", line 2507, in get_type_hints
    value = _eval_type(value, globalns, localns, (), format=format, owner=obj, prefer_fwd_module=True)
  File "/Users/henryfs/.local/share/uv/python/cpython-3.15.0b4-macos-aarch64-none/lib/python3.15/typing.py", line 475, in _eval_type
    return evaluate_forward_ref(t, globals=globalns, locals=localns,
                                type_params=type_params, owner=owner,
                                _recursive_guard=recursive_guard, format=format)
  File "/Users/henryfs/.local/share/uv/python/cpython-3.15.0b4-macos-aarch64-none/lib/python3.15/typing.py", line 1041, in evaluate_forward_ref
    value = forward_ref.evaluate(globals=globals, locals=locals,
                                 type_params=type_params, owner=owner, format=format)
  File "/Users/henryfs/.local/share/uv/python/cpython-3.15.0b4-macos-aarch64-none/lib/python3.15/annotationlib.py", line 203, in evaluate
    raise NameError(_NAME_ERROR_MSG.format(name=arg), name=arg)
NameError: name 'StrPath' is not defined
```

So the short form is fine.

henryiii:
To be fair, there are lot of existing 3-item Generators and no 1-item generators in the codebase currently. ;)

I'll change them all in a commit here.

layday:
> Okay, there's not a way to see the types at runtime this anyway (the TYPE_CHECKING breaks it).

I'm not a fan of the `TYPE_CHECKING` blocks.  The types we do import are really quite lightweight.

layday (approved):

PR #1155, opened 2026-08-13
refactor(typing): apply review feedback from #1093

henryiii:
Since @layday asked for an agent to address comments, here's an agent addressing the comments. :)

(and a handwritten followup comment with an improvement)

Applies the unresolved review comments from #1093 (merged before they were addressed):

- Move the `JSONValue` and `TOMLValue` aliases into `build._types`.
- Rename `_Args` to `_CliArgs`.
- Collapse the versioned `typing_extensions`/`typing` import in `_builder.py` into a single `typing_extensions` import (the block is only seen by type checkers).
- Loosen `**distargs` in `env.py` from `list[str]` to `object`.
- Reuse the `SubprocessRunner` type in `tests/test_main.py` instead of the duplicate `CapturedRunner` protocol.
- Add a `_is_list_of_str` `TypeGuard` helper so the `requires`/`backend-path` validation both raises and narrows with one shared check.

### DIFF
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -70,13 +70,9 @@ TYPE_CHECKING = False
 if TYPE_CHECKING:
     from collections.abc import Iterator, Mapping, Sequence
 
-    from build._types import ConfigSettings, Distribution, StrPath, SubprocessRunner
+    from build._types import ConfigSettings, Distribution, JSONValue, StrPath, SubprocessRunner
 
-    # A decoded JSON value, as produced by ``--config-json``. Uses the covariant ``Sequence``/``Mapping``
-    # so the ``str | list[str]`` settings from ``--config-setting`` are also assignable to it.
-    JSONValue = str | int | float | bool | None | Sequence['JSONValue'] | Mapping[str, 'JSONValue']
-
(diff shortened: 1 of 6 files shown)


henryiii:
By the way, as to the "upstream" question, I believe these recursive types are not clearly correct. I believe they are actually list/dict, but that causes covariance issues with potential subclasses (that's the reason for the comment above it). If you can edit the list or dict later, that breaks. Something like `Frozen`, if that existed in Python, would fix it.

So there are two choices, both with downsides:

* Invariant list/dict (ty wants this one): this is what mypy's docs show. But it doesn't allow assignment: `dict[str, list[str]]` is not assignable to it, since it might get mutated later.
* Covariant Sequence/Mapping is too wide, but giving up mutability causes the assignment to work.

Since we are read-only, the latter is probably better. But it can't be upstreamed.

(Or we could follow typeshed and make this Any)

henryiii:
I cleaned up the namspace; this was loosely typed, since `Namespace` has a `__getattr__`, anything is valid. Also there's no need for a cast here, you can pass a namespace object in.

layday (approved):

layday:
Thank you.

PR #1153, opened 2026-08-13
chore: turn up strictness on mypy

henryiii:
## Description

This enables all the easy settings that are not part of strict=true according to --help. Three were skipped, as they flagged intentional design patterns.

Assisted-by: ClaudeCode:claude-opus-5
(used to compare `--help`'s strict listing with the fully list from `--help`)

## Changelog

None needed.

- [ ] Added changelog fragment: `docs/changelog/<pr_number>.<type>.rst`
  - Types: `feature`, `bugfix`, `doc`, `removal`, `misc`
  - Example: `123.feature.rst` containing `` Add custom backend support - by :user:`yourname`  ``

## Checklist

- [ ] Tests pass locally (`tox`)
- [ ] Code follows project style (`tox -e fix`)
- [ ] Type checks pass (`tox -e type`)
- [ ] Documentation builds (`tox -e docs`)

PR #1151, opened 2026-08-13
chore: run mypy on more parts of the code

henryiii:
## Description

I applied full-repo checking to mypy to match #1136 and #1135.

This removes one of the two reasons listed to switch: Full tree checking.

The other reason is "maturity", which IMO has a pretty obvious answer. ;)

If we still want to evaluate a switch, this reduces the diff.

## Changelog

No fragment needed.

- [ ] Added changelog fragment: `docs/changelog/<pr_number>.<type>.rst`
  - Types: `feature`, `bugfix`, `doc`, `removal`, `misc`
  - Example: `123.feature.rst` containing `` Add custom backend support - by :user:`yourname`  ``

## Checklist

- [ ] Tests pass locally (`tox`)
- [x] Code follows project style (`tox -e fix`)
- [x] Type checks pass (`tox -e type`)
- [ ] Documentation builds (`tox -e docs`)

PR #1156, opened 2026-08-13
chore: faster mypy, fix merge error

henryiii:
## Description

* Fix merge error - #1151 and #1153 together without a rebase cause a finding in the scripts. Which was an issue, an unrecongised bump type fell through instead of erroring.
* Bump mypy
* Use the faster new parser, since it's becoming the default soon (and is 29% faster for us uncached)

## Changelog

None needed.

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -117,7 +117,7 @@ typing = [
   { include-group = "extra" },
 ]
 mypy = [
-  "mypy ~= 2.1.0",
+  "mypy ~= 2.3.0",
   { include-group = "typing" },
 ]
 release = [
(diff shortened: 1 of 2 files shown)


gaborbernat (approved):

henryiii:
And here's the speed comparison:

**`--no-incremental`, 5 interleaved pairs, direct `.venv/bin/mypy`:**

|                      | mean   |
| -------------------- | ------ |
| `--no-native-parser` | 1.27 s |
| `--native-parser`    | 0.91 s |

That is within noise of the cold-cache numbers (1.24 s → 0.88 s), which makes sense: a deleted cache and a disabled cache both force a full parse of your sources and typeshed. `--no-incremental` still writes a cache dir afterwards, so the only real difference between the two setups is that one run *reads* nothing either way.

Summary across all three cache states, native vs not:

- **No cache / cold cache:** 1.25 s → 0.89 s (~29 % faster)
- **Warm cache:** 0.10 s, parser irrelevant

PR #1157, opened 2026-08-13
tests: download integration sources once per run

henryiii:
We are downloading 192 archives per run, which are starting to get throttled. This cuts it to 4 (unique) ones.

The integration tests download four project archives from GitHub. Every matrix job downloads its own copy, which is 48 jobs × 4 archives, about 440 MB per run. GitHub started to drop those requests, so recent PRs lose a job or two to `http.client.RemoteDisconnected` in the `integration_source` fixture. For example [this run](https://github.com/pypa/build/actions/runs/31667179004) lost three jobs, each on a different `pip` case.

One new job now downloads the archives and shares them with the matrix as an artifact, so the run makes one request per archive instead of 48. The job and the artifact download are best effort: if either fails, the test fixture downloads the missing source as before.

`tests/integration_sources.py` holds the project list and the download, and `get_project` calls it. CI runs the same module as a script, which needs the standard library only. Because the fixture shares the code, a local run also gets:

- a retry with a backoff, which the fixture did not have
- a write through a temporary name, so a dropped connection cannot leave a truncated archive in `.integration-sources` that every later run then reuses

The gain here is fewer requests to GitHub, not a faster run. The new job takes 5 to 10 seconds, and the matrix starts 30 to 70 seconds later than before, most of which is the queue for one more runner. Each job still transfers the same bytes, from artifact storage instead of GitHub.

This does not replace #1150, which repairs the `--only-rerun` filter. The two changes are complementary: this one removes most of the requests, #1150 retries the ones that still fail.


tests/integration_sources.py
```diff
@@ -0,0 +1,78 @@
...
+        except NETWORK_ERRORS as exception:  # noqa: PERF203
+            partial.unlink(missing_ok=True)
+            if attempt == ATTEMPTS:
+                raise
+            print(f'attempt {attempt}/{ATTEMPTS} failed for {url}: {exception}')
+            time.sleep(2**attempt)
+        else:
+            break
+    os.replace(partial, target)
+    return target
```
layday: This could return the version as well, so you don't have to reindex `INTEGRATION_SOURCES` in `tests_integration`.

tests/test_integration.py
```diff
@@ -58,23 +51,13 @@ def _ignore_folder(base: str, filenames: list[str]) -> list[str]:
...
-    github_org_repo, version = INTEGRATION_SOURCES[name]
-    tar_filename = f'{name}-{version}.tar.gz'
-    tarball = os.path.join(tar_store, tar_filename)
-    with filelock.FileLock(os.path.join(tar_store, f'{tar_filename}.lock')):
-        if not os.path.exists(tarball):
-            url = f'https://github.com/{github_org_repo}/archive/{version}.tar.gz'
-            with urllib.request.urlopen(url) as request, open(tarball, 'wb') as file_handler:
-                shutil.copyfileobj(request, file_handler)
+    STORE_DIR.mkdir(exist_ok=True)
+    tarball = archive_path(name)
```
layday: I think these two lines can go.  You can lock on the name, which is unique.

henryiii: filelock 3.12.2 (asked AI to find the version) is required to get rid of both lines, that's the first one to make the directory automatically. That's 2023, so I think it's fine to bump.

tests/integration_sources.py
henryiii: Nice!

tests/test_integration.py
henryiii: Ah, the oldest virtualenv requires 3.20.1 anyway. :)

tests/integration_sources.py
henryiii: Ahh, did this by hand and make a mistake (still need to be extracted).

tests/constraints.txt
```diff
@@ -1,3 +1,4 @@
+filelock==3.20.1
```
gaborbernat, line 1: Why do we need to lock this? Do we have something that bumps this periodically?

henryiii: This is used by the minimum job only. If we update pyproject.toml's mins, then this should update too. (basically this is uv's `--resolution=lowest-direct`, which I forgot why we don't use here instead).

gaborbernat (approved):

layday (approved):

PR #1158, opened 2026-08-13
Revert "⚠️ feat(util): deprecate project_wheel_metadata"

henryiii:
Reverts pypa/build#1084. Per the plan in #1131, my idea is:

* Revert this (undeprecate) in 1.6.0
* Add a replacement in 1.6.0 (or later)
* Reapply this in a later release so there's a released way to avoid it.

People are depending on this, and we need to be sure the replacement is usable.

### DIFF
--- a/src/build/util.py
+++ b/src/build/util.py
@@ -3,11 +3,10 @@
 from __future__ import annotations
 
 
-__lazy_modules__ = [f'{__spec__.parent}._compat', f'{__spec__.parent}.env', 'pathlib', 'tempfile', 'warnings']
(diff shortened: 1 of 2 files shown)


layday (approved):

layday:
I'm okay reverting this.  I think the replacement needs a bit more discussing and could come in a later release.

henryiii:
> could come in a later release.

Another benefit of reverting, removes pressure on getting a replacement out. :)

gaborbernat (approved):

PR #1161, opened 2026-08-16
Drop redundant `exc_info` parameter from backend exception wrapper

layday:
## Description

Drops the `exc_info` parameter from the backend exception wrapper.  I don't recall why we had this originally - possibly for Python 2 compat?

## Changelog

## Checklist

### DIFF
--- a/src/build/__main__.py
+++ b/src/build/__main__.py
@@ -297,11 +297,11 @@ def _handle_build_error(*, env_dir: str | None, sdist_extract_dir: StrPath | Non
             _cprint('{yellow}TIP{reset} {}', hint, file=sys.stderr)
             _error(str(e))
 
-        if e.exc_info[0] is not None:
-            tb_lines = traceback.format_exception(e.exc_info[0], e.exc_info[1], e.exc_info[2], limit=-1)
-            tb = ''.join(tb_lines)
-        else:  # pragma: no cover
-            tb = traceback.format_exc(limit=-1)
+        inner_exception = e.exception
+        tb_lines = traceback.format_exception(
+            inner_exception.__class__, inner_exception, inner_exception.__traceback__, limit=-1
+        )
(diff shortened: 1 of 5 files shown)


src/build/__main__.py
```diff
@@ -297,11 +297,11 @@ def _handle_build_error(*, env_dir: str | None, sdist_extract_dir: StrPath | Non
...
             _error(str(e))
 
-        if e.exc_info[0] is not None:
-            tb_lines = traceback.format_exception(e.exc_info[0], e.exc_info[1], e.exc_info[2], limit=-1)
-            tb = ''.join(tb_lines)
-        else:  # pragma: no cover
-            tb = traceback.format_exc(limit=-1)
+        inner_exception = e.exception
+        tb_lines = traceback.format_exception(
+            inner_exception.__class__, inner_exception, inner_exception.__traceback__, limit=-1
```
```suggestion
            inner_exception, limit=-1
```
Valid since 3.10+ (https://docs.python.org/3/library/traceback.html#traceback.print_exception)

tests/test_main.py
```diff
@@ -866,10 +866,10 @@ def test_handle_build_error_build_backend_exception(mocker: pytest_mock.MockerFi
     try:
         raise exc
     except ValueError:
-        exc_info = sys.exc_info()
+        pass
```
henryiii: This is a bit confusing, maybe mention in a comment that it's here to add non-none `__traceback__` (assuming that's why it's here)?

henryiii:
Yes, from #296, we still supported 2.7 (was dropped a month later).

tests/test_main.py
henryiii: I tried removing this block, still passes tests locally.

gaborbernat (approved):

henryiii:
PRs aren't working (status.github.com), so can't officially approve, but LGTM.

PR #1159, opened 2026-08-13
🔧 build(type): run mypy, ty, pyrefly, pyright and gate type coverage

gaborbernat:
Supersedes #1135 and #1136, per the review discussion there. mypy stays and ty, pyrefly and pyright join it. 🔧 Each checker accepts annotations the others reject, users run each of them against build's public API (pyright most of all, through Pylance), and in the two closed PRs each caught defects the others missed, so checking with all four keeps the annotations correct for each audience.

The review here made a second point. Downstream users need a fully typed public API, and which internal annotations each checker tolerates matters less. 🔍 Two gates enforce that. `pyright --verifytypes build --ignoreexternal` runs against the installed package and fails when an exported symbol has an unknown or ambiguous type, and `pyrefly coverage check --strict --fail-under 100 src/build` requires an annotation on each symbol in the package. On `main` both report gaps (`__lazy_modules__`, the `_ctx` context variables, `env.INSTALLERS` and the exception attributes are unannotated); the annotations that fix them also make one ty suppression unnecessary.

Each checker gets its own tox environment and CI job. mypy keeps the historical `type` environment, `ty`, `pyrefly` and `pyright` are new, and the `type` label (`tox -m type`) runs all four. The mypy setup carries over unchanged and pyright passes in standard mode with no suppressions. Where ty has an open defect the code carries an inline suppression linking the upstream issue instead of a runtime workaround, which drops the `__spec__` guards the #1135 review objected to: [`__spec__` narrowing](https://github.com/astral-sh/ty/issues/4017), [callable-attribute assignment](https://github.com/astral-sh/ty/issues/3962) and [`functools.partial` support](https://github.com/astral-sh/ty/issues/1536), filed while writing this. Reading wheel metadata trips [python/importlib_metadata#542](https://github.com/python/importlib_metadata/issues/542) under both ty and pyrefly. pyrefly runs its `strict` preset because since 1.2 the `all` preset reports around 300 findings here, in bulk unused call results and implicit bool conversions that ruff polices.

PR #1168, opened 2026-08-25
🔧 build(type): replace mypy with pyrefly

gaborbernat:
#1159 bundled three type checkers. Running all three in development and CI feels like overkill. I have kept this PR to switching from mypy to Pyrefly for the project and its public types as requested by @layday.

### DIFF
--- a/src/build/_ctx.py
+++ b/src/build/_ctx.py
@@ -30,8 +30,8 @@ def _log_default(message: str, *, kind: tuple[str, ...] | None = None) -> None: 
     _default_logger.log(logging.INFO, message, stacklevel=2)
 
 
-LOGGER = contextvars.ContextVar('LOGGER', default=_log_default)
-VERBOSITY = contextvars.ContextVar('VERBOSITY', default=0)
+LOGGER: typing.Final[contextvars.ContextVar[Logger]] = contextvars.ContextVar('LOGGER', default=_log_default)
+VERBOSITY: typing.Final[contextvars.ContextVar[int]] = contextvars.ContextVar('VERBOSITY', default=0)
 
 
 def log_subprocess_error(error: subprocess.CalledProcessError) -> None:
--- a/src/build/_ctx.py
+++ b/src/build/_ctx.py
@@ -21,8 +21,7 @@ class Logger(typing.Protocol):  # pragma: no cover
     def __call__(self, message: str, *, kind: tuple[str, ...] | None = None) -> None: ...
 
 
-_package_name = __spec__.parent
-_default_logger = logging.getLogger(_package_name)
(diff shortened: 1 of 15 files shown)


src/build/_ctx.py
```diff
@@ -21,17 +21,16 @@ class Logger(typing.Protocol):  # pragma: no cover
...
 
 def _log_default(message: str, *, kind: tuple[str, ...] | None = None) -> None:  # noqa: ARG001
     # the log function that works in tests, real log function is set in __main__
     _default_logger.log(logging.INFO, message, stacklevel=2)
 
 
-LOGGER = contextvars.ContextVar('LOGGER', default=_log_default)
-VERBOSITY = contextvars.ContextVar('VERBOSITY', default=0)
+LOGGER: typing.Final[contextvars.ContextVar[Logger]] = contextvars.ContextVar('LOGGER', default=_log_default)
+VERBOSITY: typing.Final[contextvars.ContextVar[int]] = contextvars.ContextVar('VERBOSITY', default=0)
```
layday:
```suggestion
LOGGER = contextvars.ContextVar[Logger]('LOGGER', default=_log_default)
VERBOSITY = contextvars.ContextVar[int]('VERBOSITY', default=0)
```

src/build/env.py
```diff
@@ -62,7 +62,7 @@ class _DistArgs(typing.TypedDict, total=False):
 
 Installer = typing.Literal['pip', 'uv']
 
-INSTALLERS = typing.get_args(Installer)
+INSTALLERS: typing.Final[tuple[Installer, ...]] = typing.get_args(Installer)
```
```suggestion
INSTALLERS: tuple[Installer, ...] = typing.get_args(Installer)
```

layday (changes requested):
A couple of these were too verbose I thought, otherwise looks good.

tox.toml
```diff
@@ -47,10 +47,13 @@ commands = [
...
 
 [env.type]
-description = "run type check on code base"
+description = "run pyrefly type checks and package type coverage"
 set_env = { PYTHONWARNDEFAULTENCODING = "" }
-commands = [["mypy", { replace = "posargs", extend = true }]]
-dependency_groups = ["mypy"]
+commands = [
+    ["pyrefly", "check", { replace = "posargs", extend = true }],
+    ["pyrefly", "coverage", "check", "--strict", "--fail-under", "100", "src/build"],
```
layday: Let's do `--public-only` here so we don't have to worry about over-annotating internal symbols.

layday:
I'm okay with this but let's wait for Henry's input as well.

layday (approved):

gaborbernat:
> I'm okay with this but let's wait for Henry's input as well.

Sorry missed that, if Henry has anything to add happy to open follow-uyp PR.

PR #1169, opened 2026-08-25
👷 ci: use app token for releases

gaborbernat:
Removing the PAT owner's admin access broke the pre-release workflow. A personal token ties releases to one maintainer.

Approval may wait for hours. The job mints the one-hour App token after approval and checks out the current `main`.

An atomic push sends both refs. Admins must add App credentials and grant it bypass access to protected `main`.


layday:
I don't know enough about the GH parts of this, so I'll defer to your judgment, Bernát.

layday (approved):

PR #1171, opened 2026-08-29
docs: fix changelog page heading levels and sidebar

henryiii:
## Description

Fix up the display of the changelog; now just a normal entry, the ToC is now nested properly (first entry was raised a level before due to docstrfmt rewriting it). 

## Changelog

None needed.

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -44,6 +44,8 @@ repos:
       - id: docstrfmt
         args: ["-l", "120"]
         additional_dependencies: ["sphinx>=9.1"]
+        # title must outrank the included CHANGELOG.rst headings
+        exclude: ^docs/changelog\.rst$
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -44,6 +44,8 @@ repos:
       - id: docstrfmt
         args: ["-l", "120"]
         additional_dependencies: ["sphinx>=9.1"]
+        # title must outrank the included CHANGELOG.rst headings
```
gaborbernat, line 47: We're relying on this to reformat the changelog fragments. Is this the right approach to give up on that?

henryiii: This is not where the changelog fragments go. It’s a nearly empty file that includes the one without a title.

layday (approved):

PR #1176, opened 2026-09-03
fix: always use copies for the isolated venv on Windows

### LINKED ISSUE #1175: Windows source build failure with build 1.6.0.
Hey folks, we are running into a build failure on the Windows CI after upgrading to the latest Python build (1.6.0), here's a snippet of the CI log:
```
<   File "C:\Users\runneradmin\miniconda3\envs\sdist_test\Lib\site-
  packages\pip\_vendor\rich\console.py", line 2038, in _check_buffer
<     self._write_buffer()
<   File "C:\Users\runneradmin\miniconda3\envs\sdist_test\Lib\site-
  packages\pip\_vendor\rich\console.py", line 2067, in _write_buffer
<     from pip._vendor.rich._win32_console import LegacyWindowsTerm
<   File "C:\Users\runneradmin\miniconda3\envs\sdist_test\Lib\site-
  packages\pip\_vendor\rich\_win32_console.py", line 6, in <module>
<     import ctypes
<   File
  "C:\Users\runneradmin\miniconda3\envs\sdist_test\Lib\ctypes\__init__.py",
  line 8, in <module>
<     from _ctypes import Union, Structure, Array
< ImportError: DLL load failed while importing _ctypes: The specified module
  could not be found.
< Call stack:
```

The full log is available [here](https://github.com/dmlc/xgboost/commit/f843db0101e78aaa2f785853f729437c100ce3d8/checks/100422763664/logs). The backtrace looks complicated as errors are raised and reraised from other dependencies. To verify, we got the CI to pass by restricting `build<1.6.0`: https://github.com/dmlc/xgboost/pull/12517 , similar to what other projects have done.

Relevant issues:
- https://github.com/lightgbm-org/LightGBM/issues/7415 cc @jameslamb
- https://github.com/partcad/partcad/pull/563

henryiii:
The broken probe fixed in #1118 (1.6.0, broken previously) wasn't actually supposed to be there; `python -m venv` doesn't check for symlink support. Even if symlinks are supported, python don't always work symlinked on Windows. This PR removes the check.

Fixes #1175.

Version 1.6.0 fixed the symlink probe in #1118. Before that fix the probe always failed on Windows, so every isolated environment was built from copies. After the fix, GitHub Windows runners report symlink support, and `venv` started to symlink `python.exe`. A symlinked interpreter cannot find the DLLs of some Python builds, most visibly conda, and pip fails with `DLL load failed while importing _ssl` / `_ctypes`. Projects such as xgboost, LightGBM, and partcad have pinned `build<1.6` to work around this.

This change removes the probe and hardcodes `symlinks=os.name != 'nt'`, which is the `venv` CLI default and the behavior all releases before 1.6.0 had in practice. The two tests for the probe are removed with it.

### DIFF
--- a/docs/changelog/1175.bugfix.rst
+++ b/docs/changelog/1175.bugfix.rst
@@ -1,0 +1,1 @@
+Avoid trying to detect symlinks on Windows, regression in 1.6.0 - by :user:`henryiii` (:issue:`1175`)
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -66,6 +66,9 @@ Installer = typing.Literal['pip', 'uv']
 
 INSTALLERS: tuple[Installer, ...] = typing.get_args(Installer)
 
+# Match the ``venv`` CLI default. Symlinked interpreters on Windows can fail, see #1175.
+_USE_SYMLINKS = os.name != 'nt'
+
 
 class IsolatedEnv(typing.Protocol):
     """Isolated build environment ABC."""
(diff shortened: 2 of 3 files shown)


henryiii:
It looks like we've already started merging 2.0 things, but this should be a 1.6.1 patch.

layday (approved):

layday:
I don't know how the whole release process we got set up now is gonna work off another branch if we were to cherry-pick this change from an earlier commit.

henryiii:
I'm sure it won't work the first 7 times, but we'll get there. :)

I'm going to make a v1 branch and try to set it up with cherry-picks shortly after this goes in.

PR #1179, opened 2026-09-04
fix: honor _fresh installs with the uv installer

r3wretrhy:
## Description

`--installer uv` ignored `_fresh=True`. Pip maps that flag to `--ignore-installed` so the CLI and `project_wheel_metadata` reinstall build requirements when `PYTHONPATH` is preset. The uv backend accepted `_fresh` but never forwarded it.

Pass `uv pip install --reinstall` in that case, matching pip's behavior.

## Changelog

- [x] Added changelog fragment: `docs/changelog/1179.bugfix.rst`

## Checklist

- [x] Tests pass locally (`uv run pytest tests/test_env.py::test_uv_impl_install_cmd_well_formed tests/test_env.py::test_default_impl_install_cmd_well_formed` — 24 passed)
- [x] Code follows project style (`ruff check` / `ruff format --check`)

### DIFF
--- a/docs/changelog/1179.bugfix.rst
+++ b/docs/changelog/1179.bugfix.rst
@@ -1,0 +1,1 @@
+Honor ``_fresh`` isolated installs when using the uv installer - by :user:`r3wretrhy`
--- a/src/build/env.py
+++ b/src/build/env.py
@@ -473,7 +473,12 @@ class _UvBackend(_EnvBackend):
         if (verbosity := _ctx.verbosity) > 1:
             cmd += [f'-{"v" * min(2, verbosity - 1)}']
 
-        cmd += ['install', *requirements, '--python', self.python_executable]
+        cmd += ['install']
+        if _fresh:
+            # Match pip's ``--ignore-installed`` so CLI / project_wheel_metadata
+            # ``_fresh=True`` (preset PYTHONPATH) actually reinstalls into the env.
+            cmd += ['--reinstall']
+        cmd += [*requirements, '--python', self.python_executable]
 
         if constraints_txt_path:
             cmd += ['-c', str(constraints_txt_path)]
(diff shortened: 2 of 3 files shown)


layday:
Please provide evidence that uv is impacted by the same issue.

henryiii:
FYI, I was suspicious too that this was valid, and ran a check with Claude:

**No, the PR fixes a bug that does not exist, and the two flags are not equivalent.**

**The premise is wrong.** The reason pip needs `--ignore-installed` is that pip runs inside the target interpreter and sees packages on `PYTHONPATH` as "already satisfied". uv is an external binary that inspects the venv's site-packages directly, so it never sees `PYTHONPATH` packages. On top of that, the uv backend already strips `PYTHONPATH` from the subprocess environment at `src/build/env.py:486`. I verified this empirically with uv 0.12.10: with `PYTHONPATH` pointing at a directory containing `packaging`, `uv pip install packaging` into an empty venv installs it anyway, while plain `pip install` skips it and `pip install --ignore-installed` installs it. So the uv backend already behaves correctly without this change.

**The flags differ in semantics:**

- `--ignore-installed` (pip) does not uninstall anything. It overwrites files on top of whatever is there and can leave stale files behind.
- `--reinstall` (uv) uninstalls and reinstalls every package, and also implies `--refresh`, which forces revalidation of cached index metadata against the server. That is extra network traffic on every isolated build for no benefit, since the venv is always freshly created.

**Recommendation:** close the PR or ask for a reproducer. If a reproducer surfaces, the right uv analogue would be to look at why `PYTHONPATH` stripping did not work, not to add `--reinstall`.

r3wretrhy:
I don't have a uv reproducer. `_UvBackend.install_dependencies` already drops `PYTHONPATH` from the subprocess environment, so uv does not see host packages the way in-process pip does. `--reinstall` is also not equivalent to pip's `--ignore-installed` (it uninstalls and implies `--refresh`). Closing.

